PR #2, opened 2018-10-02
Eslint setup

eszthoff:


.eslintrc.json
```diff
@@ -0,0 +1,17 @@
...
+        "experimentalDecorators": 0,
+        "react/jsx-indent": [2, 4]
+    },
+    "globals": {
+        "describe": false,
+        "it": false,
+        "expect": false,
+        "shallow": false
+    }
+}
```
ribeiroguilherme: Might makes sense to remove `.json` from the file name, so we can have consistency with other config files

package.json
```diff
@@ -7,6 +7,7 @@
     "build": "webpack --config ./etc/webpack/webpack.dev.config.js",
     "build:prod": "webpack --config ./etc/webpack/webpack.prod.config.js",
     "start": "webpack-dev-server --config ./etc/webpack/webpack-storybook.dev.config.js",
+    "lint": "eslint .",
```
ribeiroguilherme: During the linting execution seems to have some error at `webpack-storybook.dev.config.js`

etc/webpack/utils/utils.js
```diff
@@ -1,17 +1,15 @@
-
 function getRuleJS(...includePaths) {
     return {
         test: /\.js$/,
         include: includePaths,
         use: [
             {
-                loader: 'babel-loader',
+                loader: 'babel-loader'
```
ribeiroguilherme: Might be good idea to add `eslint-loader` here, so during the build process or dev mode, we can catch errors easily

.eslintrc.json
eszthoff: sorry. that is deprecated
see: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

ribeiroguilherme: Oh ok!

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+    "printWidth": 100,
+    "singleQuote": true,
+    "tabWidth": 4,
+    "useTabs": false
```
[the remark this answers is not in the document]
eszthoff: removed

etc/webpack/utils/utils.js
eszthoff: added for prod but not for dev

PR #4, opened 2018-10-04
Add sass loader, autoprefixer, stylelint, stylelint-order plugin, css minification for production

antipin:


.stylelintrc
```diff
@@ -0,0 +1,28 @@
...
+        "font-family-name-quotes": "always-where-recommended",
+        "font-weight-notation": "numeric",
+        "indentation": 4,
+        "no-missing-end-of-source-newline": true,
+        "selector-attribute-brackets-space-inside": "never",
+        "selector-attribute-quotes": "always",
+        "selector-max-id": 0,
+        "unit-case": "lower",
+        "order/order": [
+			[
```
carlobernardini: Seems this block is indented too much @antipin

.stylelintrc
```diff
@@ -0,0 +1,28 @@
+{
+    "extends": "stylelint-config-recommended-scss",
```
eszthoff: why not use the airbnb standard for sass linting too? https://github.com/airbnb/css

PR #11, opened 2018-12-10
replace color shades with sass functions

eszthoff:


antipin:
@eszthoff nice idea with $*-factor variables – it allows us to fine-tune colors later!

The only enhancement I'd like to propose is to use `mix()` sass function as far as seems like it does exactly what we need (in the context of the idea with background and foreground colors):

```sass
$mix-factor-25: 25%;
$mix-factor-75: 75%;

$background-color: #EEE;
$foreground-color: #333;

$color-brand-25: mix($background-color, $color-brand, $mix-factor-25);
$color-brand-75: mix($foreground-color, $color-brand, $mix-factor-75);

...

```

What mix() function does:
The weight specifies the amount of the first color that should be included in the returned color. The default, `50%`, means that half the first color and half the second color should be used. `25%` means that a quarter of the first color and three quarters of the second color should be used.


Actually I've just found an excellent example and demonstration of the idea with smooth shades we discussed, please have a look at Mastering color with mixing section: https://alistapart.com/article/mixing-color-for-the-web-with-sass#section4

fkupper:
You can integrate tests into your pipeline.

PR #15, opened 2018-12-12
Move props.block and props.elem to props.bem namespace

antipin:
* Since now for stateless components you need to call props.bem.block() props.bem.elem('elemName')

### DIFF
--- a/src/packages/bem/README.md
+++ b/src/packages/bem/README.md
@@ -96,10 +96,10 @@ import classnamesMap from './Button.scss';
 
 const ButtonStateless = (props) => {
     return (
-      {/*  1. Add { ...props.block() } construction to declare node as a block */}
-      <button { ...props.block() }>
-        {/*  2. Add { ...props.elem('label') } construction to declare node as a label element */}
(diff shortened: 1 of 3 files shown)


antipin:
A question about this one. Would like to know what you guys think, as far as it affects the way we write components.

**The problem:**

Having `props.block()` and `props.elem()` functions on a props levels can easily conflict with some other props named like `block` of `elem`.


**Solution 1**

For stateless components: to rename `props.block()` and `props.elem()` to `props.bemBlock()` and `props.bemElem()` to make it more specific.

For stateful components: to rename `this.block()` and `this.elem()` to `this.bemBlock()` and `this.bemElem()` to make it more specific.

**Cons**

As for me looks a bit "bulky".

**Solution 2**

For stateless components: create a bem namespace, so it is like `props.bem.block()` and `props.bem.elem()`.

For stateful components leave it as it is now: `this.block()` and `this.elem()`.

ribeiroguilherme:
I prefer solution 1. For me looks more explicit that we are using the methods from the `bem` module.

PR #17, opened 2018-12-12
Button component

carlobernardini:
- Button component with tests + story
- Extended variable scheme
- Maps Sass variables to CSS variables


src/components/Button/Button.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
```
antipin: Button component should be listed in `src/index.js`

```js
export { default as Dummy } from './components/Dummy';
export { default as Button } from './components/Button';
```

stories/Button.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import { boolean, select, text, withKnobs } from '@storybook/addon-knobs'; // eslint-disable-line import/no-extraneous-dependencies
+import Button from '../src/components/Button';
+import { CONTEXTS, SIZES } from '../src/constants';
+
+storiesOf('Button', module)
+    .addDecorator(withKnobs)
+    .add('Button', () => (
+        <Button
```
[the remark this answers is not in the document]
carlobernardini: Yes, @antipin is looking into a fix

src/components/Button/Button.js
```diff
@@ -0,0 +1,62 @@
...
+    /** The size of the button */
+    size: PropTypes.oneOf(SIZES),
+    /** whether or not to show block - level button(full width) */
+    isBlock: PropTypes.bool,
+    /** should button be disabled or not */
+    disabled: PropTypes.bool,
+    /** type of the button */
+    type: PropTypes.oneOf(['submit', 'button']),
+    /** optional html attrinutes such as target, href, mouseOver etc. */
+    htmlAttributes: PropTypes.objectOf(
```
antipin: I think this one is redundant (also can be applied to `style` and `className`):

We can do something like that:
```js
const Button = props => {
    const { children, disabled, onClick, style, type, block, ...rest } = props;
    return (
        <button
            {...block()}
            type={type}
            onClick={onClick}
            disabled={disabled}
            {...rest}
        >
            {children}
        </button>
    );
};
```

src/components/Button/Button.js
```diff
@@ -0,0 +1,62 @@
...
+    type: 'button',
+    htmlAttributes: {},
+    onClick: null,
+    style: null,
+    className: ''
+};
+
+Button.propsToMods = ['context', 'size', 'isBlock'];
+
+export default bem(styles)(Button);
```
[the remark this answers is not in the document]
carlobernardini: Good point -- done

src/components/Button/Button.js
```diff
@@ -0,0 +1,62 @@
...
+        >
+            {children}
+        </button>
+    );
+};
+
+Button.propTypes = {
+    /** The label of the button */
+    children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+    /** The button context (e.g. brand, primary, bad, good etc. - defaults to brand) */
```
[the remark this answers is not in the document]
carlobernardini: Yep, fixed

src/components/Button/Button.js
carlobernardini: Already fixed

src/components/Button/Button.js
carlobernardini: Good point -- fixed

stories/Button.js
```diff
@@ -0,0 +1,19 @@
...
+            context={select('Context', ['link', ...CONTEXTS], CONTEXTS[0])}
+            size={select('Size', SIZES, SIZES[1])}
+            isBlock={boolean('isBlock', false)}
+            type={select('Type', ['submit', 'button'], 'submit')}
+            disabled={boolean('Disabled', false)}
+        >
```
ribeiroguilherme: Adding `className='my-custom-class'` seems to break the component. Am I doing something wrong here?

PR #24, opened 2018-12-12
Update webpack configuration

antipin:
* Bring autoprefixer back
* Remove hash appending to classnames for dev env

### DIFF
--- a/etc/webpack/utils/utils.js
+++ b/etc/webpack/utils/utils.js
@@ -1,13 +1,44 @@
-function getRuleJS(useLinting = true, ...includePaths) {
-    const loaders = useLinting ? ['babel-loader', 'eslint-loader'] : ['babel-loader'];
+const postcssAutoprefixer = require('autoprefixer'); // eslint-disable-line import/no-extraneous-dependencies
 
+function getRuleJS({ includePaths }) {
     return {
         test: /\.js$/,
         include: includePaths,
-        use: loaders
+        use: ['babel-loader']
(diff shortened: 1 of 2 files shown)


etc/webpack/utils/utils.js
```diff
@@ -1,4 +1,6 @@
-function getRuleJS(useLinting = true, ...includePaths) {
+const postcssAutoprefixer = require('autoprefixer'); // eslint-disable-line import/no-extraneous-dependencies
+
+function getRuleJS({ useLinting = true, includePaths }) {
     const loaders = useLinting ? ['babel-loader', 'eslint-loader'] : ['babel-loader'];
```
eszthoff: this is not needed

etc/webpack/webpack.config.js
```diff
@@ -27,34 +26,16 @@ const plugins = {
...
-            {
-                loader: 'sass-loader',
-                options: {
-                    includePaths: [SOURCE_PATH]
-                }
-            }
-        ]
-    }
+    js: getRuleJS({
+        useLinting: env === 'prod',
```
eszthoff: no need to set this

etc/webpack/utils/utils.js
antipin: done

etc/webpack/webpack.config.js
antipin: done

PR #25, opened 2018-12-13
Heading & Text component

carlobernardini:
* Implements a Heading component that renders h1...h4
* Implements a Text component that renders a paragraph or span (if inline)
* Adds `color-muted` (doesn't need shades, we already have `color-neutral` that comes with 4 shades)

### DIFF
--- a/src/components/Heading/Heading.js
+++ b/src/components/Heading/Heading.js
@@ -1,17 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import bem from 'bem'; // eslint-disable-line import/no-unresolved
+import bem from 'bem';
 import styles from './Heading.scss';
 import { HEADING_SIZES } from '../../constants';
 
 const { block } = bem({
     name: 'Heading',
     classnames: styles,
-    propsToMods: ['alignRight', 'level']
+    propsToMods: ['align', 'level']
 });
 
 const Heading = props => {
-    const { alignRight, children, level, ...rest } = props;
+    const { align, children, level, ...rest } = props;
     const HtmlNodeType = level;
 
     return (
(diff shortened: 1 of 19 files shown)


src/components/Heading/Heading.js
```diff
@@ -0,0 +1,38 @@
...
+    return (
+        <HtmlNodeType {...rest} {...block(props)}>
+            {children}
+        </HtmlNodeType>
+    );
+};
+
+Heading.propTypes = {
+    /** Indicates that heading text should be right-aligned */
+    alignRight: PropTypes.bool,
```
eszthoff: allow for align centre too

src/components/Heading/Heading.js
```diff
@@ -0,0 +1,38 @@
...
+    /** Heading text */
+    children: PropTypes.string.isRequired,
```
eszthoff: allow for nodes (<span> <icon> etc)

src/components/Heading/Heading.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../themes/nice/nice.scss';
+
+.Heading {
+    font: {
+        family: var(--heading-font-family);
+        weight: var(--font-weight-normal);
```
eszthoff: I think this should be themeble, some people might like bold headings and possibly colored...

src/components/Text/Text.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem'; // eslint-disable-line import/no-unresolved
```
eszthoff: we can add a eslint rule in eslintrc to ignore bem import so we don't have to add this comment each time: `import/no-unresolved: [2, { ignore: ['bem'] }]`
probably there is also a more proper solution with https://github.com/benmosher/eslint-plugin-import/blob/master/README.md#resolver-plugins. but maybe the above would do

src/components/Text/Text.js
```diff
@@ -0,0 +1,37 @@
...
+        </HtmlNodeType>
+    );
+};
+
+Text.propTypes = {
+    /** Text should be rendered inline */
+    inline: PropTypes.bool,
+    /** Render muted (auxiliary) text */
+    muted: PropTypes.bool,
+    /** Text content */
```
eszthoff: this can be also node. with <em> <span> etc.

PR #26, opened 2018-12-13
Implemented Alert component

carlobernardini:
* Adds an Alert component with tests and story


src/components/Alert/Alert.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from '../../packages/bem';
```
eszthoff: maybe use `import bem from 'bem'` not the full path

PR #27, opened 2018-12-13
Input component

carlobernardini:
* Adds Input component

### DIFF
--- a/src/components/Input/Input.js
+++ b/src/components/Input/Input.js
@@ -1,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from '../../packages/bem';
+import styles from './Input.scss';
+import { CONTEXTS, INPUT_TYPES, SIZES } from '../../constants';
+
+const { block } = bem({
+    name: 'Input',
+    classnames: styles,
+    propsToMods: ['context', 'isBlock', 'size']
+});
+
+const Input = props => {
+    const { children, context, disabled, isBlock, size, type, value, ...rest } = props;
+    return <input {...rest} {...block(props)} type={type} disabled={disabled} value={value} />;
+};
+
+// Any other attributes (onChange, onKeyUp etc.) are
+// supported although not defined in propTypes
+Input.propTypes = {
+    /** The input field context (e.g. brand, primary, bad, good etc. - defaults to brand) */
+    context: PropTypes.oneOf(CONTEXTS),
+    /** Should the input field be disabled or not */
+    disabled: PropTypes.bool,
+    /** Whether or not to show block-level input field (full width) */
+    isBlock: PropTypes.bool,
+    /** The size of the input field */
+    size: PropTypes.oneOf(SIZES),
+    /** Type of the input field */
+    type: PropTypes.oneOf(INPUT_TYPES),
+    /** Current input field value */
+    value: PropTypes.string
+};
+
+Input.defaultProps = {
+    context: 'brand',
+    disabled: false,
+    isBlock: false,
+    size: 'normal',
+    type: 'text',
+    value: undefined
+};
+
+export default Input;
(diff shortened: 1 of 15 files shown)


stories/Input.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import { boolean, select, text, withKnobs } from '@storybook/addon-knobs'; // eslint-disable-line import/no-extraneous-dependencies
+import Input from '../src/components/Input';
+import { CONTEXTS, INPUT_TYPES, SIZES } from '../src/constants';
+
+storiesOf('Input', module)
```
antipin: Would be nice to add a note like "Any other properties supplied will be spread to the underlined `<input />`  element"

src/constants/index.js
```diff
@@ -1 +1,13 @@
...
+    'text',
+    'password',
+    'email',
+    'date',
+    'datetime-local',
+    'month',
+    'tel',
+    'number',
+    'url'
+];
```
antipin: `search`, `time`, `week`, are missing

src/components/Input/Input.scss
```diff
@@ -0,0 +1,52 @@
...
+        size: var(--font-size-normal);
+        weight: var(--font-weight-normal);
+    }
+    outline: none;
+    padding: var(--spacing-2x);
+    transition: border-color var(--transition-duration);
+
+    &:focus {
+        border-color: var(--color-neutral);
+        box-shadow: 0 0 0 3px var(--color-neutral-25);
```
antipin: Oh my god! Dat is magic number! 🎃

src/components/Input/Input.scss
```diff
@@ -0,0 +1,52 @@
...
+        color: var(--color-neutral-75);
+        cursor: not-allowed;
+    }
+
+    &--context {
+        @each $context in (brand, primary, accent, info, good, warning, bad) {
+            &_#{ $context } {
+                &:focus {
+                    border-color: var(--color-#{ $context });
+                    box-shadow: 0 0 0 3px var(--color-#{ $context }-25);
```
antipin: another one magic

src/components/Input/Input.js
```diff
@@ -0,0 +1,45 @@
...
+    );
+};
+
+Input.propTypes = {
+    /** The input field context (e.g. brand, primary, bad, good etc. - defaults to brand) */
+    context: PropTypes.oneOf(CONTEXTS),
+    /** Should the input field be disabled or not */
+    disabled: PropTypes.bool,
+    /** Callback function on change */
+    onChange: PropTypes.func,
```
antipin: Trying to come up with some solid idea about which event handlers should be listed in docs, and which can be passed silently by `{ ... rest }`. To list them all is stupid, to list only some of them can be confusing (like we support only that one), list none of them... don't know, but it's consistent.
@carlobernardini @ribeiroguilherme @eszthoff what do you think?

stories/Input.js
carlobernardini: Can we maybe just add a code comment for now and add @storybook/addon-notes later in a new branch?

antipin: sure, just maybe worth to add a task about it in order not to miss it

src/components/Input/Input.js
carlobernardini: I think you're right. But if we don't define any of them, none of them will have any validation. Defining all of them isn't a good idea and would take a lot of work.
I can't think of a better solution than making it very clear through docs (as I suggest above we can also add @storybook/addon-notes for extra notes in Storybook)

antipin: > But if we don't define any of them, none of them will have any validation.

You mean propTypes validation? But do we really need to manage it? As far as we don't add any extra code around `onChange` callback, we don't need to validate it I guess. Or I don't get your point

carlobernardini: Yes I meant propTypes validation.

For the record, we decided not to explicitly define props like `onChange` or `onClick` but just propagate all of these through `{...rest}`. Will update `<Button />` as well.

PR #28, opened 2018-12-13
Add webpack uglify options

eszthoff:


antipin:
Reviewed and tested, works good, except `Button_Button_Button`, that is ok so far :)
Can be merged

PR #33, opened 2018-12-14
Checkbox component

eszthoff:


carlobernardini:
The checkbox label is unstyled. It should use the correct font family / size / line-height etc.

src/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,51 @@
...
+                <svg {...elem('svg', props)} width="12px" height="10px" viewBox={viewbox}>
+                    <polyline points="1.5 6 3.5 9 8 3" />
+                </svg>
+            </span>
+            <span {...elem('text', props)}>{children}</span>
+        </label>
+    );
+};
+
+Checkbox.propTypes = {
```
carlobernardini: This should probably have an `onChange` callback function, otherwise you cannot capture the checkstate

src/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,51 @@
...
+    /** A unique id to reference this checkbox */
+    id: PropTypes.string.isRequired,
+    /** If the checkbox should be disabled */
+    disabled: PropTypes.bool,
+    /** The label for the checkbox */
+    children: PropTypes.string.isRequired,
+    /** Viewbox dimentions, allows to costumise the checkbox */
+    viewbox: PropTypes.arrayOf(PropTypes.number)
```
carlobernardini: I suspect that `viewBox` only makes sense when the checkbox is scalable (in height / width). Maybe not needed for now after all.

src/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,51 @@
...
+            <span {...elem('text', props)}>{children}</span>
+        </label>
+    );
+};
+
+Checkbox.propTypes = {
+    /** A unique id to reference this checkbox */
+    id: PropTypes.string.isRequired,
+    /** If the checkbox should be disabled */
+    disabled: PropTypes.bool,
```
carlobernardini: Can we make the checkbox gray when disabled? Without label it's not clear that its state is disabled.

carlobernardini:
How can you affect the initial check state? I.e. what if I need it to be checked by default?

src/components/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,98 @@
+@import '../../themes/nice/nice.scss';
+
+.Checkbox {
+    margin: auto;
+    user-select: none;
+    cursor: pointer;
+    display: block;
+
+    &__text {
```
carlobernardini: Why not `&__label`?

src/components/Checkbox/Checkbox.js
carlobernardini: Now that I'm thinking about it once more, considering this structure any additional attributes from `...rest` would be applied to `<label />` while `onChange` should be applied to `<input />` AFAICT. Can you please double check the implementation so that we're sure that a callback can be fired (also including the checkstate as argument) when the checkstate has changed?

eszthoff: I added a story to demonstrate it

src/components/Checkbox/Checkbox.js
eszthoff: can you specify? which grey and where?

src/components/Checkbox/Checkbox.js
eszthoff: removed from props

src/components/Checkbox/Checkbox.scss
eszthoff: so not to confuse with the `<label>` element. This is a span. but I can change

eszthoff:
> How can you affect the initial check state? I.e. what if I need it to be checked by default?

just pass `checked` as a prop

carlobernardini:
@eszthoff the checkbox label is still missing styles (font family, size, weight)

carlobernardini:
screenshot 2018-12-18 at 16 04 05
@eszthoff Checkbox / label alignment is still a bit off. Do you prefer me to quickly fix styling?

PR #34, opened 2018-12-14
Tab components

carlobernardini:
* Adds component for tabs container (`<Tabs />`) and individual tabs (`<Tab />`)
* Tabs can be used controlled (manage active tab through props / `onChange` callback) _or_ stateful (active tab is managed within component)

Kept the implementation simple for now -- we can extend functionality and make it more flexible when necessary.


src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,90 @@
...
+                        if (!id) {
+                            return null;
+                        }
+                        const active = activeTabId === id;
+                        return (
+                            <a
+                                href={href || '#'}
+                                {...this.elem(active ? 'tabActive' : 'tab')}
+                                key={id}
+                                onClick={e => {
```
antipin: Can event handler  be extracted to a handleTabSwitch() method for better readability?

src/components/Tab/Tab.js
```diff
@@ -0,0 +1,37 @@
...
+    return (
+        <div {...rest} {...block(props)}>
+            {children}
+        </div>
+    );
+};
+
+Tab.propTypes = {
+    /** The tab content */
+    children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
antipin: does node assumes string as well?

src/components/Tab/Tab.js
```diff
@@ -0,0 +1,37 @@
...
+    /** URL / route for this tab */
+    href: PropTypes.string,
+    /** Unique ID for this tab */
+    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+    /** Label for this tab */
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired
```
antipin: node includes string

src/components/Tabs/Tabs.js
carlobernardini: Yes, makes sense to me

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,93 @@
...
+            <div {...rest} {...this.block()}>
+                <div {...this.elem('container')}>
+                    {Children.map(children, tab => {
+                        const { href, id } = tab.props;
+                        if (!id) {
+                            return null;
+                        }
+                        const active = activeTabId === id;
+                        return (
+                            <a
```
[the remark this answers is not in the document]
antipin: Also, I think we should render `<Tab>` here, not `<a>`, so we should change `Tab` component correspondently.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,93 @@
...
+                                href={href || '#'}
+                                {...this.elem(active ? 'tabActive' : 'tab')}
+                                key={id}
+                                onClick={e => {
```
antipin: I've also checked the MaterialUI API for tabs. Seems like it really make sense to remove some control from Tabs component and give to to a user (developer) to make it more flexible.

As you can see below they only care about tabs navigation bar, and it's up to a user to render tab's content anywhere he/she wants. I can imagine it can be really useful in a real-world product. For sure it makes usage of the component bit trickier, but we can always create something like `SimpleTabs` component based on this one if we need it.

image

carlobernardini: So if I get your point, you mean decoupling tabs and tab content but also adding one more wrapping component if you do want to use them next to each other (like it's now) Can you provide a use case / example for which this would be necessary (and cannot yet be done currently)? It feels to me like making it more flexible than we really need it to be..

src/components/Tabs/Tabs.js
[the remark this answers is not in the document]
carlobernardini: My reasoning to with `<a />` is because your tabs may be routed (e.g. each tab corresponds to a particular route, using React router for example). What element would you suggest it should render instead (even if it would be wrapped by `<Tab />`)?

src/components/Tabs/Tabs.scss
```diff
@@ -0,0 +1,68 @@
...
+            top: var(--spacing-2x);
+            left: var(--spacing-2x);
+            right: var(--spacing-2x);
+        }
+    }
+    &--justified {
+        .Tabs__container {
+            justify-content: space-between;
+        }
+    }
```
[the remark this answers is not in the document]
carlobernardini: Yes I added this prop initially but then chose to remove it because we could still add it once needed, but forgot to remove the corresponding class modifiers

src/components/Tabs/Tabs.js
carlobernardini: I just gave it a little try but it really is much more complicated and would need major overhauling. I think we need to discuss if this is something we really might need, and if so, we'll have to look at it together (it involves assigning matching ids between `Tab`'s and `TabContainer`'s, passing on callbacks back and forth, conditionally showing a certain `TabContainer` (it's not the only type of child within `SimpleTabs`) etc.

carlobernardini:
**Don't merge** - may be closed and replaced with #41  if it passes review

PR #35, opened 2018-12-14
Textarea component

carlobernardini:
* Adds TextArea component, styling is in line with already existing Input component


src/components/TextArea/TextArea.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from '../../packages/bem';
```
eszthoff: let just use `import bem from 'bem'` everywhere for consistency

src/components/TextArea/TextArea.js
```diff
@@ -0,0 +1,38 @@
...
+import { CONTEXTS, SIZES } from '../../constants';
+
+const { block } = bem({
+    name: 'TextArea',
+    classnames: styles,
+    propsToMods: ['context', 'isBlock', 'size']
+});
+
+const TextArea = props => {
+    const { context, disabled, isBlock, size, ...rest } = props;
```
eszthoff: did this pass linting? `context`, `isBlock` and `size` are never used. So no need to define them. Or is there?

stories/TextArea.js
```diff
@@ -0,0 +1,37 @@
...
+            }}
+            size={select('Size', SIZES, SIZES[1])}
+        />
+    ))
+    .add('Controlled component', () => (
+        <TextArea
+            context={select('Context', CONTEXTS, CONTEXTS[1])}
+            disabled={boolean('Disabled', false)}
+            isBlock={boolean('isBlock', false)}
+            onChange={e => {
```
eszthoff: The function here and in the other story are the same, but written differently. Consider writing them the same, so that it is clear that this is not the main point. Also might be good to list the props in the same order in both cases.

eszthoff:
I think you forgot to export it in `src/index.js`

PR #38, opened 2018-12-17
Loading spinner component

carlobernardini:
* Adds a LoadingSpinner component with only context and custom size props (to be extended when necessary)

### DIFF
--- a/src/components/LoadingSpinner/LoadingSpinner.js
+++ b/src/components/LoadingSpinner/LoadingSpinner.js
@@ -7,11 +7,11 @@ import { CONTEXTS } from '../../constants';
 const { block, elem } = bem({
     name: 'LoadingSpinner',
     classnames: styles,
-    propsToMods: ['context']
+    propsToMods: ['context', 'hidden']
 });
 
 const LoadingSpinner = props => {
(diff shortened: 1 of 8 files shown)


src/components/LoadingSpinner/LoadingSpinner.js
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import bem from '../../packages/bem';
```
eszthoff: let use `import bem from 'bem'` everywhere for sake of consistency

src/components/LoadingSpinner/LoadingSpinner.js
```diff
@@ -0,0 +1,69 @@
...
+                <Text inline {...elem('label', props)}>
+                    {label}
+                </Text>
+            )}
+        </div>
+    );
+};
+
+LoadingSpinner.propTypes = {
+    center: PropTypes.oneOf(['parent', 'viewport']),
```
eszthoff: add description for `center`. Also maybe call it `centerTo` or `centerContext` or something that suggests it sets the reference.

PR #40, opened 2018-12-18
Candidate avatar

carlobernardini:
* Adds candidate avatar (profile image) component with match indicator, to be used as atom of future result list item


src/components/CandidateAvatar/CandidateAvatar.js
```diff
@@ -0,0 +1,89 @@
...
+                }}
+            >
+                <circle
+                    r={radius}
+                    cx={fixedSize / 2}
+                    cy={fixedSize / 2}
+                    strokeWidth="4"
+                    {...elem('circle', {
+                        ...props,
+                        context: context(percentage)
```
carlobernardini: @antipin when looking at my console log I see the following warning: `BEM component "CandidateAvatar" prabably has incorrect declaration of "context"`. Doesn't this make sense? Do you see another way to compute the context on the basis of `matchPercentage`? Also, the warning has a typo ("prabably") ;)

src/components/CandidateAvatar/CandidateAvatar.js
```diff
@@ -0,0 +1,89 @@
...
+
+CandidateAvatar.propTypes = {
+    imageUrl: PropTypes.string,
+    matchPercentage: PropTypes.number,
+    showPercentageOnHover: PropTypes.bool,
+    size: PropTypes.number
+};
+
+CandidateAvatar.defaultProps = {
+    imageUrl: null,
```
eszthoff: I think we should add some default image not just leave it blank.

src/components/CandidateAvatar/CandidateAvatar.js
```diff
@@ -0,0 +1,89 @@
...
+                style={{
+                    width: fixedSize,
+                    height: fixedSize
+                }}
+            >
+                <circle
+                    r={radius}
+                    cx={fixedSize / 2}
+                    cy={fixedSize / 2}
+                    strokeWidth="4"
```
eszthoff: when the image is small, this looks really thick in comparison. could we set it conditionally based on image size?

src/components/CandidateAvatar/CandidateAvatar.scss
```diff
@@ -0,0 +1,70 @@
...
+        height: 72px;
+    }
+
+    &__circle {
+        fill: none;
+        stroke: var(--color-neutral);
+        stroke-dasharray: 0 999;
+        stroke-linecap: round;
+        shape-rendering: geometricPrecision;
+        transition: all .3s ease;
```
eszthoff: .3s should be a variable

src/components/CandidateAvatar/CandidateAvatar.scss
```diff
@@ -0,0 +1,70 @@
...
+    width: 72px;
+
+    &__ring {
+        width: 72px;
+        height: 72px;
+    }
+
+    &__circle {
+        fill: none;
+        stroke: var(--color-neutral);
```
eszthoff: this is never used. Why do you add it here?

src/components/CandidateAvatar/CandidateAvatar.scss
```diff
@@ -0,0 +1,70 @@
...
+                }
+            }
+        }
+    }
+
+    &__percentage {
+        background-color: rgba(0,0,0,.25);
+        border-radius: 50%;
+        color: white;
+        cursor: default;
```
eszthoff: what if the image is used as a link? This will overwrite the cursor. Maybe better not to set it explicitly?

src/components/CandidateAvatar/CandidateAvatar.scss
```diff
@@ -0,0 +1,70 @@
...
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        height: 100%;
+        width: 100%;
+        opacity: 0;
+        user-select: none;
+        font: {
+            family: var(--font-family-primary);
+            size: var(--font-size-large);
```
eszthoff: when the image is small, this is bigger than the image. Can it be set conditionally depending on the image size?

src/components/CandidateAvatar/__tests__/CandidateAvatar.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('<CandidateAvatar> that renders a candidate profile image with match indication', () => {
+    it('should render a default avatar', () => {
+        const wrapper = shallow(<CandidateAvatar />);
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+
+    it('should render a custom sized avatar with good match percentage', () => {
+        const wrapper = shallow(<CandidateAvatar size={128} percentage={67} />);
+        expect(toJson(wrapper)).toMatchSnapshot();
```
eszthoff: instead of just adding a snapshot, could you add a specific test if what to look for in the snapshot? like `wrapper.find('circle').hasClass('CandidateAvatar__circle--context_bad')`. That would make the test more future proof. 
The same for all the following tests.

src/components/CandidateAvatar/CandidateAvatar.js
carlobernardini: @antipin

src/components/CandidateAvatar/CandidateAvatar.js
carlobernardini: I've added a `stroke` prop that can take value `normal` (default) or `thin`. I don't think it should have more flexibility, because that can result in very ugly implementations. Making it conditional is quite complicated.

src/components/CandidateAvatar/CandidateAvatar.scss
carlobernardini: Removed it

src/components/CandidateAvatar/__tests__/CandidateAvatar.spec.js
carlobernardini: Done - but in some cases a snapshot really is more convenient: the circle `stroke-dasharray` computation is a bit complex and results in a number that you can't easily derive from `size`

src/components/CandidateAvatar/CandidateAvatar.scss
carlobernardini: Yep

src/components/CandidateAvatar/CandidateAvatar.scss
carlobernardini: Yeah... not really necessary indeed

src/components/CandidateAvatar/CandidateAvatar.js
carlobernardini: We haven't yet figured out how to deal with static assets.. So I'm not sure how to include an image without linking to an external host

eszthoff: good point. maybe can add to the statics ticket as a task to add an image here. so that it isn't forgotten.

src/components/CandidateAvatar/CandidateAvatar.scss
carlobernardini: Can we leave this for later (if necessary)? I would have to add a class on the basis of `size`, which I think would result in the same issue as https://github.com/textkernel/nice2/pull/40#discussion_r242445265

src/components/CandidateAvatar/CandidateAvatar.js
eszthoff: 2 values is ok. Maybe you can deduct them from the image size though, and not have it as a prop. Maybe more 'magic' but works more consistent. something like `const stroke = size < 50 ? thin : normal`

carlobernardini:
> Currently there is no way to disable the ranking, even if you don't define percentage. Should it not be an option?

Done - if `matchPercentage` is not provided (it's null), the circle and percentage label is not rendered.

src/components/CandidateAvatar/CandidateAvatar.js
carlobernardini: Ok, I'll make it respond to `size` and set the threshold on `60`. I also added something that will make `size` always an even number (internally) because otherwise the image would not be centered properly.

src/components/CandidateAvatar/CandidateAvatar.js
carlobernardini: Done

src/components/CandidateAvatar/CandidateAvatar.js
antipin: Hi there, I’ve investigated the issue.
In this case we have `context` modifier which doesn't depend neither of props nor of state.
It's kind of a new (but absolutely sensible) use-case for BEM library, which was not addressed so far. Good thing is that it works, Warning is just a warning and it is onited in production mode. So far just ignore it.

I created a ticket to address it https://trello.com/c/e5cNklRv/60-address-a-way-of-creating-modifiers-that-do-not-depend-neither-of-props-nor-of-state

PR #41, opened 2018-12-21
New implementation of tab components

carlobernardini:
* Adds component for integrated / managed tabs container (`<Tabs />`) and individual tabs (`<Tab />`)
* Adds components for individual elements so tab menu (`<TabMenu />`) / items (`<TabItem />`) / content (`<TabContent />`) can be used individually:
* Tabs can be used controlled (manage active tab through props / onChange callback) or stateful (active tab is managed within component)

Kept the implementation simple for now -- we can extend functionality and make it more flexible when necessary.


src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,95 @@
...
+
+class Tabs extends PureComponent {
+    constructor(props) {
+        super(props);
+
+        const { activeTabId } = props;
+
+        this.state = {
+            activeTabId,
+            derivedProp: activeTabId // eslint-disable-line react/no-unused-state
```
antipin: `derivedProp` sounds too abstract. Can we rename it to something like `derivedTabId` or store whole props object  as `prevProps`?

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+        const { activeTabId } = props;
+        return {
+            activeTabId,
+            derivedProp: activeTabId
+        };
+    }
+
+    handleTabChange(e, tabId) {
+        if (e && e.preventDefault) {
```
antipin: Can't understand why do we need this check. In what scenario `e` can be not an Event?

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,95 @@
...
+            return null;
+        }
+        const { activeTabId } = props;
+        return {
+            activeTabId,
+            derivedProp: activeTabId
+        };
+    }
+
+    handleTabChange(e, tabId) {
```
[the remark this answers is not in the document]
carlobernardini: As far as I can tell `e` is commonly used for events (besides catch-statements) ([evidence](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclick#JavaScript), [more evidence](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#Event_objects)) and `event` is a global name in IE. What would you suggest otherwise?

src/components/Tabs/Tabs.js
carlobernardini: Not really necessary indeed, but I reckoned we might as well check to be 100% sure. If you think it's better not to check I'll remove it.

src/components/Tabs/Tabs.js
carlobernardini: Makes sense, I will change it to `derivedTabId` as I don't want to duplicate anything else that we don't need

stories/Tabs.js
```diff
@@ -0,0 +1,57 @@
...
+        </Tabs>
+    ))
+    .add('Custom tabs', () => (
+        <div>
+            <TabMenu gutters={boolean('Show gutters', false)}>
+                <TabItem label="Some tab" isActive />
+                <TabItem label="Another tab" />
+            </TabMenu>
+            <div style={{ padding: 9 }}>Some other content here</div>
+            <TabContent>Some content</TabContent>
```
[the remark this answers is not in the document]
carlobernardini: Seems that that's because for some reason it sees `#` as a relative path and link clicks inside the iframe actually affect the main window location..? I suspect this is a Storybook issue: an `#` href should normally only affect the location hash and not make the browser navigate. With this implementation you're supposed to handle tab clicks yourself anyway.

src/components/Tabs/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './Tabs';
```
antipin: Here it looks like the only thing this component exposes to a developer is `Tabs`.
But in fact we also provide `Tab`, `TabMenu` and `TabItem`. I'd specify it here to make it more clear.

And furthermore, now (as far as we expose 4 components) it looks a bit confusing. 
To make it more clear I'd suggest the following structure (split TabsMenu and Tabs to two different components, so that TabsMenu is a "atom" and  Tabs is a "molecule"):

```
TabsMenu
├── TabItem
├── TabMenu.js
├── __tests__
└── index.js

Tabs
├── Tab
├── TabContent
├── Tabs.js
├── __tests__
└── index.js
```

stories/Tabs.js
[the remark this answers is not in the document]
carlobernardini: Updated

src/components/Tabs/Tabs.js
carlobernardini: Updated

src/components/Tabs/index.js
carlobernardini: I'm not really looking forward to reorganizing this structure (I'm short on time). I think it makes sense that `TabMenu` is housed under `Tabs`.. And everything you need is exported and listed in `src/index.js` anyway, right?

src/components/Tabs/Tabs.js
carlobernardini: Updated

src/components/Tabs/Tabs.js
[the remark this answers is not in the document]
carlobernardini: Updated

src/components/Tabs/index.js
antipin: File structure is indeed not a small change, but I think at least we should export everything that supposed to be exposed

carlobernardini: But... everything that needs to be exposed is exported? `src/index.js`

antipin:
Looks good!

PR #43, opened 2019-01-10
Radio button

eszthoff:


src/components/RadioButton/RadioButton.js
```diff
@@ -0,0 +1,52 @@
...
+                <Text {...elem('text', props)} inline muted={disabled}>
+                    {children}
+                </Text>
+            </label>
+        </div>
+    );
+};
+
+RadioButton.propTypes = {
+    /** A unique id to reference this radio button */
```
carlobernardini: Shouldn't it have a `name` prop? Radios are typically single-choice and need a `name` attribute for that

carlobernardini: Nvm, `...rest`...

src/components/RadioButton/RadioButton.scss
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+    }
+
+    &__svg {
+        position: absolute;
+        top: 0;
+        left: 0;
+        fill: var(--color-background);
+        stroke: var(--color-background);
```
carlobernardini: Is this needed?

stories/RadioButton.js
```diff
@@ -0,0 +1,25 @@
...
+import { boolean, text, withKnobs } from '@storybook/addon-knobs'; // eslint-disable-line import/no-extraneous-dependencies
+import RadioButton from '../src/components/RadioButton';
+
+storiesOf('RadioButton', module)
+    .addDecorator(withKnobs)
+    .add('RadioButton', () => (
+        <div>
+            <RadioButton
+                disabled={boolean('Disabled 1', false)}
+                id={text('Id 1', 'radio-1')}
```
carlobernardini: ID's probably don't need knobs (it's sufficient to see them in the code example, don't you think?)

src/components/RadioButton/RadioButton.js
```diff
@@ -0,0 +1,52 @@
...
+    classnames: styles,
+    propsToMods: ['disabled']
+});
+
+const RadioButton = props => {
+    const { id, children, viewbox, disabled, ...rest } = props;
+
+    return (
+        <div {...block(props)}>
+            <input type="radio" id={id} {...rest} {...elem('input', props)} disabled={disabled} />
```
carlobernardini: @eszthoff Now that I think of it, we may want `{...rest}` to come before the `type` attribute, otherwise you could for instance change input type to `text` which doesn't make sense for this component. Same for Checkbox.js, actually..

src/components/RadioButton/RadioButton.js
```diff
@@ -0,0 +1,52 @@
...
+                    <svg
+                        {...elem('svg', props)}
+                        width="12px"
+                        height="10px"
+                        viewBox={CHECKBOX_VIEWBOX}
+                    >
+                        <circle cx="5" cy="6" r="3" />
+                    </svg>
+                </span>
+                <Text {...elem('text', props)} inline muted={disabled}>
```
carlobernardini: We should probably only render this if `children` is not `null`, e.g.:

```
{ !!children && (
   <Text ... />
)}
```

(Same for Checkbox component, now that I think of it)

src/components/RadioButton/RadioButton.js
```diff
@@ -0,0 +1,52 @@
...
+import styles from './RadioButton.scss';
+import { CHECKBOX_VIEWBOX } from '../../constants';
+
+const { block, elem } = bem({
+    name: 'RadioButton',
+    classnames: styles,
+    propsToMods: ['disabled']
+});
+
+const RadioButton = props => {
```
carlobernardini: @eszthoff Ideally we should also start forwarding refs, at least to `input`'s and `button`'s

https://reactjs.org/docs/forwarding-refs.html

src/components/RadioButton/RadioButton.js
eszthoff: children isRequired via props. Otherwise you have no text at all, only a button/checkbox.

src/components/RadioButton/RadioButton.js
eszthoff: Let's discuss it together for all components, and see if this is a pattern we should make standard for all components we make. Then we can add it everywhere.
Thinking on it, we should have a developer's guide where we document all these standards, including how to use ...rest, and other things we agreed upon.

src/components/RadioButton/RadioButton.js
eszthoff: added

stories/RadioButton.js
eszthoff: removed it

src/components/RadioButton/RadioButton.js
eszthoff: moved it

src/components/RadioButton/RadioButton.scss
eszthoff: removed it

src/components/RadioButton/RadioButton.js
carlobernardini: At this point I don't exactly remember the use case I had in mind but I think you should be able to render just a checkbox / radio, without requiring it to come with a label

As an example, consider this use case in the maintenance ui:
screenshot 2019-01-28 at 17 05 00

src/components/RadioButton/RadioButton.js
carlobernardini: Ok, let's keep this for later so that we can apply it to all components for which it makes sense.

src/components/RadioButton/RadioButton.js
eszthoff: would allowing for children to be `node` or `string` but still required make sense. Then if it is a string, we make it a `Text` as now, or render it without change if it is a node.

eszthoff:
@carlobernardini could you resolve all discussions that you are satisfied with? Gives and easier overview. Thanks

src/components/RadioButton/RadioButton.js
carlobernardini: I'm still not sure why any children would be required. Native HTML radios don't require any label  either: http://jsfiddle.net/7joxtrzn/

eszthoff: I think it is strange to allow for that. But whatever

carlobernardini:
All good

PR #44, opened 2019-01-25
Change library name

carlobernardini:
* Changes all references from Nice to OneUI following library name change


example/src/App.js
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { Heading, Button } from 'nice2'; // eslint-disable-line import/no-unresolved, import/no-extraneous-dependencies
+import { Heading, Button } from 'oneui'; // eslint-disable-line import/no-unresolved, import/no-extraneous-dependencies
```
antipin: Shouldn't it be `from '@textkernel/oneui';`

example/package.json
```diff
@@ -12,6 +12,6 @@
         "build": "react-scripts build",
         "test": "react-scripts test",
         "eject": "react-scripts eject",
-        "link:nice2": "cd .. && npm link && cd ./example && npm link nice2"
+        "link:oneui": "cd .. && npm link && cd ./example && npm link oneui"
```
antipin: `npm link @textkernel/oneui`?

example/src/App.js
carlobernardini: Yes, you're right. I'll update it. Also, I noticed that the dist files are now capitalized (e.g. OneUI.js, OneUI.css) - we probably don't want that

PR #46, opened 2019-01-28
Add prepublish script, make sure only necessary files are packaged

carlobernardini:
* Adds ~~`prepublishOnly`~~ [`prepare`](https://docs.npmjs.com/misc/scripts#description) script: will run before package is prepared and packed (`npm publish`) and on `npm install` (also when installing from Github). It will run tests and produce a fresh production build (or make `npm publish` fail in case of any errors).
* White listed resources to be included in package, considering recommendations [in this article](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d). ~~Still added a `.npmignore` for certain common files.~~
* Rename compiled CSS (`dist/oneui.css`) to `oneui.min.css`


package.json
```diff
@@ -14,8 +14,12 @@
         "lint": "npm run lint:ci -- --fix",
         "lint:ci": "eslint .",
         "lint-styles": "stylelint 'src/**/*.s(a|c)ss'",
-        "test": "jest"
+        "test": "jest",
+        "prepublishOnly": "npm test && npm run build:prod"
```
eszthoff: shouldn't check eslint and sasslint too?

carlobernardini: Fair point, added (while at it, I renamed `lint-styles` to `lint:styles` for consistency)

carlobernardini:
Verified package content:

```
$ npm pack --dry-run
textkernel-oneui-0.0.1.tgz
$ tar -tvf textkernel-oneui-0.0.1.tgz 
-rw-r--r--  0 502    20       3346 Jan 28 20:43 package/package.json
-rw-r--r--  0 502    20       2516 Jan 28 20:40 package/README.md
-rw-r--r--  0 502    20       1070 Oct  3 20:17 package/LICENSE
-rw-r--r--  0 502    20      59102 Jan 28 20:32 package/dist/oneui.min.js
-rw-r--r--  0 502    20      21235 Jan 28 20:32 package/dist/oneui.css
```

.npmignore
```diff
@@ -0,0 +1,173 @@
+### VisualStudioCode template
```
eszthoff: following the article you quoted in the PR description, this list should be added to .gitignore not .npmignore. Having said that, I would try to reduce this to something reasonable and not list things we don't use for sure (nyc for example)

carlobernardini: Yes, I think you're right. I already ran into an issue where some dir / file was mistakenly ignored by such a all-encompassing gitignore / npmignore template. Shall we just remove this npmignore and keep gitignore as is? Or do you see something we should definitely add to gitignore right away?

PR #48, opened 2019-01-29
Make CandidateAvatar percentage scale to fit

carlobernardini:
* Switched to use SVG text to make percentage scale to fit parent container, no matter what size.


ribeiroguilherme:
Reviewed. looks good!

PR #49, opened 2019-01-30
Icon component

carlobernardini:
* Added component to render vector icon for each main TK product, with support for custom size, context and SVG title ([reference](https://css-tricks.com/creating-svg-icon-system-react/))
* Scale and align icon with text by default; absolute dimensions can be applied using `size` prop (pixels, 1:1 aspect ratio)
* Alignment strategy based on [this article](https://blog.prototypr.io/align-svg-icons-to-text-and-say-goodbye-to-font-icons-d44b3d7b26b4)


src/components/Icon/Icon.js
```diff
@@ -0,0 +1,69 @@
...
+const { block, elem } = bem({
+    name: 'Icon',
+    classnames: styles,
+    propsToMods: ['context', 'margin']
+});
+
+const Icon = props => {
+    const { context, name, size, title, ...rest } = props;
+
+    const currentIcon = ICONS[name];
```
antipin: I'd suggest to implement this bit differently (see below why):

### 1. To create a base IconBase
Component that is basically just creates an <svg> element with all the boring default <svg> attrs like:
```js
<svg
  xmlns="http://www.w3.org/2000/svg"
  stroke="currentColor"
  fill="currentColor"
  strokeWidth="0"
  ... // all common svg stuff
>
  {children}
</svg>
```

### 2. To create a IconLogo as a common wrapper for all logo icons

```js
import IconBase from './IconBase';

...

return (
        <IconBase
            {...elem('svg', props)}>
            viewBox={currentIcon.viewBox}
            ... // all common logo icon stuff
        >
            {!!title && <title>{title}</title>}
            {children}
        </svg>
    </div>
);
```

### 3. And for each product logo create actual component

**./src/components/logos/ExtractLogo/ExtractLogo.js**
```js
import IconLogo from './IconLogo';

return (
  <IconLogo viewBox="0 0 77.36 93.32">
    <path d="M5.17,87.36l-.46,4c0,1,2.14,1.91,4.71,1.91H67.93c2.58,0,4.71-.9,4.71-1.91l-.45-4Z" />
    <path d="M2.81,78.6l-.45,3.71c0,1.12,2.25,2,5.05,2H70c2.81,0,5-.9,5-2l-.56-3.71Z'" />
    <path d="M75.35,57.15C75.35,55.91,73,55,70,55H55v3.81a3.71,3.71,0,1,1-7.42,0V55H28.75v3.81a3.71,3.71,0,1,1-7.42,0V55H7.3c-2.91,0-5.39,1-5.39,2.13L0,73.55c0,1.24,2.36,2.13,5.39,2.13H72c2.91,0,5.38-1,5.38-2.13Z" />
     ...
  </IconLogo>
);
```

## Why

1. Keeping all the icons as constants in  **src/constants/icons.js** will force us (users of oneUI) to load all icons in a bundle when we need only one of them.
2. `<svg>` element can contain not only `<path>` but a lot of other tags like `<circle>` for instance.
3. `<IconBase/>` can be nicely reused later anytime we want to use svg

src/components/Icon/Icon.scss
```diff
@@ -0,0 +1,31 @@
...
+        height: 1em;
+    }
+
+    &__path {
+        transition: fill var(--transition-duration);
+
+        &--context {
+            @each $context in (neutral, brand, primary, accent, info, good, warning, bad) {
+                &_#{ $context } {
+                    fill: var(--color-#{ $context });
```
antipin: Color can be handled by top level component and here we can use `fill: currentColor` (or even nothing is needed to inherit fill value from a parent)

src/components/Icon/Icon.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+        <div {...rest} {...block(props)}>
+            <svg
+                {...elem('svg', props)}
+                xmlns="http://www.w3.org/2000/svg"
+                viewBox={currentIcon.viewBox}
+                aria-labelledby="title"
+                style={(() => {
```
antipin: Inline functions inside a render method triggers unnecessaery rerenderings as far as they being recreatred on each render. Better to extract it to something like `function calculateSizes(size) { ... }`.

And also it makes code herder to read.

src/components/Icon/Icon.js
carlobernardini: Ok, I think that all makes sense. But do you really think we also need `IconLogo`? It seems sort of useless to me. The outer div as well as the optional title can also be part of `IconBase`, right?

Then, in terms of naming, I would maybe go with `Icon` and `IconProduct`

src/components/Icon/Icon.scss
carlobernardini: If you can apply `fill` on the SVG instead of each path while preserving the current aesthetics then it makes sense indeed

src/components/Icon/Icon.js
carlobernardini: Would you split it off into a separate util or have it as a class method?

antipin: I'd just declare a function before in the same file
```
const Icon = props => {
```

src/components/Icon/Icon.js
antipin: Having `IconBase` may be indeed too much for just `<svg/>`.

In terms of naming I think make sense to name base icon as `ProductIcon` or `ProductLogo` or just `Logo`. Point here is that just `Icon` is too abstract and in this case we deal only with our logos.

So I'd go with something like `Logo` and then bunch of `SearchLogo`, `ExtractorLogo`, `JobfeedLogo` etc...

carlobernardini: ...apparently my comment was deleted because I didn't finish the 'review' 🙄 

Anyway, so I would then suggest to have `IconBase` which is a combination of what you first suggested to be `IconBase` and `IconLogo` (don't see a reason for those to be two different components), and then a component for each distinct icon: `IconSearch`, `IconJobfeed` etc.

The tree structure could be something like this:

```
└── components
    └── Icon
        ├── IconBase
        │   ├── IconBase.js
        │   ├── IconBase.scss
        │   └── tests
        ├── IconSearch
        │   ├── IconSearch.js
        │   ├── IconSearch.scss
        │   └── tests
        └── IconJobfeed
            ├── IconJobfeed.js
            ├── IconJobfeed.scss
            └── tests
```
Makes sense?

antipin: Yes, looks good!
Except one thing, as I said above, `Icon` is a too generic name as for me.
Put yourself in a shoes of a new developer who come to our project and sees this:
```
├── components
│   ├── Alert
│   ├── Button
│   ├── ...
│   ├── Icon
│   ├── Link
```

No way that he/she will guess that it is about product logos.

carlobernardini: But it does not necessarily have to be only product icons, right? Can be any custom icon of ours, who knows.. Although for UI purposes we will probably be using another package like FA5 Pro.

antipin: Exactly, something like https://github.com/react-icons/react-icons would make more sense as a icon pack instead of having it in the OneUI lib.

PR #50, opened 2019-02-01
Deploy Storybook to Github Pages

carlobernardini:
**NOTE**
~~Do not merge before [`add-release-presequisites`](https://github.com/textkernel/oneui/pull/46)~~ Ignore - PR has been merged.

Running `npm run storybook:deploy` will build and deploy Storybook from current `master` to [Github Pages](https://textkernel.github.io/oneui/)

* Added `@storybook/storybook-deployer` and reconfigured `package.json` to include `storybook:deploy` script
* Added `displayName` to all components because component names would get minified (not applied to Dummy component as it is to be removed in another branch)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "main": "dist/oneui.min.js",
     "scripts": {
         "storybook": "start-storybook -s ./stories/static -p 9001 -c .storybook",
-        "storybook:build": "build-storybook -s ./stories/static -c .storybook",
+        "storybook:build": "build-storybook -s ./stories/static -c .storybook -o .out",
         "storybook:deploy": "storybook-to-ghpages --script=storybook:build --out=.out",
         "build": "webpack --config ./etc/webpack/webpack.dev.config.js",
         "build:prod": "webpack --config ./etc/webpack/webpack.prod.config.js",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "main": "dist/oneui.min.js",
     "scripts": {
         "storybook": "start-storybook -s ./stories/static -p 9001 -c .storybook",
-        "storybook:prod": "build-storybook -s ./stories/static -c .storybook",
+        "storybook:build": "build-storybook -s ./stories/static -c .storybook",
+        "storybook:deploy": "storybook-to-ghpages --script=storybook:build --out=.out",
         "build": "webpack --config ./etc/webpack/webpack.dev.config.js",
         "build:prod": "webpack --config ./etc/webpack/webpack.prod.config.js",
         "lint": "npm run lint:ci -- --fix",
@@ -31,7 +32,7 @@
         "type": "git",
         "url": "git+https://github.com/textkernel/oneui.git"
     },
-    "author": "Textkernel OneUI People <open-source@textkernel.nl>",
+    "author": "Textkernel B.V. <open-source@textkernel.nl>",
     "license": "MIT",
     "bugs": {
         "url": "https://github.com/textkernel/oneui/issues"
@@ -55,6 +56,7 @@
         "@storybook/addon-info": "4.0.8",
         "@storybook/addon-knobs": "4.0.12",
         "@storybook/react": "4.0.8",
+        "@storybook/storybook-deployer": "2.8.1",
         "autoprefixer": "9.1.5",
         "babel-core": "7.0.0-bridge.0",
         "babel-eslint": "10.0.0",
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 ### OneUI project
 /dist
 /.cache
+/.out
 
 ### VisualStudioCode template
 /.vscode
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -8,7 +8,8 @@
     "main": "dist/oneui.min.js",
     "scripts": {
         "storybook": "start-storybook -s ./stories/static -p 9001 -c .storybook",
-        "storybook:prod": "build-storybook -s ./stories/static -c .storybook",
+        "storybook:build": "build-storybook -s ./stories/static -c .storybook",
+        "storybook:deploy": "storybook-to-ghpages --script=storybook:build --out=.out",
```
ribeiroguilherme: Now we have two generated folders for the storybook static files.: `.out` created by `storybook:deploy`, and `storybook-static` created by the `storybook:build` command. Perhaps make sense to configure them to output in the same folder

package.json
```diff
@@ -8,7 +8,8 @@
...
         "build": "webpack --config ./etc/webpack/webpack.dev.config.js",
         "build:prod": "webpack --config ./etc/webpack/webpack.prod.config.js",
         "lint": "npm run lint:ci -- --fix",
```
ribeiroguilherme: Not related to the changes here. But now when I only want to install the library dependencies to work locally (like now, I had to install the `storybook-to-ghpages`), I've to run `npm install` as usual. Although it fires `prepare` script, that run the tasks that I don't want to. Is this the desired behavior?

package.json
ribeiroguilherme: error

I can't deploy it. Am I missing something? Did you get this too?

carlobernardini: Where do you see `storybook-static`? If you mean `./stories/static`, that's the directory where to load static files from (e.g. avatar image)

carlobernardini:
> Do you think is worth it to set up a CD every time when we merge to master? So we won't need to deploy manually.

Ideally yes, in such way that it gets deployed each time `master` is updated like you say. I'm not sure how to set it up that way, and maybe it can even be done in a separate ticket.

package.json
ribeiroguilherme: If you run `storybook:build` you can see

carlobernardini: I just checked and when running `storybook:deploy` I don't see any `storybook-static` directory. Maybe it's a leftover from another branch?

carlobernardini: Seems to be caused by branch protection, now that I've turned that off for `gh-pages` it seems to work again

package.json
ribeiroguilherme: I see.. what is the reason? do we have to do this because the package isn't published yet to `npm`?

package.json
ribeiroguilherme: Now it worked!

ribeiroguilherme:
> > Do you think is worth it to set up a CD every time when we merge to master? So we won't need to deploy manually.
> 
> Ideally yes, in such way that it gets deployed each time `master` is updated like you say. I'm not sure how to set it up that way, and maybe it can even be done in a separate ticket.

Let's do it in another ticket then

package.json
carlobernardini: My reasoning was so that you can install it as a package from whichever Github branch you want, before it is versioned and released to NPM. E.g. for tryout purposes (e.g. you're developing a new application feature which relies on a new component in OneUI).

package.json
ribeiroguilherme: Run only `storybook:build`. Doesn't it create the static files in the `storybook-static` folder ?

carlobernardini: Done

package.json
ribeiroguilherme: I see your point.. but why not publish to `npm` as a beta version then? I don't think it is a bad idea. Would be also useful to use in the proof-of-concept web app.

I don't see us, as developers, using components from branches, even to try it out. We have storybook for that. And before implementing components, we also have this design process (UI/UX, prototyping) for them to be aligned with the company UI standards. So I'd assume that there is no need to tryout.

carlobernardini: Yes but you need to be able to try out your new component in some other application, I reckon. To make sure that it fits the use case and integrates as expected. Otherwise you would have to wait until a new version is released, or make releases much more often, possibly while still needing some changes to your new components. See what I mean?

ribeiroguilherme: To avoid this we have the whole design process (UX/UI prototyping & components api specification) before implementing a component

antipin: If I understood correctly, the problems are:

* We don't want to run unnecessary `build` if we just run `npm install` (see N.B. here) locally in order to update deps
* But we need to be able to test our non-published branches with our products
* So we do want to be able to install package from github repo for testing and therefore we need to automatically perform the build task after package installation

For testing purposes I don't actually think it's a good idea to either rely on fetching from npm or pulling from github's branch. Both options would make development/testing process really annoying:
* You make a small change in OneUI
* You commit it and publish/push it
* You reinstall deps for your project to fetch new version
* Ta-daam, you hate your job...

Might be i thinking too simple and missing something, but why `npm link` couldn't solve this use case? In this case flow would look like:
* You make a small change in OneUI
* You instantly have it in your project

@carlobernardini @ribeiroguilherme  What do you think?

antipin: One more thing: please use `npm ci` instead of `npm install`. `npm ci` relies only on a `package-lock.json` => produces stable build and **doesn't change it**. Use npm install only when you adding a new module to the project (updating deps).

carlobernardini: Ok, since this change is not in scope of this PR and it's now already there in `master` I would suggest to create a separate ticket to reconsider the implementation and discuss it with our next meeting, considering all the comments above. Then we can still continue merging this branch - it doesn't affect anything related to prepare / prepublishOnly etc.

PR #51, opened 2019-02-02
Set up Coveralls reporting

carlobernardini:
* Run CI with coverage reporting
* Integrate with [coveralls.io](https://coveralls.io)
* Add coverage badge to README.md
* Add dependencies badge to README.md

Note that the badge will say 'coverage unknown' until this has been merged with `master`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
 install:
   - npm ci
 script:
-  - npm run test -- --coverage
+  - npm run test:coverage
   - npm run build
   - npm run build:prod
   - npm run lint:ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,10 @@ cache:
 install:
   - npm ci
 script:
-  - npm run test
+  - npm run test -- --coverage
   - npm run build
   - npm run build:prod
   - npm run lint:ci
   - npm run lint:styles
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -7,8 +7,10 @@ cache:
 install:
   - npm ci
 script:
-  - npm run test
+  - npm run test -- --coverage
```
ribeiroguilherme: Doesn't make sense to add `--coverage` to `npm run test` locally as well? So we can run tests and see what lines are not covered

carlobernardini: I wouldn't do that by default because it's quite a bit slower than running tests without coverage. You can still run `npm test -- --coverage` yourself to get coverage reporting.

ribeiroguilherme: In terms of time, doesn't slow down that much. I got from 4s to 4.4s here. But indeed there is no need to add when we run `test`.

I'd add as a separate script `test:coverage` so we can recall ourselves to use during tests implementation if we want to keep the codebase well covered. 

And in case we want to run tests fast, use the normal `test` script, or even add a `test:watch` to watch file changes.

carlobernardini: Alright, that makes sense. I'll add `test:coverage`, watch is not really in this scope.

PR #52, opened 2019-02-04
Enable Storybook action logger

carlobernardini:
* Added `@storybook/addon-actions`, `@storybook/addon-console` to enable logging any console events to action logger

### DIFF
--- a/stories/Alert.js
+++ b/stories/Alert.js
@@ -16,7 +16,6 @@ storiesOf('Alert', module)
                         ? {
                               label: 'Dismiss',
                               onClick: e => {
-                                  // eslint-disable-next-line no-console
                                   console.log('Clicked alert action', e);
                               }
                           }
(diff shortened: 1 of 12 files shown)


stories/Alert.js
```diff
@@ -15,9 +15,9 @@ storiesOf('Alert', module)
                     hasAction
                         ? {
                               label: 'Dismiss',
-                              onClick: () => {
-                                  // eslint-disable-next-line no-alert
-                                  window.alert('Alright');
+                              onClick: e => {
+                                  // eslint-disable-next-line no-console
+                                  console.log('Clicked alert action', e);
```
antipin: Seems like using `console.log`in ./stories directory is a valid thing, so we can just set a specific rules for this folder by creating the following file:

**./stories/.eslintrc.json**
```json
{
    "overrides": [{
        "files": [ "*.js" ],
        "rules": {
            "no-console": 0
        }
    }]
}
```

carlobernardini: Yes that makes more sense, will do

PR #57, opened 2019-02-05
Add ability to define "wildcard" class name for modifier

antipin:
Applied if modifier has any value

### DIFF
--- a/src/components/Alert/__tests__/__snapshots__/Alert.spec.js.snap
+++ b/src/components/Alert/__tests__/__snapshots__/Alert.spec.js.snap
@@ -2,13 +2,13 @@
 
 exports[`<Alert> that renders an alert should add classes when props are changed 1`] = `
 <div
-  className="Alert Alert--context_bad"
+  className="Alert Alert--context Alert--context_bad"
 >
   <div
-    className="Alert__content Alert__content--context_bad"
+    className="Alert__content Alert__content--context Alert__content--context_bad"
   >
     <strong
-      className="Alert__title Alert__title--context_bad"
+      className="Alert__title Alert__title--context Alert__title--context_bad"
     >
       Hey there
     </strong>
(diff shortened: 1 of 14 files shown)


src/components/Alert/__tests__/__snapshots__/Alert.spec.js.snap
```diff
@@ -2,13 +2,13 @@
...
-  className="Alert Alert--context_bad"
+  className="Alert Alert--context Alert--context_bad"
 >
   <div
-    className="Alert__content Alert__content--context_bad"
+    className="Alert__content Alert__content--context Alert__content--context_bad"
   >
     <strong
-      className="Alert__title Alert__title--context_bad"
+      className="Alert__title Alert__title--context Alert__title--context_bad"
```
ribeiroguilherme: I noticed some classnames that are not used at all, they don't exist. Ex: `Alert__title--context` , `Alert__content--context`. Is it the desired behavior?

src/packages/bem/__tests__/bem.spec.js
```diff
@@ -117,6 +127,16 @@ describe('BEM decorator', () => {
             expect(buttonLabel.hasClass('ButtonStateless__label--size_2')).toBe(true);
         });
 
+        it('should add "wildcard" class names for modifier with value (striong or number)', () => {
```
ribeiroguilherme: `string` :)

src/packages/bem/__tests__/bem.spec.js
```diff
@@ -37,6 +37,16 @@ describe('BEM decorator', () => {
             expect(buttonLabel.hasClass('Button__label--size_2')).toBe(true);
         });
 
+        it('should add "wildcard" class names for modifier with value (striong or number)', () => {
```
ribeiroguilherme: might be good to validate that the classname with the value is added too in case if it is defined? ex: `Button__icon--size-2`

src/components/Alert/__tests__/__snapshots__/Alert.spec.js.snap
antipin: I also noticed it. That's because this `identity-obj-proxy`thing that we use in jset to mock css imports.

I created a ticket for it here: https://trello.com/c/1J8lsZsE/78-make-jest-use-real-style-declarations-instead-of-mocked-one

src/packages/bem/__tests__/bem.spec.js
antipin: This behavior is covered in the previous test case "should add proper class names based on propsToMods and passed props"

src/packages/bem/__tests__/bem.spec.js
antipin: Fixed

PR #58, opened 2019-02-05
Improve code coverage

carlobernardini:
Slightly improved codecov for Tabs and CandidateAvatar components

### DIFF
--- a/src/components/Tabs/__tests__/Tabs.spec.js
+++ b/src/components/Tabs/__tests__/Tabs.spec.js
@@ -19,8 +19,10 @@ describe('<Tabs> that renders tab container with some tabs', () => {
     });
 
     it('should switch tabs correctly', () => {
+        const onChange = jest.fn();
+
         const wrapper = mount(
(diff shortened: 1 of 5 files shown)


src/components/Tabs/__tests__/Tabs.spec.js
```diff
@@ -35,5 +37,28 @@ describe('<Tabs> that renders tab container with some tabs', () => {
...
+                </Tab>
+            </Tabs>
+        );
+        wrapper.setProps(
+            {
+                activeTabId: '2'
+            },
+            () => {
+                expect(wrapper.state().activeTabId).toBe('2');
+                expect(wrapper.state().derivedTabId).toBe('2');
```
eszthoff: Please add a check to see that the correct content is displayed. Maybe a snapshot test, or something more explicit,  like adding a paragraph to the content that you can search the wrapper for.

carlobernardini: Good idea, done

PR #59, opened 2019-02-06
Reimplementation of Icon components

carlobernardini:


antipin:
Tested and reviewed, looks good for me.
Should we create a ticket to address the problem with copy-pasting similar prop types?

PR #63, opened 2019-02-07
feat(theming): oneui themeloader

ribeiroguilherme:
To apply an theme, provide the URL param `themeId=theme-custom.css`


example/public/index.html
```diff
@@ -1,49 +1,24 @@
...
+        <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+        <link href="https://fonts.googleapis.com/css?family=Charm" rel="stylesheet">
+
+        <!--
+            [ REQUIREMENT ]
+            Loads the polyfills according to the user browser. Used mainly to run
+            this demo on IE11
+        -->
+        <script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
```
antipin: So this one is not something mandatory for a project? Just for demo purposes?

example/src/ThemeLoader.js
```diff
@@ -0,0 +1,68 @@
...
+
+        cssVarsPonyfill({
+            onlyLegacy: false, // False, so we can use it on Chrome for test
+            onlyVars: true,
+            watch: true,
+            onBeforeSend: function (xhr, node, url) {
+                console.log('onBeforeSend', url);
+            },
+            onSuccess: function (cssText, node, url) {
+                console.log('onSuccess', url);
```
antipin: The promise from `static load()` should be resolved at this point to make it work

example/src/ThemeLoader.js
```diff
@@ -0,0 +1,70 @@
...
+
+                queryString.split('&').forEach((item) => {
+
+                    const queryParam = item.split('=');
+
+                    if (queryParam[0] === THEME_URL_PARAM) {
+                        const styleElement= document.createElement('link');
+                        styleElement.type = 'text/css';
+                        styleElement.rel = 'stylesheet';
+                        styleElement.href = `${THEME_CDN_URL}${queryParam[1]}`;
```
antipin: We also should detect the moment when this style is loaded.
See this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Stylesheet_load_events

example/src/ThemeLoader.js
```diff
@@ -0,0 +1,70 @@
...
+
+            startCssVarsPonyfill();
+            applyThemeIfExists();
+
+        }));
+    }
+}
+
+
+export default ThemeLoader;
```
antipin: Ok, now it's better, but still some room for improvement:

1. I'd leave `startCssVarsPonyfill()` and `applyThemeIfExists()` methods as static in order not to put everything inside the `load()`
2. Both methods should return a `Promise`
3. Then load would look like:
```js
static load() {
        return Promise.all([
            ThemeLoader.applyThemeIfExists(),
            ThemeLoader.startCssVarsPonyfill(),
        ])
}
```

P.S. Both methods should do proper error handling (call `reject` in case of error, not `console.log`)

antipin: Also probably, `ThemeLoader.applyThemeIfExists()` and `ThemeLoader.startCssVarsPonyfill()` should be called in a sequence one after another. Need to be testes

example/public/index.html
ribeiroguilherme: Is mandatory to run the demo on IE, since it doesn't support Promises. But in modern browsers it's not required.

example/src/ThemeLoader.js
ribeiroguilherme: I tested here and seems that the order doesn't affect the parsing

example/src/ThemeLoader.js
ribeiroguilherme: I think I got it. Pushed the changes

example/public/index.html
carlobernardini: So would it still be loaded even if you use a modern browser?

example/src/ThemeLoader.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+
+    /**
+     * Loads the CSS ponyfill.
+     *
+     * In case the property `onlyLegacy` is set to true, it will run only in legacy browsers like IE11.
+     * For test purposes, it's set to true, so can be tested on modern browsers.
+     */
+    static startCssVarsPonyfill() {
```
carlobernardini: What if you're not using the polyfill, or it's not loaded because you're on a modern browser?

example/src/index.js
```diff
@@ -1,5 +1,13 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import ThemeLoader from './ThemeLoader';
+
+
+import '@textkernel/oneui/dist/oneui.css';
```
carlobernardini: Is it also possible to import the source in your own css and compile it as part of the application build process? Or can you only use the precompiled version like this?

example/public/index.html
antipin: This exact polyfill actually is pretty smart. It detects a browser version by a user agent string and returns only necessary javascript:

image

But I believe that deal with polyfills is a responsibility of an app/product. We just need to declare required tech we rely on.

example/src/ThemeLoader.js
antipin: It has the flag `{ onlyLegacy: true  | false }` so it's not doing anything if a browser support css vars

example/src/index.js
antipin: But that's exactly what is demonstrated here, it's it?
`example/src/index.js` is your application and you just importing `'@textkernel/oneui/dist/oneui.css';` in it.

carlobernardini: What I mean is your application has a Scss of its own and you import the OneUI scss in it: `@import '@textkernel/oneui/src/themes/oneui/oneui.scss'`. So everything (your application's CSS + OneUI CSS) can be bundled into 1.

carlobernardini:
In general a design for this implementation wouldn't have hurt

example/src/index.js
antipin: I really think that importing sass from OneUI is something we shouldn't encourage/support.
Otherwise any variable we have in sass becomes oneUI's "API" and therefore we should support it and maintain backwards compatibility etc.

ribeiroguilherme: As long as we expose the `vars` to the user, I don't see any use case for this scenario. Might be good indeed to consider it later on when gets necessary.

carlobernardini: Ok, if you have arguments against it, it's fine -- just wanted to make sure it was considered

carlobernardini:
> To apply an theme, provide the URL param `themeId=theme-custom.css`

How / where would the URL parameter need to be configured?

antipin:
> How / where would the URL parameter need to be configured?

@carlobernardini that's a good point. I think it's out of the scope of the `ThemeLoader.js` library.
`ThemeLoader.js` should only accept `themeId` and where it comes from is up to an app (can be url query param, some fixed configuration or whatever.. personal theme of a user from a local storage)

carlobernardini:
> > How / where would the URL parameter need to be configured?
> 
> @carlobernardini that's a good point. I think it's out of the scope of the `ThemeLoader.js` library.
> `ThemeLoader.js` should only accept `themeId`. And where it comes from is up to an app (can be url query param, some fixed configuration or whatever.. personal theme of a user from a local storage)

I don't see how it is out of scope of this implementation if it's an essential part of the problem. So how exactly can you tell ThemeLoader which theme ID it's supposed to use -- regardless of where that id comes from?

antipin:
> So how exactly can you tell ThemeLoader which theme ID it's supposed to use -- regardless of where that id comes from?

As I wrote above `ThemeLoader.js` should only accept `themeId`

Like this:
```js
ThemeLoader.load('SOME_THEME_ID')
   .then(() => ...)
```

Because now `ThemeLoader` relies on URL query param and tries to get a `themeId` from there.

carlobernardini:
I noticed that `ThemeLoader.js` is now part of the example app and not of OneUI. Shouldn't we ship this utility with OneUI? And it looks like its responsibility is not only loading a theme, but also the main CSS and polyfill, so shouldn't it's class rather be named something like `OneUI` with an `init` method? E.g. `OneUI.init({ theme: 'THEME_ID' }).then(() => ...)`

ribeiroguilherme:
Makes sense to pass the` themeId` as a argument, so the class doesn't need to care how to get the theme ID param.

Carlo, so what basically ThemeLoader would do? Attach an HTML element with the theme to the DOM, and that is it?

If this is the case, I imagine the scenario that the application would start like:


```
Promise.all([
   OneUI.init({  theme: 'THEME_ID' }), 
   CssVarsPolyfill.init()
])
.then(() => ReactDOM.render....)
```

antipin:
@ribeiroguilherme in the end I also see it like @carlobernardini suggested:
For application that uses OneUI it should be like:
```js
import { ThemeLoader } from 'oneui'; // or you name it

ThemeLoader.init(SOME_THEME_ID).then(
  () => render(...)
)
```

All the complexity and polyfilling should be hidden from the end user

carlobernardini:
So, as discussed, we agreed on the following:

1. `ThemeLoader` is to be exposed by OneUI (consider renaming it to `OneUI` so it does not necessarily have to revolve around themes only? E.g. polyfill is not only necessary when using themes). Maybe a nice way of importing would be e.g. `import OneUI, { ComponentA, ComponentB, ... } from '@textkernel/oneui'`?
2. CSS vars polyfill is to be included as a OneUI dependency and is configured / initialized by `ThemeLoader`
3. 'Base CSS' (oneui.min.css) is to be imported on application level
4. `ThemeLoader` configuration params can be provided through configuration object, e.g. `init({ themeId: THEME_ID, themeCdn: CUSTOM_CDN, ... })`
5. README.md should be updated with clear examples of how to set things up

Please double check this list :)

antipin:
We should think carefully about the `customCDN` thing.

Assuming that our `OneUI.init({ cdn: 'http://foo.bar/css', themeId: 'theme42' })` works internally like this:

```js
... 
// Building theme url based on cdn and themeId
const themeURL = `${options.cdn}/${options.themeId}.css`
```

If you think about this, basically, it allows a customer to inject any css file again :)

carlobernardini:
Yes, the way themes are to be loaded needs more thought and a design. We can at least leave the CDN configuration option out for starters.

ribeiroguilherme:
I pushed the latest changes as we discussed. What is missing is only the README.md with the instructions on how to use the ThemeLoader

src/ThemeLoader.js
```diff
@@ -0,0 +1,98 @@
...
+        return Promise.race([loadTheme, timeout]);
+    }
+
+    /**
+     * Loads the CSS Vars ponyfill in case the browser is Internet Explorer 11. It can also
+     * forces to load in modern browsers in case the user passes `onlyLegacy` property as false
+     */
+    static startCssVarsPonyfill(ponyfillOptions = { onlyLegacy: true }) {
+        return new Promise((resolve, reject) => {
+            if (!isInternetExplorer11() && ponyfillOptions.onlyLegacy === true) {
```
eszthoff: The default object won't work if ponyfillOptions is defined, but without the `onlyLegacy` key. In that case `ponyfillOptions.onlyLegacy` will be undefined. Since the ponyfill's default is `true` anyway, I suggest to drop the default in the definition above. In anycase, should add `|| ponyfillOptions.onlyLegacy===undefined`

src/ThemeLoader.js
```diff
@@ -0,0 +1,98 @@
...
+                resolve();
+                return;
+            }
+
+            cssVarsPonyfill({
+                ...ponyfillOptions,
+                // eslint-disable-next-line no-unused-vars
+                onComplete(cssText, node, url) {
```
eszthoff: this and the next will override possible options passed from the user. So maybe good to call that here, to make sure: 
`onComplete(cssText, node, url) { if(ponyfillOptions.onComplete) {ponyfillOptions.onComplete(cssText, node, url)} resolve()}`
Then you can also remove the eslint-disable.
Same for `onError`

src/ThemeLoader.js
```diff
@@ -0,0 +1,98 @@
+import cssVarsPonyfill from 'css-vars-ponyfill';
+
+const buildURL = (cdn = '', themeId = '') => {
+    if (cdn === '' && themeId === '') return '';
+
+    if (cdn.charAt(cdn.length - 1) !== '/') {
+        return `${cdn}/${themeId}`;
+    }
+
+    return `${cdn}${themeId}`;
```
eszthoff: shouldn't the url have .css at the end?

src/ThemeLoader.js
```diff
@@ -0,0 +1,98 @@
...
+
+const DEFAULT_LOADING_TIMEOUT = 2000;
+
+class ThemeLoader {
+    /**
+     * Loads the theme and the CSS vars ponyfill if necessary
+     *
+     * @param {Object} themeConfig - Object that contains the configuration to initialize the theme
+     * @param {string} themeConfig.cdn - Server URL that hosts the theme
+     * @param {string} themeConfig.themeId - Theme unique idenfitier
```
eszthoff: maybe point it out here that the theme css is expected to live under cdn/themid.css

src/ThemeLoader.js
```diff
@@ -0,0 +1,98 @@
+import cssVarsPonyfill from 'css-vars-ponyfill';
```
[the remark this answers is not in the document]
ribeiroguilherme: Packages in my point of view is a folder that we can add code that isn't related to OneUI at all. As for example the BEM js package that, at some point, can be extracted as a separate library.
ThemeLoader is something specific to OneUI. For me makes more sense to create a folder `utils` and leave it there. What do you think?

src/ThemeLoader.js
ribeiroguilherme: I don't think so. Themes are identified by ID. So at the end we would end up in a URL like: `http://cdn.com/themes?id=12388123` , that returns a CSS . This is how I see

src/ThemeLoader.js
ribeiroguilherme: good idea!

src/ThemeLoader.js
antipin: So far the implementation of ThemeRoller service PoC assumes that we should request it like http://cdn.com/themes/THEME_ID.css
But it's something to discuss

src/ThemeLoader.js
antipin: I'd not allow to override callbacks at all. Might be better to explicitly set all the props that can be overwritten?

example/src/index.js
```diff
@@ -1,5 +1,14 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import ThemeLoader from '@textkernel/oneui';
+import '@textkernel/oneui/dist/oneui.css';
```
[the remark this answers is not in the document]
antipin: App developer always can set up webpack in a way that it ignores particular imports.
As an example, if you want to use `normalize.css` on your project it's totally possible.

Here is a discussion about it with lots of ways to deal with it https://github.com/css-modules/css-modules/pull/65

src/ThemeLoader.js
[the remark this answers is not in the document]
eszthoff: that would work as well. As long as it isn't in src/ I'm good.

src/ThemeLoader.js
eszthoff: It isn't props really. It is Ponyfill options. And you need to extend it with local things. In this specific case, I think having a props specifically would be a lot more confusing then the suggested above

src/ThemeLoader.js
eszthoff: maybe it is good to add support for it when there is a cdn to support. Also it seems to be very 3rd party thing, and implementing it to this detail here is perhaps a bit too opinionated. What about having only one prop url, which can be constructed on the application level?

src/ThemeLoader.js
ribeiroguilherme: Done

src/utils/OneUI/__tests__/OneUI.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    it('should attach <link> element with theme URL to DOM tree', () => {
+        const themeURL = 'http://theme-cdn.com/mytheme.css';
+        document.body.innerHTML = '<head></head>';
+
+        OneUI.applyTheme(themeURL);
+
+        expect(document.getElementsByTagName('link')[0].href).toBe(themeURL);
+    });
+
+    it('should load the ponyfill if browser is IE11', () => {
```
eszthoff: maybe add also a test to check that ponyfill is not loaded for Chrome for instance

example/package.json
```diff
@@ -3,6 +3,7 @@
     "version": "0.1.0",
     "private": true,
     "dependencies": {
+        "@textkernel/oneui": "^0.1.0",
```
carlobernardini: @ribeiroguilherme would it make more sense to  use `*` since the example is always part of the same (latest) release? Otherwise we'd have to keep updating this I suppose

carlobernardini:
@ribeiroguilherme By the way, don't forget to update the readme - if not in this branch, add another one. I sent a MD snippet a while ago.

example/package.json
ribeiroguilherme: Done

eszthoff:
last commits, after I already approved, LGTM

PR #64, opened 2019-02-15
Text improvements

carlobernardini:
* Added `size` prop for Text component (small / normal / large)

### DIFF
--- a/src/components/Text/Text.js
+++ b/src/components/Text/Text.js
@@ -7,7 +7,7 @@ import { SIZES } from '../../constants';
 const { block } = bem({
     name: 'Text',
     classnames: styles,
-    propsToMods: ['bold', 'muted', 'size']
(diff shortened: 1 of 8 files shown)


src/components/Text/Text.js
```diff
@@ -2,15 +2,16 @@ import React from 'react';
...
 const { block } = bem({
     name: 'Text',
     classnames: styles,
-    propsToMods: ['muted']
+    propsToMods: ['muted', 'size']
 });
 
 const Text = props => {
-    const { children, inline, ...rest } = props;
+    const { bold, children, inline, size, ...rest } = props;
```
eszthoff: `bold`?

carlobernardini: I can add it if you think it should be configured through props but I vaguely recall you challenging that, hence I didn't add it yet

PR #66, opened 2019-02-15
new header component

eszthoff:
[ONEUI-22](https://jira.textkernel.nl/browse/ONEUI-22)


src/components/Header/Header.js
```diff
@@ -0,0 +1,45 @@
...
+            </div>
+        </div>
+    );
+};
+
+Header.displayName = 'Header';
+
+Header.propTypes = {
+    /** path to the logo source file */
+    logoSrc: PropTypes.string.isRequired,
```
carlobernardini: Maybe prettier to do use 'scoped' props?

```
logo: PropTypes.shape({
   src: PropTypes.string.isRequired,
   link: PropTypes.string,
   title: PropTypes.string
}).isRequired
```

src/components/Header/Header.js
```diff
@@ -0,0 +1,45 @@
...
+    /** path to where the logo should link to if clicked */
+    logoLink: PropTypes.string,
+    /** a name for the logo, it will used as alternative text to the img */
+    logoName: PropTypes.string,
```
carlobernardini: `title` instead of `name`?

src/components/Header/Header.js
```diff
@@ -0,0 +1,45 @@
...
+    );
+};
+
+Header.displayName = 'Header';
+
+Header.propTypes = {
+    /** path to the logo source file */
+    logoSrc: PropTypes.string.isRequired,
+    /** path to where the logo should link to if clicked */
+    logoLink: PropTypes.string,
```
carlobernardini: What if you need to use a callback function or using React router?

src/components/Header/Header.scss
```diff
@@ -0,0 +1,24 @@
...
+        display: flex;
+        justify-content: space-between;
+        align-items: baseline;
+        align-content: center;
+    }
+
+    &__logo {
+        height: 40px;
+    }
+}
```
carlobernardini: I would add `&__menu` and at least middle-align the content (vertically) using flexbox (`display: flex; align-items: center;`?)

stories/Header.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import { text, withKnobs } from '@storybook/addon-knobs'; // eslint-disable-line import/no-extraneous-dependencies
+import { Header } from '@textkernel/oneui';
+
+storiesOf('Header', module)
+    .addDecorator(withKnobs)
+    .add('Header', () => (
+        <Header
+            logoSrc={text('Image URL', 'jobfeed-logo.svg')}
```
carlobernardini: Maybe you can just link to https://www.jobfeed.nl/images/jobfeed-logo.svg so you don't have to add it as static resource ;) But doesn't matter that much

src/components/Header/Header.js
eszthoff: I was hoping you won't ask. If we want to handle those cases, we need to define them in more detail. I think the most common case for the logo is to just link to Home ('/'). Which should work even without react-router link. I suggest that we handle other cases when we actually come across them.

stories/Header.js
eszthoff: ok

src/components/Header/Header.scss
eszthoff: ok

src/components/Header/Header.js
eszthoff: it is prettier, but I found a few times that flat structures are easier to handle. For instance, setting defaults on an object is not possible without some magic, since they aren't deep merged. Since these are the only props, I think it is ok to leave them flat.

src/components/Header/Header.js
eszthoff: ok

src/components/Header/Header.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+const Header = props => {
+    const { logoSrc, logoLink, logoName, children, ...rest } = props;
+
+    return (
+        <div {...rest} {...block(props)}>
+            <div {...elem('wrapper', props)}>
+                <a href={logoLink}>
+                    <img src={logoSrc} alt={logoName} {...elem('logo', props)} />
```
antipin: What if in **Search!** we have a component which is already a logo itself?

Assuming that logo should be something configurable (different customers – different logos etc), I don't see it as a part of `<Header />`.

You're putting too much responsibility to one component. That also relates to the question from @carlobernardini about onClick handling.

I'd like to be able to do this:
```js
import { SearchLogo } from 'some/common/search/components';

...

render() {
  return (
    <Header
      logo={<SearchLogo onClick={...} />}
    >
      some extra content  
    </Header>
  );
}





<IconSearch size={48} title="Search!" />

Comment (eszthoff) [reply src/components/Header/Header.js]: What do you expect to happen if the logo supplied is larger then the header container? According to @carlobernardini Carlo it should have a fix height. If that is not a constrain, we can change the logo to be a single prop that accepts only node. Or do you want to keep the current options too?

File: src/components/Header/Header.scss [hunk 7]
```diff
@@ -0,0 +1,24 @@
+@import '../../themes/oneui/oneui.scss';
+
+.Header {
+    width: 100%;
```
carlobernardini: Missing `background-color`: it will be transparent now. It would also be good to add a component-level variable for that, so you can have whichever header background color you want.

src/components/Header/Header.scss
carlobernardini: Sure, it can be applied to `wrapper` too

src/components/Header/Header.js
carlobernardini: As long as you say it can work with React router without having to reload a whole page it's fine with me for now

src/components/Header/Header.js
antipin: e.g. we can set a fixed height to the `<Header/>` and "tell" to the `<Logo/>` (via passing some prop for instance) that it should take all the vertical space.

eszthoff: Ok

src/components/Header/Header.js
carlobernardini: But in this case `src` is required anyway and the others don't necessarily need a default, don't they? But as you say, it's mostly a matter of 'prettier props'

src/themes/oneui/base-variables/_sizing.scss
```diff
@@ -18,6 +18,9 @@ $spacing-5x: 5 * $spacing-normal !default;
 
 $margin:    0.5 * $spacing-normal !default;
 
+// containers
+$site-container-size: 1024px !default;
```
carlobernardini: Can we set this to 1150 for starters? Most displays support at least 1280 nowadays. We can still fine-tune it later but this is what I currently use in my designs

stories/Header.js
```diff
@@ -0,0 +1,23 @@
...
+    .addDecorator(withKnobs)
+    .add('Header with simple logo', () => (
+        <Header
+            logo={{
+                src: text('Image URL', 'https://www.jobfeed.nl/images/jobfeed-logo.svg'),
+                link: text('URL where click on logo will lead', '/'),
+                title: text('Alt text for the logo', 'Jobfeed')
+            }}
+        >
+            {text('Left side', 'This is a placeholder for children')}
```
carlobernardini: Right side?

stories/Header.js
```diff
@@ -0,0 +1,23 @@
...
+        </Header>
+    ))
+    .add('Header with logo as component', () => (
+        <Header logo={<IconTextkernel />}>
+            {text('Left side', 'This is a placeholder for children')}
```
carlobernardini: Right side?

src/components/Header/Header.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../themes/oneui/oneui.scss';
+
+.Header {
```
carlobernardini: The header is currently not using `font-family-primary`, but I suppose this should ideally be implemented globally instead of for every component

eszthoff: I agree. it would be annoying to set it everywhere all the time.

PR #67, opened 2019-02-18
Footer

eszthoff:


carlobernardini:
Eventually the footer [should be 'sticky'](https://css-tricks.com/couple-takes-sticky-footer/), meaning that it should stick to the bottom of the page even if the page has little content. I think this probably needs to be dealt with on application level, but it may have some implications for the component design (in terms of CSS / props) too.

src/components/Footer/Footer.js
```diff
@@ -0,0 +1,51 @@
...
+});
+
+const Footer = props => {
+    const { copyright, children, ...rest } = props;
+
+    const tkCopyright = (
+        <span>
+            {`\u00a9 ${new Date().getFullYear()} `}
+            <strong>
+                <IconTextkernel context="background" />
```
carlobernardini: I don't think we should add `background` and `foreground` as contextual colors. Moreover, `background` can be different from `footer-foreground-color`. Instead, for such a default we may be able to do something like this (I think):

```
<IconTextkernel {...elem('logo', props) />
...
.Footer__logo {
   fill: var(--footer-foreground-color);
}
```

...correct me if I'm wrong.

Also, if we apply a default when `copyright` is not provided, I think we should somehow use the logo text (SVG). If that's hard or we're not yet sure how exactly to do that, I would suggest to not have a default '`tkCopyright'` for now (so either you provide something yourself or it will not render a copyright - then we can still add a default later). Agree?

PR #68, opened 2019-02-19
fix: improved icon coloring

carlobernardini:
* Apply `fill: var(--color-foreground)` by default (make context optional);
* Apply `fill: var(--link-color-normal)` / `fill: var(--link-color-hover)` when used within a Link component


eszthoff:
If you already revising the Icons, can you add tests to IconSearch? That is the only one that has none. The rest of the changes look good.

carlobernardini:
@eszthoff Yes good catch, I'll add some tests for the Search icon
@antipin You had a suggestion for making the icon inherit the color of its parent container? If you can point me to what you mean I can consider applying it in this PR

eszthoff:
I just had an idea of how to possibly simplify adding more icons. It would change the API slightly though. Instead of creating all the IconName as components, we could just host svg file with the path and keep the IconBase component only. Then the API would look like:

eszthoff:
Looking at this implementation again, I feel uneasy about it. I really don't like an API where you can set a prop to something, but the default has no name. I feel that it creates a magical situation. @antipin and @ribeiroguilherme  could you comment on it?
Also please see the proposal above about simplifying this component in general.

PR #69, opened 2019-02-19
navigation component

eszthoff:
[ONEUI-24](https://jira.textkernel.nl/browse/ONEUI-24)


carlobernardini:
I've taken the liberty to push a commit with some _suggested_ changes. We can walk through and discuss these together. It mostly concerns styling and a bit of reorganization.

One thing that I didn't touch but I would really prefer to prevent is introducing react-router as a dependency, only to make `activeClassName` work. I'm confident we can find a solution to that while keeping OneUI independent from react-router. Maybe @antipin or @ribeiroguilherme have opinions or ideas on that as well.

PR #76, opened 2019-03-15
WIP: List, ListItem and ListActions

eszthoff:
[ONEUI-33](https://jira.textkernel.nl/browse/ONEUI-33)

### DIFF
--- a/src/components/List/List.js
+++ b/src/components/List/List.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem';
+import styles from './List.scss';
+
+const { block } = bem({
+    name: 'List',
+    classnames: styles,
+    propsToMods: ['isDividedList', 'Component']
+});
+
+const List = props => {
+    const { children, Component, isDividedList, ...rest } = props;
+    return (
+        <Component {...rest} {...block(props)}>
+            {React.Children.map(children, child => React.cloneElement(child, { isDividedList }))}
+        </Component>
+    );
+};
+
+List.displayName = 'List';
+
+List.propTypes = {
+    /** List Items */
+    children: PropTypes.node,
+    /** The HTML or React component to be used to render the list  */
+    Component: PropTypes.string,
+    /** Adds dividing lines between the list items */
+    isDividedList: PropTypes.bool
+};
+
+List.defaultProps = {
+    children: null,
+    Component: 'ul',
+    isDividedList: false
+};
+
+export default List;
--- a/src/components/List/List.js
+++ b/src/components/List/List.js
@@ -24,7 +24,7 @@ List.propTypes = {
     /** List Items */
     children: PropTypes.node,
     /** The HTML or React component to be used to render the list  */
-    Component: PropTypes.string,
+    Component: PropTypes.oneOf(['ul', 'ol', 'menu']),
     /** Adds dividing lines between the list items */
     isDividedList: PropTypes.bool
 };
--- a/src/components/List/List.js
+++ b/src/components/List/List.js
@@ -3,17 +3,17 @@ import PropTypes from 'prop-types';
 import bem from 'bem';
 import styles from './List.scss';
 
-const { block } = bem({
+const { block, elem } = bem({
     name: 'List',
     classnames: styles,
-    propsToMods: ['isDividedList', 'Component']
+    propsToMods: ['isDivided', 'Component']
 });
 
 const List = props => {
-    const { children, Component, isDividedList, ...rest } = props;
+    const { children, Component, isDivided, ...rest } = props;
     return (
         <Component {...rest} {...block(props)}>
-            {React.Children.map(children, child => React.cloneElement(child, { isDividedList }))}
+            {React.Children.map(children, child => React.cloneElement(child, elem('item', props)))}
         </Component>
     );
 };
@@ -26,13 +26,13 @@ List.propTypes = {
     /** The HTML or React component to be used to render the list  */
     Component: PropTypes.oneOf(['ul', 'ol', 'menu']),
     /** Adds dividing lines between the list items */
-    isDividedList: PropTypes.bool
+    isDivided: PropTypes.bool
 };
 
 List.defaultProps = {
     children: null,
     Component: 'ul',
-    isDividedList: false
+    isDivided: false
 };
 
 export default List;
--- a/src/components/List/List.scss
+++ b/src/components/List/List.scss
@@ -1,10 +1,16 @@
 .List {
     &--Component_ul {
-        padding: 0 var(--spacing-2x);
+        padding: var(--spacing-2x);
         list-style: none;
     }
 
     &--Component_menu {
         list-style: none;
+    }
+
+    &--isDivided {
+        .List__item + .List__item {
+            border-top: 1px solid var(--color-neutral-25);
+        }
     }
 }
--- a/src/components/List/List.scss
+++ b/src/components/List/List.scss
@@ -1,0 +1,10 @@
+.List {
+    &--Component_ul {
+        padding: 0 var(--spacing-2x);
+        list-style: none;
+    }
+
+    &--Component_menu {
+        list-style: none;
+    }
+}
--- a/src/components/List/ListItem/ListItem.js
+++ b/src/components/List/ListItem/ListItem.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem';
+import Text from '../../Text';
+import styles from './ListItem.scss';
+
+const { block, elem } = bem({
+    name: 'ListItem',
+    classnames: styles,
+    propsToMods: ['isDivided', 'isDividedList', 'isSelected', 'onClick']
+});
+
+const ListItem = props => {
+    const { children, isDivided, isDividedList, isSelected, onClick, ...rest } = props;
+    const customBlockMod = { clickable: typeof onClick === 'function' };
+
+    return (
+        <li {...rest} {...block(props, customBlockMod)}>
+            <div onClick={onClick} role="presentation" {...elem('container', props)}>
+                {React.Children.map(
+                    children,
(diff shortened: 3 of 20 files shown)


src/components/List/List.js
```diff
@@ -0,0 +1,38 @@
...
+    );
+};
+
+List.displayName = 'List';
+
+List.propTypes = {
+    /** List Items */
+    children: PropTypes.node,
+    /** The HTML or React component to be used to render the list  */
+    Component: PropTypes.string,
```
carlobernardini: I'm not sure if we really need to make the list type configurable, I'd probably just go for `ul` out of the box. This makes it more complicated, especially considering class naming / bem. But if you really want to keep it, I would at least predefine the types: `oneOf(['ul', 'ol', ...])`.

src/components/List/List.js
```diff
@@ -0,0 +1,38 @@
...
+    /** Adds dividing lines between the list items */
+    isDividedList: PropTypes.bool
+};
+
+List.defaultProps = {
+    children: null,
+    Component: 'ul',
+    isDividedList: false
```
carlobernardini: I would change this to just `isDivided`. To render the divider, you could then do e.g. this:

```
{React.Children.map(children, child => React.cloneElement(child, { ...elem('item', props) }))}
...
.List__item + .List__item {
   border-top: 1px solid color;
}
```

(note that I'm adding a element class to the cloned element)

src/components/List/ListItem/ListItem.scss
```diff
@@ -0,0 +1,27 @@
...
+
+    &:hover {
+        background-color: var(--color-brand-10);
+    }
+
+    &--isSelected {
+        background-color: var(--color-brand-10);
+    }
+
+    &--isDividedList:not(:last-child) {
```
carlobernardini: `:not` [doesn't have much support](https://caniuse.com/#feat=css-not-sel-list)

I'd still go for using adjacent sibling selectors (+)

src/components/List/ListItem/ListItem.js
```diff
@@ -0,0 +1,52 @@
...
+
+ListItem.propTypes = {
+    /** List Items */
+    children: PropTypes.node,
+    /** Formats this item as selected  */
+    isSelected: PropTypes.bool,
+    /** Adds a dividing line below the items. Do not use on divided lists */
+    isDivided: PropTypes.bool,
+    /** For internal use: adds a dividing line below all items but the last. isDivided set to true will overwrite this param */
+    isDividedList: PropTypes.bool,
```
carlobernardini: I don't think we need / should have this prop, see https://github.com/textkernel/oneui/pull/76/files#r266105091

src/components/List/ListItem/ListItem.js
```diff
@@ -0,0 +1,52 @@
...
+
+ListItem.displayName = 'ListItem';
+
+ListItem.propTypes = {
+    /** List Items */
+    children: PropTypes.node,
+    /** Formats this item as selected  */
+    isSelected: PropTypes.bool,
+    /** Adds a dividing line below the items. Do not use on divided lists */
+    isDivided: PropTypes.bool,
```
carlobernardini: We don't need this for now so maybe just leave it out until we do?

src/components/List/List.js
eszthoff: This is just an other approach to what I did. Is there a specific advantage you see in this one as opposed to the current one?

src/components/List/ListItem/ListItem.scss
eszthoff: you link is incorrect. It is about having a list of selector to be ignored. Simple `:not()` is [well supported](https://caniuse.com/#feat=css-sel3).

src/components/List/List.js
eszthoff: this is just an other approach to what is currently implemented. Is there a specific reason you think would be better?

carlobernardini: In a nutshell I just think that supporting division on Item level is not really necessary, and `isDivided` on list level makes more sense to me in terms of naming because it's on List level anyway (so the name `isDividedList` is a bit double). In terms of CSS I personally think using adjacent sibling selectors is a bit more readable (and that's how I think it's typically done) over using multiple pseudo-selectors, but that's more a matter of preference. And as mentioned before I would consider always using one particular list type (`ul` would be fine for the current use case?) for now instead of making that configurable, to keep it simpler. If necessary we can make it configurable later on, right?

src/components/List/ListItem/ListItem.js
```diff
@@ -0,0 +1,52 @@
...
+    );
+};
+
+ListItem.displayName = 'ListItem';
+
+ListItem.propTypes = {
+    /** List Items */
+    children: PropTypes.node,
+    /** Formats this item as selected  */
+    isSelected: PropTypes.bool,
```
carlobernardini: Is this prop used both for 'selected' (result is selected) and 'active' (viewing details) state? Would it make sense to have a separate `isActive` prop? Even if by default both states would visually be the same, one may want to have it styled differently in a custom theme, and I'd say it's more explicitly differentiated

eszthoff: See discussion in the design document. My understanding that for now we keep it with one prop. In the future if we have an actual use case of differentiating them, and we know what they really need to be called, we can add an other prop.

src/components/List/List.js
eszthoff: I don't see this as a problem to have it. `ul` and `ol` are standard html tags that one expects to have for a list. It is already implemented, I rather not spend time on removing it for no real reason.
I fixed the prop types as suggested.

carlobernardini: The reason for not having it is that there is no necessity and because we don't have a requirement we are implementing this on the basis of an assumption and it makes the component more complicated. Hence my comment.

src/components/List/List.js
antipin: I don't see much sense in implementing such an abstract and generic `List` component. If we really need to support ordered and unordered lists (which I doubt), might be make sense to create two different components: `OrderedList` and `UnorderedList`. Or at least the interface of the component should be more semantic, like  `<List  type="ordered|unordered" />`.

One more thought on this. I think it's a part of responsibility of this component to render tags for items. In this case we can do 2 things:

1. control that `<ul>` tag contains only allowed `<li>` tags inside
2. Don't need to clone children elements and passing `isDividedList`.

```
return (
  <ul {...rest} {...block(props)}>
     {React.Children.map(children, child => (
        <li {...elem('item', props)}>
          {child}
        </li>
      ) 
    ))}
  </Component>
);
```

src/components/List/ListItem/ListItem.js
```diff
@@ -0,0 +1,52 @@
...
+import styles from './ListItem.scss';
+
+const { block, elem } = bem({
+    name: 'ListItem',
+    classnames: styles,
+    propsToMods: ['isDivided', 'isDividedList', 'isSelected', 'onClick']
+});
+
+const ListItem = props => {
+    const { children, isDivided, isDividedList, isSelected, onClick, ...rest } = props;
```
antipin: `isDivided, isDividedList` looks too unclead. What prop is handle what?

src/components/List/ListItem/ListItem.scss
```diff
@@ -0,0 +1,27 @@
...
+        border-bottom: 1px solid var(--color-neutral);
```
carlobernardini: Can you make the division border color lighter please? `var(--color-neutral-25)` should do - in general, you can take the tryout I made (65e8f88c0c8489188462b3344d828c3026e7ee5c) as a visual reference if that helps you

src/components/List/List.js
eszthoff: I don't see it is as obstacle and it was part of the original design

src/components/List/List.js
eszthoff: changed to use `isDivided` on list only.

src/components/List/ListItem/ListItem.js
eszthoff: removed

src/components/List/ListItem/ListItem.js
eszthoff: removed

src/components/List/ListItem/ListItem.scss
eszthoff: done

src/components/List/List.js
carlobernardini: I don't get why you're digging in your heels to keep this but I'm not going to continue challenging it either. Not saying it's an obstacle besides it makes the code more complex while we probably won't even need it.

src/components/List/List.scss
```diff
@@ -0,0 +1,16 @@
+.List {
+    &--Component_ul {
+        padding: var(--spacing-2x);
+        list-style: none;
```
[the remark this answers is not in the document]
eszthoff: yes, ol should have list-style, otherwise you won't see the numbering.

src/components/List/List.js
```diff
@@ -0,0 +1,38 @@
...
+    );
+};
+
+List.displayName = 'List';
+
+List.propTypes = {
+    /** List Items */
+    children: PropTypes.node,
+    /** The HTML or React component to be used to render the list  */
+    Component: PropTypes.oneOf(['ul', 'ol', 'menu']),
```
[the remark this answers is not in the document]
eszthoff: perhaps, but since the design was approved, let's not change it at this stage.

src/components/List/List.js
```diff
@@ -0,0 +1,38 @@
...
+    /** Adds dividing lines between the list items */
+    isDivided: PropTypes.bool
```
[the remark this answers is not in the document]
eszthoff: same as above

src/components/List/List.js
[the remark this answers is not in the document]
SebastiaanPasterkamp: Perhaps this can be done in a separate PR then? If we change this later, then we (and our users) will have to fix it in more places.

PR #79, opened 2019-03-22
Content placeholder

carlobernardini:
Adds a simple `ContentPlaceholder` component

* `height`: Variable placeholder height in pixels. When not set, it will inherit `lineHeight` by default.
* `width`: Variable placeholder width, relative to the parent container. When not set, it will default to `100%`.
* `duration`: Custom animation duration in seconds. When not set, it will default to `1s`.

The component is to be used inside a block-level parent element (or an element that has a width defined).

### DIFF
--- a/src/components/ContentPlaceholder/ContentPlaceholder.js
+++ b/src/components/ContentPlaceholder/ContentPlaceholder.js
@@ -1,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem';
+import styles from './ContentPlaceholder.scss';
+
+const { block, elem } = bem({
+    name: 'ContentPlaceholder',
+    classnames: styles,
+    propsToMods: []
+});
+
+const calcDuration = seconds => {
+    if (!seconds) {
+        return null;
+    }
+
+    return `${seconds}s`;
+};
+
+const calcMaskWidth = width => {
+    if (!width) {
+        return null;
+    }
+
+    return {
+        width: `${100 - Math.max(Math.min(width, 100), 0)}%`
+    };
+};
+
+const ContentPlaceholder = props => {
+    const { duration, height, width, ...rest } = props;
+
+    return (
+        <div
+            {...rest}
+            {...block(props)}
+            style={{
+                animationDuration: calcDuration(duration),
+                height
+            }}
+        >
+            <div {...elem('mask', props)} style={calcMaskWidth(width)}>
+                &nbsp;
+            </div>
+        </div>
+    );
+};
+
+ContentPlaceholder.displayName = 'ContentPlaceholder';
+
+ContentPlaceholder.propTypes = {
+    /** Animation duration in seconds */
+    duration: PropTypes.number,
+    /** Custom height in pixels for the content placeholder */
+    height: PropTypes.number,
+    /** Width of the content placeholder, relative to its parent */
+    width: PropTypes.number
+};
+
+ContentPlaceholder.defaultProps = {
+    duration: 1,
+    height: null,
+    width: 100
+};
+
+export default ContentPlaceholder;
--- a/src/components/ContentPlaceholder/ContentPlaceholder.js
+++ b/src/components/ContentPlaceholder/ContentPlaceholder.js
@@ -56,9 +56,9 @@ ContentPlaceholder.propTypes = {
 };
 
 ContentPlaceholder.defaultProps = {
-    duration: 1,
+    duration: null,
     height: null,
-    width: 100
+    width: null
 };
 
 export default ContentPlaceholder;
--- a/src/components/ContentPlaceholder/ContentPlaceholder.js
+++ b/src/components/ContentPlaceholder/ContentPlaceholder.js
@@ -6,26 +6,14 @@ import styles from './ContentPlaceholder.scss';
 const { block, elem } = bem({
     name: 'ContentPlaceholder',
     classnames: styles,
-    propsToMods: []
+    propsToMods: ['withoutMargin']
 });
 
-const calcDuration = seconds => {
-    if (!seconds) {
-        return null;
-    }
+const calcDuration = seconds => `${seconds}s`;
 
-    return `${seconds}s`;
-};
-
-const calcMaskWidth = width => {
-    if (!width) {
-        return null;
-    }
-
-    return {
-        width: `${100 - Math.max(Math.min(width, 100), 0)}%`
-    };
-};
+const calcMaskWidth = width => ({
+    width: `${100 - Math.max(Math.min(width, 100), 0)}%`
+});
 
 const ContentPlaceholder = props => {
     const { duration, height, width, ...rest } = props;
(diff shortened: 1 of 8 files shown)


src/components/ContentPlaceholder/ContentPlaceholder.js
```diff
@@ -0,0 +1,65 @@
...
+    height: PropTypes.number,
+    /** Width of the content placeholder, relative to its parent */
+    width: PropTypes.number
+};
+
+ContentPlaceholder.defaultProps = {
+    duration: null,
+    height: null,
+    width: null
+};
```
eszthoff: I think it would be easier to read if the defaults would be the actual defaults defined in the css. `duration: 1, width: 0`. Then the `if` can be removed from the calc functions, which makes it simpler (and a bit faster)

src/components/ContentPlaceholder/ContentPlaceholder.scss
```diff
@@ -0,0 +1,35 @@
...
+    background: var(--color-neutral-25);
+    background: linear-gradient(to right, var(--color-neutral-10) 5%, var(--color-neutral-25) 15%, var(--color-neutral-10) 30%);
+    background-size: 200% 100%;
+    position: relative;
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+
+    &__mask {
+        background-color: white;
```
eszthoff: this should use `var(--color-background)`

src/components/ContentPlaceholder/ContentPlaceholder.scss
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    100% {
+        background-position: -100% 0;
+    }
+}
+
+.ContentPlaceholder {
+    animation:{
+        duration: 1s;
```
eszthoff: since it is set in the style element, it is redundant here. can it be removed? Or you think it is makes it more readable?

src/components/ContentPlaceholder/ContentPlaceholder.scss
```diff
@@ -0,0 +1,35 @@
...
+    background-size: 200% 100%;
+    position: relative;
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+
+    &__mask {
+        background-color: white;
+        height: inherit;
+        width: 0;
```
eszthoff: since it is set in the style element, it is redundant here. can it be removed? Or you think it is makes it more readable?

src/components/ContentPlaceholder/ContentPlaceholder.js
```diff
@@ -0,0 +1,65 @@
...
+    return (
+        <div
+            {...rest}
+            {...block(props)}
+            style={{
+                animationDuration: calcDuration(duration),
+                height
+            }}
+        >
+            &nbsp;
```
eszthoff: necessary?

stories/ContentPlaceholder.js
```diff
@@ -0,0 +1,37 @@
...
+                </div>
+                <div style={{ margin: '6px 0', lineHeight: '16px' }}>
+                    <ContentPlaceholder
+                        duration={duration}
+                        height={height}
+                        width={number('Width for placeholder 3', 65)}
+                    />
+                </div>
+            </div>
+        );
```
eszthoff: I think this structure will result in a not very beautiful html when rendered, with many nested divs. Can you give example that is simpler? Like adding the `<ContentPlaceholder>` without wrapping each in an individual `div`. E.g. add a classname so that they get the margins they need?
Also wrapping them in `p` or `h1` might be more intuitive when trying to mimic a look, if wrapping is necessary. There is a problem with that though, since the placeholder itself is a `div`, one cannot nest it in `p`. The solution might be to change the placeholder to render `span` instead.
What do you think?

src/components/ContentPlaceholder/ContentPlaceholder.js
carlobernardini: The problem with that, I think, is that the styles will always be declared inline (even by default). I don't think that's a very good practice in terms of specificity (inline styling will always take precedence).

src/components/ContentPlaceholder/ContentPlaceholder.scss
carlobernardini: It's not set as style attribute by default (this is the default). Only if the prop is actually used to apply a custom duration, it will be declared inline (which takes precedence over the stylesheet declaration).

src/components/ContentPlaceholder/ContentPlaceholder.js
carlobernardini: Yes I added this white space on purpose, otherwise the box will not have a width / height

stories/ContentPlaceholder.js
carlobernardini: I'm not sure how to produce a simpler DOM structure than having it render two elements while preserving the required functionality. It shouldn't use `span`, because spans are not meant to be block-level elements - this requires a block-level display.
I think you just shouldn't use them inside a paragraph or heading (doesn't make much sense); just show the placeholder instead of the heading, for example?

src/components/ContentPlaceholder/ContentPlaceholder.scss
carlobernardini: Makes sense, will update this

src/components/ContentPlaceholder/ContentPlaceholder.scss
eszthoff: yes. that what I meant. have it inline only. But have it there always (with default prop set to 1). Then the default and the custom works in the same way, and is easier to understand.

src/components/ContentPlaceholder/ContentPlaceholder.scss
carlobernardini: See previous comments

src/components/ContentPlaceholder/ContentPlaceholder.js
eszthoff: cool

stories/ContentPlaceholder.js
eszthoff: yes. Probably the best would be if we don't need to wrap each placeholder at all, apart from the main parent element, which sets the width. but can we add the spacing otherwise? Maybe even by default by adding a rule for `ContentPlaceholder + ContentPlaceholder`?
Of course, one can still do the div thing as here, but that if you want to do fancier things. It would be nice if something like this would work out of the box:
```
<div style={width: '400px'}>
<ContnetPlaceholder/>
<ContnetPlaceholder/>
<ContnetPlaceholder/>
</div>
```

src/components/ContentPlaceholder/ContentPlaceholder.js
carlobernardini: (continuing here)

I don't think we should declare the default both in stylesheet and inline, so then it would be all-inline. I think it depends on what you think is more important:

1. Readability: Keep default and custom value both inline for sake of readability because they're all in one place
2. How your code works: keep default CSS in stylesheet, declare inline only in case of custom values (because CSS should be kept in stylesheets as much as possible, otherwise you may run into unexpected specificity issues etc. And presentation & markup should be separated, ideally)

I personally prefer 2

stories/ContentPlaceholder.js
carlobernardini: Would it be better to apply the spacing on application level, since it depends quite much on how you wish it to be implemented? Otherwise I suppose we would have to implement various levels of spacing, making assumptions on how the placeholder is used. I'd say the most convenient thing to do is to pass an additional classname of your own and have the spacing be defined in the application CSS. What do you say?

PR #83, opened 2019-03-27
feat(component): pagination components

carlobernardini:


src/components/Pagination/__tests__/__snapshots__/Pagination.spec.js.snap
```diff
@@ -0,0 +1,93 @@
...
+    aria-disabled={true}
+    className="Pagination__button Pagination__button--align Pagination__button--align_center"
+    context="link"
+    data-page={0}
+    disabled={true}
+    isBlock={false}
+    isInline={false}
+    onClick={[Function]}
+    size="normal"
+    type="button"
```
carlobernardini: @eszthoff Minor detail but something I didn't quite understand myself is why the snapshot contains this `type` attribute since it is not explicitly applied in `Pagination.js` (so the `Button` component would internally use the default, which is `button`, but normally that wouldn't be reflected in the snapshot, or would it?) - Same for the `PaginationButton` snapshot.

Only relevant if you choose to proceed with this implementation of course.

SebastiaanPasterkamp: AFAIK e.g. `Button.defaultProps` are included in snapshots.

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,164 @@
...
+            )}
+            <PaginationButton
+                onClick={handleClick}
+                isActive={currentPage === 1}
+                data-page={1}
+                key="page_1"
+            >
+                {1}
+            </PaginationButton>
+            {range[0] > 2 && <div {...elem('gap', props)}>&hellip;</div>}
```
SebastiaanPasterkamp: I'm missing the `&hellip;` between `current ... next> last>>`

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,164 @@
...
+        const range = new Array(end - start + 1).fill().map((_, i) => start + i);
+
+        return range;
+    };
+
+    const isPrevDisabled = currentPage === 1;
+    const isNextDisabled = currentPage === totalPages;
+    const range = defineRange();
+
+    const handleClick = e => {
```
SebastiaanPasterkamp: This means all buttons will always rerender, because the `handleClick` function has changed.

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,164 @@
...
+        ...rest
+    } = props;
+
+    const defineRange = () => {
+        const showBefore = Math.floor((maxPageButtons - 1) / 2);
+        const start = Math.max(
+            2,
+            Math.min(currentPage - showBefore, totalPages - maxPageButtons + 1)
+        );
+        const end = Math.min(totalPages, start + maxPageButtons - 2);
```
SebastiaanPasterkamp: ` - 2` means the last button is always missing.

`currentPage={8} totalPages={8} maxPageButtons={7}`

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,164 @@
...
+        const range = new Array(end - start + 1).fill().map((_, i) => start + i);
+
+        return range;
+    };
+
+    const isPrevDisabled = currentPage === 1;
+    const isNextDisabled = currentPage === totalPages;
```
SebastiaanPasterkamp: `next` or `last` will not be disabled if somehow `currentPage > totalPages`. Since neither variable is owned by this component, it'd make sense to at least not provide buttons to non-existent pages.

Same goes for `currentPage < 1`.

`currentPage={100} totalPages={8} maxPageButtons={5}`
`currentPage={0} totalPages={8} maxPageButtons={5}`

src/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,164 @@
...
+            {!!firstLabel && (
+                <Button
+                    {...elem('button', props)}
+                    onClick={handleClick}
+                    disabled={isPrevDisabled}
+                    context="link"
+                    data-page={1}
+                    aria-disabled={isPrevDisabled}
+                >
+                    {`\u00ab ${firstLabel}`}
```
SebastiaanPasterkamp: Would it make sense to style these arrows, instead of concatenating? This button will always rerender now.

src/components/Pagination/__tests__/Pagination.spec.js
```diff
@@ -0,0 +1,112 @@
...
+    afterEach(() => {
+        jest.resetAllMocks();
+    });
+
+    it('should render correctly with default props', () => {
+        wrapper = shallow(<Pagination totalPages={20} />);
+
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+    it('should render correctly with all props', () => {
```
SebastiaanPasterkamp: Can we add test coverage for all funky corner cases:
* `maxPageButtons >= totalPages`
* `currentPage = 1`
* `currentPage = totalPages` (currently broken)
* gaps between `start ... current block ... end`
* `maxPageButtons = 1, currentPage != 1` (currently broken)

They don't have to be snapshots, but something like `given props {....} we expect to see buttons 1, 4, 5, 6 and 1 hellip`

carlobernardini:
> There's a very strange interaction with the forced [1] button and maxPageButtons. 

I think page 1 should be included with `maxPageButtons` count because it's still a 'page button' like all others. Even though the `firstLabel` prop is now implemented it is not meant to be used in the actual application. Perhaps the most straightforward solution would be that `maxPageButtons` cannot be < 2, *or* the logic should be that when `maxPageButtons === 1` you always show the current page (so in this particular case the 1st page is not always shown).

src/components/Pagination/Pagination.js
carlobernardini: Rerender because of what? Btw is there a reason to prefer unicode over html entities (like before - or even the `«` character directly)? This syntax reads more complicated than...

```
<Button>
    &laquo; {prevLabel}
</Button>
```

src/components/Pagination/Pagination.js
carlobernardini: Because it's not there in the design.

SebastiaanPasterkamp: > When some pages are not listed, show a 'gap' (horizontal ellipsis, see design attached)

It doesn't specify which unlisted pages are being referred to, and there is no design attached.

I read this as including _'pages not listed between current and last'_.

src/components/Pagination/Pagination.js
SebastiaanPasterkamp: This will work too, and it even honors the spacing.

src/components/Pagination/Pagination.js
carlobernardini: The designs are both on conceptboard as well as in the story.

What I mean with 'some pages are not listed' are those in between the first page and the current. As you can see in the design there is no gap beyond that, nor is the last page always shown.

SebastiaanPasterkamp:
> **or** the logic should be that when `maxPageButtons === 1` you always show the current page (so in this particular case the 1st page is not always shown).

I vote for this

src/components/Pagination/Pagination.js
eszthoff: > I don't think the assumption that someone might accidentally remove a whitespace is a strong argument. We have reviews for that.
I saw large refactors more then once, that effected such cases. Reviews missed it because they are a needle in a haystack. The latest example was introducing eslint in the current JF stack. Whitespace deletions are really hard to spot, and since we don't have visual regression tests, I would really like to make the code more resilient in this way

src/components/Pagination/Pagination.js
eszthoff: now it throws an error if  current page is illogical

src/components/Pagination/Pagination.js
eszthoff: works ok now

src/components/Pagination/__tests__/Pagination.spec.js
eszthoff: added some

PR #89, opened 2019-04-23
docs(releasing): describe procedure for making new release

carlobernardini:
Add RELEASING.md file describing specifications for version numbering and release procedure

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,3 +81,6 @@ Each implementation should be thoroughly covered with tests. DOM structure can i
 2. Start with simple component rendering tests, possibly using a snapshot
 3. Test additional interactivity / behaviour with explicit assertions
 4. Strive for 100% code coverage
+
+## Releasing
+The procedure for making a new release is [described here](RELEASING.md).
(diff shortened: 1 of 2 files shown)


carlobernardini:
When we decide on the new workflow we can change this accordingly but for now it's useful to have it documented.

eszthoff:
LGTM. Maybe it is a good time to make it even more complete. 
I think we should add something about merging (squashed) and commit messages for merges (what to write).  This might be part of the CONTRIBUTE.md, but I think it is related. 
Also might be good to add something about how to name branches after our last meeting ("creating branches"). 
Lastly, shouldn't storybook be published with every release?

carlobernardini:
> I think we should add something about merging (squashed) and commit messages for merges (what to write). This might be part of the CONTRIBUTE.md, but I think it is related.

I agree but that would be part of the new workflow (when we start using standard-version - I already do it that way but that's only by choice). I will update everything accordingly once we get to implement it. Then the release procedure should only be a matter of running `npm run release` and it's more important to describe squashing and commit messages, as you say.

> Also might be good to add something about how to name branches after our last meeting ("creating branches").

Did we discuss anything about that? I can't recall

> Lastly, shouldn't storybook be published with every release?

This is done automatically by the CI pipeline when a new tag is pushed, no additional instructions needed

PR #94, opened 2019-04-30
adjust heading variables

eszthoff:
[ONEUI-46](https://jira.textkernel.nl/browse/ONEUI-46)

### DIFF
--- a/src/components/Heading/Heading.scss
+++ b/src/components/Heading/Heading.scss
@@ -7,11 +7,12 @@
     }
     color: var(--color-foreground);
     text-align: left;
+    margin: var(--heading-margin) 0;
 
     &--level {
         @for $level from 1 to 7 {
             &_h#{ $level } {
-                font-size: var(--h#{ $level }-font-size);
+                font-size: var(--h#{$level}-font-size);
             }
         }
     }
--- a/src/components/Heading/Heading.scss
+++ b/src/components/Heading/Heading.scss
@@ -7,7 +7,7 @@
     }
     color: var(--color-foreground);
     text-align: left;
-    margin: var(--heading-margin) 0;
+    margin: 0;
 
     &--level {
         @for $level from 1 to 7 {
--- a/src/themes/oneui/base-variables/_typography.scss
+++ b/src/themes/oneui/base-variables/_typography.scss
@@ -1,64 +1,66 @@
 // fonts sizes
 $font-size-base: 14px !default;
-$font-size-small: .8 * $font-size-base !default;
+$font-size-small: 0.8 * $font-size-base !default;
 $font-size-normal: 1 * $font-size-base !default;
 $font-size-large: 1.2 * $font-size-base !default;
 
-$line-height-small:     1.4 * $font-size-small !default;
-$line-height-normal:    1.4 * $font-size-normal !default;
-$line-height-large:     1.4 * $font-size-large !default;
+$line-height-small: 1.4 * $font-size-small !default;
+$line-height-normal: 1.4 * $font-size-normal !default;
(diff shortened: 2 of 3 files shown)


src/components/Heading/Heading.scss
```diff
@@ -7,11 +7,12 @@
     }
     color: var(--color-foreground);
     text-align: left;
+    margin: var(--heading-margin) 0;
```
carlobernardini: I'm not sure if we should make this themable just yet, since it would affect **all** headings in the application. I think we should better set it to `0` without using variables for now.

src/themes/oneui/base-variables/_typography.scss
```diff
@@ -1,64 +1,66 @@
...
+$line-height-normal: 1.4 * $font-size-normal !default;
+$line-height-large: 1.4 * $font-size-large !default;
 
-$h1-font-size: 1.8 * $font-size-base !default;
-$h2-font-size: 1.6 * $font-size-base !default;
-$h3-font-size: 1.4 * $font-size-base !default;
-$h4-font-size: 1.2 * $font-size-base !default;
-$h5-font-size: 1.0 * $font-size-base !default;
-$h6-font-size: 0.8 * $font-size-base !default;
+$h1-font-size: 4 + $font-size-base !default;
```
carlobernardini: This sum looks quite random, you might as well just write `18px` (same for heading sizes below)

src/themes/oneui/base-variables/_typography.scss
```diff
@@ -1,64 +1,66 @@
 // fonts sizes
 $font-size-base: 14px !default;
-$font-size-small: .8 * $font-size-base !default;
+$font-size-small: 0.8 * $font-size-base !default;
 $font-size-normal: 1 * $font-size-base !default;
```
carlobernardini: If you choose to use a leading zero in decimal values (`0.8`) then it would make sense to use decimal notation everywhere (1.0).

src/themes/oneui/base-variables/_typography.scss
```diff
@@ -1,64 +1,66 @@
...
 // decoration
-$link-decoration-normal:    none !default;
-$link-decoration-hover:     underline !default;
+$link-decoration-normal: none !default;
+$link-decoration-hover: underline !default;
 
 // other
-$heading-font-family:   $font-family-secondary !default;
-$heading-font-weight:   $font-weight-bold !default;
+$heading-font-family: $font-family-primary !default;
```
carlobernardini: I would keep this referenced to `font-family-secondary` and set `font-family-secondary` to Open Sans.

src/themes/oneui/base-variables/_typography.scss
eszthoff: that was eslint, not me ;-)
But I agree, maybe better to use everywhere

src/themes/oneui/base-variables/_typography.scss
```diff
@@ -1,64 +1,64 @@
...
-$h1-font-size: 1.8 * $font-size-base !default;
-$h2-font-size: 1.6 * $font-size-base !default;
-$h3-font-size: 1.4 * $font-size-base !default;
-$h4-font-size: 1.2 * $font-size-base !default;
-$h5-font-size: 1.0 * $font-size-base !default;
-$h6-font-size: 0.8 * $font-size-base !default;
+$h1-font-size: 18px !default;
+$h2-font-size: 16px !default;
+$h3-font-size: 14px !default;
+$h4-font-size: 12px - 2 !default;
```
carlobernardini: Why these subtractions (h4 - h6)?

PR #95, opened 2019-04-30
render button as ancor when href is defined

eszthoff:
[ONEUI-44](https://jira.textkernel.nl/browse/ONEUI-44)

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -13,11 +13,15 @@ const { block } = bem({
 const Button = props => {
     const { children, context, disabled, isBlock, isInline, type, href, ...rest } = props;
 
-    return href ? (
-        <a {...rest} {...block(props)} hfer={href}>
-            {children}
-        </a>
-    ) : (
+    if (href) {
+        return (
+            <a {...rest} {...block(props)} href={href}>
+                {children}
+            </a>
+        );
+    }
+
+    return (
         <button {...rest} {...block(props)} type={type} disabled={disabled}>
             {children}
         </button>
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -11,9 +11,13 @@ const { block } = bem({
 });
 
 const Button = props => {
-    const { children, context, disabled, isBlock, isInline, type, ...rest } = props;
+    const { children, context, disabled, isBlock, isInline, type, href, ...rest } = props;
 
-    return (
+    return href ? (
+        <a {...rest} {...block(props)} hfer={href}>
+            {children}
+        </a>
+    ) : (
         <button {...rest} {...block(props)} type={type} disabled={disabled}>
             {children}
         </button>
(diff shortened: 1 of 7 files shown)


src/components/Button/Button.js
```diff
@@ -11,9 +11,13 @@ const { block } = bem({
 });
 
 const Button = props => {
-    const { children, context, disabled, isBlock, isInline, type, ...rest } = props;
+    const { children, context, disabled, isBlock, isInline, type, href, ...rest } = props;
 
-    return (
+    return href ? (
+        <a {...rest} {...block(props)} hfer={href}>
```
carlobernardini: Typo (hfer)

src/components/Button/Button.js
```diff
@@ -11,9 +11,13 @@ const { block } = bem({
...
+            {children}
+        </a>
+    ) : (
```
carlobernardini: I think using a shorthand `if` is not very readable in this case. Better use two `return` statements and an explicit `if`.

```
if (href) {
   return (anchor);
}

return (button);
```

src/components/Button/__tests__/Button.spec.js
```diff
@@ -35,4 +36,10 @@ describe('<Button> that renders a button', () => {
         buttonWrapper.simulate('mouseover');
         expect(onMouseOverMock).toHaveBeenCalled();
     });
+    it('should render an ancor element if href is defined', () => {
+        const wrapper = mount(<Button href="/">Click me</Button>);
+        expect(toJson(wrapper)).toMatchSnapshot();
+        expect(wrapper.find('button')).toHaveLength(0);
```
carlobernardini: Is this assertion necessary? If it finds an `a`, doesn't that implicitly mean it does not find a `button`?

stories/Button.js
```diff
@@ -17,6 +17,7 @@ storiesOf('Button', module)
             onClick={e => {
                 console.log('Clicked button', e);
             }}
+            href={text('Href to be rendere for links', '')}
```
carlobernardini: Typos.

"Href (will render an anchor)"

carlobernardini:
Now that `Button` can render an `<a />` you most likely have to add a few more styles to have it appear as intended. An anchor is an inline element (should be block in this case, otherwise I think paddings won't show) and comes with underline text-decoration, for example.

src/components/Button/Button.js
```diff
@@ -38,7 +46,9 @@ Button.propTypes = {
     /** Should button be disabled or not */
     disabled: PropTypes.bool,
     /** Type of the button */
-    type: PropTypes.oneOf(['submit', 'button'])
+    type: PropTypes.oneOf(['submit', 'button']),
+    /** Providing an href will render an <a> element, styled as a button. */
+    href: PropTypes.string
```
carlobernardini: Alphabetical order would be nice

eszthoff: did we actually agreed on that?

carlobernardini: At least it was mentioned more than a few times... We can't be adding all of those details in a 'guideline'. It's a matter of detail but it would be nice to keep in mind.

PR #98, opened 2019-05-10
fix(a11y): accessibility improvements

carlobernardini:
Various accessibility improvements

### DIFF
--- a/src/components/Tabs/TabItem/TabItem.js
+++ b/src/components/Tabs/TabItem/TabItem.js
@@ -14,11 +14,11 @@ const TabItem = props => {
     return (
         <a
             href={href || '#'}
+            tabIndex={0}
             {...rest}
             {...block(props)}
             role="tab"
             aria-selected={isActive}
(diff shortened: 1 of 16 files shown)


src/components/Tabs/TabItem/TabItem.js
```diff
@@ -12,7 +12,14 @@ const { block } = bem({
...
     const { href, isActive, label, ...rest } = props;
     return (
-        <a href={href || '#'} {...rest} {...block(props)}>
+        <a
+            href={href || '#'}
+            {...rest}
+            {...block(props)}
+            role="tab"
+            aria-selected={isActive}
+            tabIndex={0}
```
eszthoff: maybe better to have tabIndex taken from the props, and only if it isn't define make it 0? Or simpler yet, move it before `{...rest}` then it can be overwritten

carlobernardini: Valid point, I think moving it before `rest` should be good

PR #99, opened 2019-05-10
Fix: move general dependencies to peer-dependencies

eszthoff:
Fix: move react, react-dom and prop-types to be peer-dependencies and not to bundle with package.

### DIFF
--- a/etc/webpack/webpack.config.js
+++ b/etc/webpack/webpack.config.js
@@ -3,6 +3,7 @@ const webpack = require('webpack');
 const OptimizeCssAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const { getRuleJS, getRuleCSS } = require('./utils');
 
 const PROJECT_ROOT_PATH = path.resolve(__dirname, '../../');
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -40,10 +40,12 @@
...
     "dependencies": {
         "@babel/polyfill": "7.0.0",
-        "css-vars-ponyfill": "1.16.4",
-        "prop-types": "15.6.2",
-        "react": "16.8.6",
-        "react-dom": "16.8.6"
+        "css-vars-ponyfill": "1.16.4"
+    },
+    "peerDependencies": {
+        "prop-types": ">=15.6.2",
```
carlobernardini: I'm not sure this is right, this would mean OneUI is also compatible with new major versions (with breaking changes) of these packages?

Shouldn't it be like `^15.6.2` instead? (Same for dependencies below)

carlobernardini:
It's probably preferable to have a separate PR for the changes to the pagination component otherwise they'll be part of the same commit when the PR is squashed (actually they already are)

package.json
```diff
@@ -41,9 +41,11 @@
     "dependencies": {
         "@babel/polyfill": "7.0.0",
         "css-vars-ponyfill": "1.16.4",
-        "prop-types": "15.6.2",
-        "react": "16.8.6",
-        "react-dom": "16.8.6"
+        "prop-types": "^15.6.2"
```
carlobernardini: Prop-types is now listed in both in dependencies and devDependencies, is that intended?

etc/webpack/webpack.config.js
```diff
@@ -3,6 +3,7 @@ const webpack = require('webpack');
 const OptimizeCssAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
+const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
```
SebastiaanPasterkamp: Let's put the bundle analyzer in a different PR. It's related to the _investigation_ of omitting the react libraries from the bundle, but technically a different (development) feature and irrelevant for users. The exclusion of the react libraries _is_ relevant for users.

src/components/Pagination/Pagination.js
```diff
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
```
SebastiaanPasterkamp: Let's put these changes in a different PR. This is a bug-fix to make Pagination work in ngui, while putting react in devDependencies is not.

PR #101, opened 2019-05-14
chore(storybook): use latest minor versions instead of exact versions

carlobernardini:
Change Storybook dependencies to always install latest minor version instead of exact version, so we don't get alerts about new versions all the time. New major versions should still be upgraded to manually.


package.json
```diff
@@ -54,11 +54,11 @@
...
-        "@storybook/addon-console": "1.1.0",
-        "@storybook/addon-info": "5.0.10",
-        "@storybook/addon-knobs": "5.0.10",
-        "@storybook/react": "5.0.10",
+        "@storybook/addon-actions": "^5.0.11",
+        "@storybook/addon-console": "^1.1.0",
+        "@storybook/addon-info": "^5.0.11",
+        "@storybook/addon-knobs": "^5.0.11",
+        "@storybook/react": "^5.0.11",
         "@storybook/storybook-deployer": "2.8.1",
```
eszthoff: why not this one too? In general I wonder why we don't have `^` for all dependencies

carlobernardini: Because for Storybook I'm quite confident that nothing would break when it is updated to a new minor version, and even if it would, it wouldn't affect the actual library. For other dependencies that's a bit different as those impact the actual code / build. For those I don't think we should update and check ourselves to make sure everything still works as intended. We can still agree to change it for other dependencies too but for now I'd rather keep it only to Storybook (because it's notifying about new versions all the time).

PR #102, opened 2019-05-14
fix: uncapitalize oneui util and fix example app

carlobernardini:
Lowercase the OneUI util file name and load correct stylesheet in example app


eszthoff:
I think we should stick to naming files the same as their exports. Perhaps in this case, since it is only `default export` it can be confusing as the name is not that important. However since we advocate for `import OneUI from '@textkernel/oneui'` syntax, I would keep the file name with capital too.

carlobernardini:
Well then we should change the ones in `themes/` because this doesn't look consistent

Screenshot 2019-04-23 at 17 02 27

eszthoff:
we haven't quite decided for naming convention for themes. Maybe we can discuss it at some point in the OneUI meeting and then based on that we can keep/change the file name for that file. But for the util, I would keep it as it is.

carlobernardini:
Let's discuss tomorrow then. Either one of two will need to be renamed though.

PR #103, opened 2019-05-15
feat(link): add context to link component

eszthoff:
[JF-1518](https://jira.textkernel.nl/browse/JF-1518)

### DIFF
--- a/src/components/Link/Link.js
+++ b/src/components/Link/Link.js
@@ -1,15 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import bem from 'bem';
+import { CONTEXTS } from '../../constants';
 import styles from './Link.scss';
 
 const { block } = bem({
     name: 'Link',
-    classnames: styles
+    classnames: styles,
+    propsToMods: ['context']
 });
 
 const Link = props => {
-    const { children, ...rest } = props;
+    const { children, context, ...rest } = props;
     return (
         <a {...rest} {...block(props)}>
             {children}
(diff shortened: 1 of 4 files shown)


src/components/Link/Link.js
```diff
@@ -20,8 +22,14 @@ const Link = props => {
 Link.displayName = 'Link';
 
 Link.propTypes = {
+    /** The spinner context (e.g. brand, primary, bad, good etc.) */
+    context: PropTypes.oneOf(['default', 'muted', ...CONTEXTS]),
```
carlobernardini: Let's have only 'neutral' for now, until we really need more contexts.

Also, we currently don't have a 'default' context. The component has a default styling of its own (no prop applied), which is not a 'context'. This is also how its implemented in other components and I'd like to stick with that.

src/components/Link/Link.scss
```diff
@@ -13,4 +13,24 @@
         color: var(--link-color-hover);
         text-decoration: var(--link-decoration-hover);
     }
+
+    &--context {
+        @each $context in (brand, neutral, primary, accent, info, good, warning, bad) {
```
SebastiaanPasterkamp: Would it not make sense to define this repeating array as a variable, kind of like what `mixins/colors.scss` does? The react code already does that. IMHO it's better to have context behave consistently between different components, instead of having to think "which context do we _really_ need now". And to find out later that some existing component implements all colors _except_ the one you need...

src/components/Link/Link.scss
```diff
@@ -13,4 +13,24 @@
...
+                    color: var(--color-#{$context}-40);
+                }
+            }
+        }
+
+        &_muted {
+            color: var(--color-muted);
+
+            &:hover {
+                color: var(--color-neutral);
```
SebastiaanPasterkamp: Why is there no `--color-muted-40` :thinking: , even it literally is defined as `--color-neutral`?

@carlobernardini Perhaps it's even better to define a `--color-#{$context}-hover` to have all `:hover` colors consistent?

PR #105, opened 2019-05-17
feat: chip component

eszthoff:
[[ONEUI-52]](https://jira.textkernel.nl/browse/ONEUI-52) Added a component to render simple chips

### DIFF
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -9,10 +9,10 @@ const { block } = bem({
 });
 
 const Chip = props => {
-    const { children, title, ...rest } = props;
+    const { children, ...rest } = props;
 
     return (
-        <span title={title} {...rest} {...block(props)}>
+        <span {...rest} {...block(props)}>
             {children}
         </span>
     );
@@ -22,13 +22,7 @@ Chip.displayName = 'Chip';
 
 Chip.propTypes = {
     /** The content of the chip: it is expected to be short text possibly accompanied by an icon */
-    children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
-    /** Text for title chip that shows up when it is hovered for a few seconds */
-    title: PropTypes.string
-};
-
-Chip.defaultProps = {
-    title: null
+    children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired
 };
 
 export default Chip;
--- a/src/components/Chip/Chip.js
+++ b/src/components/Chip/Chip.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem';
+import styles from './Chip.scss';
+
+const { block } = bem({
+    name: 'Chip',
+    classnames: styles
+});
+
+const Chip = props => {
+    const { children, title, ...rest } = props;
+
+    return (
+        <span title={title} {...rest} {...block(props)}>
+            {children}
+        </span>
+    );
+};
+
+Chip.displayName = 'Chip';
+
+Chip.propTypes = {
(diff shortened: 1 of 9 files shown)


stories/Chip.js
```diff
@@ -0,0 +1,13 @@
...
+import { storiesOf } from '@storybook/react';
+import { text, withKnobs } from '@storybook/addon-knobs';
+import { Chip, IconMatch } from '@textkernel/oneui';
+
+storiesOf('Chip', module)
+    .addDecorator(withKnobs)
+    .add('Chip', () => (
+        <Chip title={text('Title', 'This is a title that uses native browser functionality')}>
+            <IconMatch />
+            {text('Content', ' This is a chip')}
```
SebastiaanPasterkamp: Without the space the icon padding is missing. Would it make sense to support icons natively for a component like this? Or at least change the example to `<IconMatch />&nbsp;`.

eszthoff: I think there is a point in that. I would however suggest to make a component that renders an icon and adds padding on right or left. Then that can be more easily reused in any component or just text. If we add it in the context of every component, it is more complicated, I believe, with at least 2 or 3 props needed... If you agree, I will make a ticket for it in ONEUI.

src/components/Chip/Chip.scss
```diff
@@ -0,0 +1,6 @@
+.Chip {
+    font-size: var(--font-size-small);
```
carlobernardini: Let's increase the size of `--font-size-small`, it's current size is hardly readable. 12 or 13 pixels would be better.

src/components/Chip/Chip.scss
```diff
@@ -0,0 +1,6 @@
...
+    padding: var(--spacing-2x) var(--spacing-5x);
+    background-color: var(--color-neutral);
```
carlobernardini: Let's make the background color a bit lighter, I checked and I think `--color-neutral-25` should be good

src/components/Chip/Chip.scss
```diff
@@ -0,0 +1,6 @@
+.Chip {
```
carlobernardini: Since you're using `<span />` you probably need to give it `display: inline-block` too, otherwise the box will not size properly

Screenshot 2019-05-20 at 12 51 39

carlobernardini:
How will you deal with spacing in between chips, so they're not attached to one another?

src/components/Chip/Chip.js
```diff
@@ -0,0 +1,34 @@
...
+    );
+};
+
+Chip.displayName = 'Chip';
+
+Chip.propTypes = {
+    /** The content of the chip: it is expected to be short text possibly accompanied by an icon */
+    children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+    /** Text for title chip that shows up when it is hovered for a few seconds */
+    title: PropTypes.string
```
carlobernardini: Wouldn't it make more sense for this to be an 'undocumented prop' that just comes from `...rest`? Eventually the idea is for those chips to have 'real' tooltips rather than html ones anyway

eszthoff: I think it is better to have it documented, exactly so that it can be easily changed to use a tooltip if we want. Then we can just change the internal implementation and the code in JF, Search doesn't have to be changed. So title will be used for the tooltip. If you think that prop should have an other name, we can change it already so that it fills more universal. Any suggestions?

carlobernardini: I think eventually the tooltip would not be managed by those components themselves (i.e. there would not be a prop for giving specific components a tooltip). Probably we shouldn't even bother with making tooltips of our own (only the styling). Most tooltip libraries (e.g. [react-tooltip](https://github.com/wwayne/react-tooltip#usage)) use attributes that you have to add to any component or element you want to give a tooltip.

PR #106, opened 2019-05-20
fix(button): fix height issues with anchor-buttons in safari

carlobernardini:
Fix issue with buttons that use `<a />` tags which in Safari have a different height from regular buttons

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -7,6 +7,7 @@
         width: var(--line-normal);
         style: solid;
     }
+    box-sizing: content-box;
     color: var(--color-brand);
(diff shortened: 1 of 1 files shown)


eszthoff:
Would it make sense add a story about it to the ButtonGroup, so that it is visible and easier to see?

carlobernardini:
> Would it make sense add a story about it to the ButtonGroup, so that it is visible and easier to see?

Not sure what you mean. There was no issue with ButtonGroup, only with Buttons using anchors. I don't think Storybook is meant to demonstrate there are no bugs in (combinations of) our components but rather to provide examples and documentation.

PR #110, opened 2019-05-23
feat: add textkernel full logo

eszthoff:


src/components/Icon/IconBase.js
```diff
@@ -10,22 +10,25 @@ const { block, elem } = bem({
     propsToMods: ['context', 'margin']
 });
 
-const adjustSize = size => {
+const adjustSize = (size, keepWidth) => {
```
carlobernardini: Maybe it makes more sense to have something like `preserveAspectRatio` instead?

src/components/Icon/IconBase.js
```diff
@@ -10,22 +10,25 @@ const { block, elem } = bem({
...
     if (!size) {
-        return null;
+        if (!keepWidth) {
+            return null;
+        }
+        return { width: 'auto' };
```
carlobernardini: Does this extra condition make a difference with returning `null`? (i.e. won't it be 'auto' implicitly / is there a need to have that extra `!keepWidth` check there?)

src/components/Icon/IconTextkernelFull/__tests__/IconTextkernelFull.spec.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import toJson from 'enzyme-to-json';
+import IconTextkernelFull from '../IconTextkernelFull';
```
carlobernardini: `IconTextkernelLogo` maybe nicer?

src/components/Icon/IconBase.js
eszthoff: great idea

src/components/Icon/IconBase.js
eszthoff: the default css defines it to `1em`, hence the check. probably would be clearer to move that part of the css here, so that it is all in one place.

src/components/Icon/IconTextkernelFull/__tests__/IconTextkernelFull.spec.js
eszthoff: ok

carlobernardini:
I was looking at the stories of 'Textkernel Icon' and 'Textkernel Icon with full name' (btw. maybe its better to rename that last one to 'Textkernel Logo') in Storybook and what I don't understand is that according to the story source the component name and attributes are exactly the same, yet they render something different. How is that possible and shouldn't the full logo be `<LogoTextkernel />`?

src/components/Icon/LogoTextkernel/LogoTextkernel.js
```diff
@@ -0,0 +1,42 @@
...
+        <path d="M208.8,67.4l-17-23.5v23.5h-11.6V0h11.6v39l15.4-18.5h12.5l-17,20.3l19.5,26.6H208.8z" />
+        <path d="M240,48.4c0.9,7.5,4.8,11.7,11.3,11.7c4.4,0,9.7-1.5,14.7-4l3.8,8.9c-5.6,2.6-12.8,3.7-19.3,3.7c-15.2,0-22.7-11.3-22.7-24.3c0-10.9,4.9-25.2,22.3-25.2c16.3,0,22.1,9.6,22.1,29.2H240z M250,28.1c-3.8,0-9.5,2.3-9.9,12h21C260.5,31.2,256,28.1,250,28.1z" />
+        <path d="M311.5,30.9c-1.7-1.3-4.2-2.9-6.8-2.9c-5.4,0-8.4,8.6-9.5,11.4v28h-11.6V31c0-4.6-0.9-8.1-1.2-10.5H293l1.8,8c1.6-3,6.7-9.4,11.7-9.4c3.9,0,6.2,0.8,8.9,2.3L311.5,30.9z" />
+        <path d="M352.7,67.4v-30c0-4.6-0.4-9.3-7.3-9.3c-4.8,0-8.1,4.4-10.3,8.8v30.6h-11.6V30.5c0-3.9-0.8-6.7-1.2-9.9H333l1.8,6.2c1.9-2.1,5.9-7.6,12.5-7.6c13.7,0,16.9,8,16.9,18.8v29.5H352.7z" />
+        <path d="M387.7,48.4c0.9,7.5,4.8,11.7,11.3,11.7c4.4,0,9.7-1.5,14.7-4l3.8,8.9c-5.6,2.6-12.8,3.7-19.3,3.7c-15.2,0-22.7-11.3-22.7-24.3c0-10.9,4.9-25.2,22.3-25.2c16.3,0,22.1,9.6,22.1,29.2H387.7z M397.6,28.1c-3.8,0-9.5,2.3-9.9,12h21C408.2,31.2,403.7,28.1,397.6,28.1z" />
+        <path d="M431.2,67.4V0h11.6v67.4H431.2z" />
+    </IconBase>
+);
+
+LogoTextkernel.displayName = 'IconTextkernel';
```
carlobernardini: I think this `displayName` is meant to be `LogoTextkernel`

stories/Icon.js
```diff
@@ -73,4 +74,11 @@ storiesOf('Icons', module)
             size={number('Size', SIZE_DEFAULT)}
             title={text('Title', 'Textkernel')}
         />
+    ))
+    .add('Textkernel icon with full name', () => (
```
carlobernardini: Textkernel logo

PR #112, opened 2019-06-04
test: make jest use real style declarations instead of mocked ones

antipin:
Resolves #ONEUI-4


etc/test/jest.config.js
```diff
@@ -0,0 +1,33 @@
...
+        '^.+\\.jsx?$': 'babel-jest',
+        '^.+\\.scss$': 'jest-css-modules-transform'
+    },
+
+    // An array of glob patterns indicating a set of files for which coverage information should be collected.
+    // If a file matches the specified glob pattern, coverage information will be collected for it even if no tests
+    // exist for this file and it's never required in the test suite.
+    collectCoverageFrom: ['**/*.{js,jsx}'],
+
+    coveragePathIgnorePatterns: ['/node_modules/', 'dummy-components']
```
carlobernardini: Should this perhaps also include `/example`?

carlobernardini:
@antipin one comment, LGTM otherwise 👍

etc/test/jest.config.js
antipin: Seems like that the only thing that affects coverage score here is `'dummy-components'` pattern, probably because of `roots: ['<rootDir>/src']`. So we can even remove `/node_modules/` :)

PR #116, opened 2019-06-05
build: enhanced dev and prod build modes

antipin:
* Added 'clean-webpack-plugin' to clead dist directory before build
* Renamed 'etc/webpack' to 'etc/build' to make it more consistent with other etc dirs
* Set bundle name to 'oneui' for dev build and 'oneui.min' for production build

Resolves #ONEUI-53.


carlobernardini:
@antipin is this still WIP?

antipin:
> @antipin is this still WIP?

not anymore :)

carlobernardini:
Two comments / questions:

1. Regarding the PR title, are you sure `build` is a valid type (like fix, feat)?
2. Can we maybe fix this bug in the example app as part of this PR?
Screenshot 2019-06-26 at 09 45 38

I already worked on a fix for the example app in #102 but that has other changes that still need discussion..

Other than that it works for me :)

antipin:
@carlobernardini 
> Regarding the PR title, are you sure build is a valid type (like fix, feat)?
I believe it's valid. I use this cheatsheet for choosing types:
image
But yeah, I don't see it in the list from CONTRIBUTING.mg.
I think it make sense to add this type :)

antipin:
> Can we maybe fix this bug in the example app as part of this PR?

Oke, I'll have a look. How can I reproduce it? Just `npm run start`?

carlobernardini:
> Oke, I'll have a look. How can I reproduce it? Just npm run start?

Yep. It's just a matter of `.min.` missing in the css import, if I recall correctly.

antipin:
@carlobernardini ready

carlobernardini:
@eszthoff I don't know if this is something you need to be aware of (for JF implementation) but please note that this will make `npm run build` produce `oneui.js` instead of `oneui.min.js` (so, without `.min`).

eszthoff:
I don't really understand the motivation here. If the prod build is minified why not to name is so? Or you don't want to minify it at all? Anyway, I don't think that would effect js imports. If the css file name changes, that would be a breaking change. But I don't have time to test.

carlobernardini:
I think the motivation it quite clearly outlined in the PR description:
1. Clean the `dist` folder before build to make sure there are no leftover from previous builds (those may end up being included with the published package, as happened before)
2. Dev build is not minified (which makes sense), thus filename shouldn't say `.min`. As opposed to production build, that does contain `.min` because it is actually minified.

eszthoff:
@carlobernardini that makes sense indeed. Only your previous comment said that prod build will not named with `min`. Hence the confusion. If only dev env changes, JF should not be effected.

carlobernardini:
> Only your previous comment said that prod build will not named with min.
I said `npm run build` (dev build) would produce filename without `.min` - production build is `npm run build:prod`.

Then I think we all agree this is approved.

PR #119, opened 2019-06-11
fix(footer): add styling to links

eszthoff:


src/components/Footer/Footer.scss
```diff
@@ -8,6 +8,14 @@
     border-bottom: 1px solid var(--border-color-strong);
     box-shadow: 0 1px 3px var(--border-color-light);
 
+    & a {
```
carlobernardini: `&` is not necessary here

PR #123, opened 2019-06-16
test: extend tests for uncovered lines

carlobernardini:
* Remove default `null` value from destructuring assignment in Pagination component that would never occur and cannot be tested
* Extend tests for Tabs component when used without `onChange` callback prop
* Extend tests for OneUI util: check that error is thrown when timeout exceeded

### DIFF
--- a/src/components/Pagination/Pagination.js
+++ b/src/components/Pagination/Pagination.js
@@ -52,7 +52,7 @@ const Pagination = props => {
 
     const handleClick = e => {
         const {
-            dataset: { page = null }
+            dataset: { page }
         } = e.target;
         const pageNum = +page;
 
--- a/src/components/Tabs/__tests__/Tabs.spec.js
+++ b/src/components/Tabs/__tests__/Tabs.spec.js
@@ -4,7 +4,9 @@ import Tabs from '../Tabs';
 import Tab from '../Tab';
 
 describe('<Tabs> that renders tab container with some tabs', () => {
-    it('should render simple tabs', () => {
+    it('should render simple tabs (without onChange callback prop)', () => {
+        const onChange = jest.fn();
+
         const wrapper = mount(
             <Tabs activeTabId="1">
                 <Tab id="1" label="First">
@@ -16,6 +18,14 @@ describe('<Tabs> that renders tab container with some tabs', () => {
             </Tabs>
         );
         expect(toJson(wrapper)).toMatchSnapshot();
+
+        wrapper
+            .find('TabMenu')
+            .find('TabItem')
+            .at(1)
+            .simulate('click');
+
+        expect(onChange).not.toHaveBeenCalled();
     });
 
     it('should switch tabs correctly', () => {
--- a/src/utils/OneUI/OneUI.js
+++ b/src/utils/OneUI/OneUI.js
@@ -16,7 +16,7 @@ class OneUI {
      * @param {number} themeConfig.maxTime - Max time to wait for the theme to be loaded
      * @param {Object} themeConfig.ponyfillOptions - Set of options that can be used to configure the ponyfill. Options: https://www.npmjs.com/package/css-vars-ponyfill#options
      */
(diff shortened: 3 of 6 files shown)


src/components/Tabs/__tests__/Tabs.spec.js
```diff
@@ -16,6 +18,14 @@ describe('<Tabs> that renders tab container with some tabs', () => {
...
         );
         expect(toJson(wrapper)).toMatchSnapshot();
+
+        wrapper
+            .find('TabMenu')
+            .find('TabItem')
+            .at(1)
+            .simulate('click');
+
+        expect(onChange).not.toHaveBeenCalled();
```
eszthoff: I'm not sure there is a point in this test. you create a variable you never use and then ask if anything happened to it?

src/utils/OneUI/OneUI.js
```diff
@@ -16,7 +16,7 @@ class OneUI {
      * @param {number} themeConfig.maxTime - Max time to wait for the theme to be loaded
      * @param {Object} themeConfig.ponyfillOptions - Set of options that can be used to configure the ponyfill. Options: https://www.npmjs.com/package/css-vars-ponyfill#options
      */
-    static init({ themeURL = '', maxTime = DEFAULT_LOADING_TIMEOUT, ponyfillOptions } = {}) {
+    static init({ themeURL = '', maxTime = DEFAULT_LOADING_TIMEOUT, ponyfillOptions }) {
```
eszthoff: removing this default, would make `OneUI.init()` throw an error, wouldn't it? Probably there should be a test for it, since that's the simplest implementation, but I cannot see any at the moment.

src/components/Pagination/Pagination.js
```diff
@@ -52,7 +52,7 @@ const Pagination = props => {
 
     const handleClick = e => {
         const {
-            dataset: { page = null }
+            dataset: { page }
```
eszthoff: why remove this default?

carlobernardini: I was thinking the default doesn't make much sense because I don't think `data-page` can ever be `undefined` and cannot be affected from outside this component. And I don't see how it can be tested.

src/components/Tabs/__tests__/Tabs.spec.js
carlobernardini: Yes, now that I'm looking at it again it doesn't seem to make much sense indeed. I'll see if I can think of something else (purpose was to check that `handleTabChange` would not trigger `onChange` even when the callback prop is not defined).

src/utils/OneUI/OneUI.js
carlobernardini: Hmm I may have been wrong here. I thought that default was superfluous since the destructured keys have defaults of their own but the default I removed was probably there in case `init` would receive a falsey argument, right?

carlobernardini:
If we're not confident that these changes improve code / tests I'm fine with closing the PR. At least I agree that there still are a few issues that need attention.

src/components/Pagination/Pagination.js
eszthoff: in some funny edge case through some unforeseen event propagation? anyway, what is the harm in having that fallback?

src/components/Tabs/__tests__/Tabs.spec.js
eszthoff: check that the click doesn't throw... but move it to a separate test, not part of initial rendering

src/components/Pagination/Pagination.js
carlobernardini: No although it may look like that it wasn't necessarily my intention to reach 100% but in this case looking at coverage can be useful to spot redundant code which I thought would be the case here. But I think you're right that it doesn't do much harm either. I will revert it (if choose to keep some other changes in this branch).

carlobernardini:
Closing this PR, it's not clear enough that these changes would result in (significant) improvements.

PR #127, opened 2019-06-25
feat(List, ListItem): add forwardRef option

eszthoff:


carlobernardini:
@eszthoff can you provide a bit of context?

eszthoff:
@carlobernardini I need to use ref for the list when implementing the multiselect in JF.

carlobernardini:
@eszthoff Okay. I'm not sure `chore` would be the most appropriate type though. Also, [according to the React docs](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers) this should formally be treated as a breaking change (even though for now it wouldn't make a big difference)

eszthoff:
BREAKING CHANGES: forwardRef applied to List component

PR #128, opened 2019-06-26
docs: extend documentation about releasing and commit message format

carlobernardini:
* Add note about doing a dry run first when releasing
* Make it more clear how to indicate breaking changes in commit message
* Add type `build` to commit message format docs

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ OneUI's [release procedure](RELEASING.md) and [changelog](CHANGELOG.md) automati
 [optional footer]
 ```
 
-Commit types include:
+Commit types (required) include:
 * `feat`: A new feature (equivalent to `MINOR`)
 * `fix`: A bug fix (equivalent to `PATCH`)
 * `docs`: Changes to documentation
@@ -105,10 +105,11 @@ Commit types include:
 * `style`: Code style change (indentation, semicolons, ...)
 * `perf`: Performance enhancements
 * `refactor`: Code refactoring without changes to public API
+* `build`: Changes that affect the build system or external dependencies
 
 Use imperative, present tense in your commit description ("change", not "changed" or "changes") without uppercases or period (.) at the end.
 
-**Breaking changes** should be explicitly marked (with uppercase) in the message footer, e.g. `BREAKING CHANGE: <note>`. This will result in a new `MAJOR` package version.
+**Breaking changes** should be explicitly marked (singular, with uppercase) in the message footer, e.g. `BREAKING CHANGE: <note>`. This will result in a new `MAJOR` package version.
 
 ### Merging
 When merging a branch, squash its commits for the changelog to be nice and tidy. Each pull request / squashed commit should eventually equal one fix / feature / change in docs or tests.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -16,7 +16,18 @@ Version numbers should not be prefixed (e.g. 1.2.3, **not** v1.2.3).
 
 ## Release Procedure
 
-To make a new release, simply run **`npm run release`**.
+**First**, make sure your master branch is up to date and do a dry-run to verify that the expected version number will be used:
+
+```bash
+(master)$ git pull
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -105,10 +105,11 @@ Commit types include:
 * `style`: Code style change (indentation, semicolons, ...)
 * `perf`: Performance enhancements
 * `refactor`: Code refactoring without changes to public API
+* `build`: Changes that affect the build system or external dependencies
 
 Use imperative, present tense in your commit description ("change", not "changed" or "changes") without uppercases or period (.) at the end.
 
-**Breaking changes** should be explicitly marked (with uppercase) in the message footer, e.g. `BREAKING CHANGE: <note>`. This will result in a new `MAJOR` package version.
+**Breaking changes** should be explicitly marked (singular, with uppercase) in the message footer, e.g. `BREAKING CHANGE: <note>`. This will result in a new `MAJOR` package version.
```
eszthoff: perhaps make the title singular too

RELEASING.md
```diff
@@ -16,7 +16,18 @@ Version numbers should not be prefixed (e.g. 1.2.3, **not** v1.2.3).
 
 ## Release Procedure
 
-To make a new release, simply run **`npm run release`**.
+**First**, make sure your master branch is up to date and do a dry-run to verify that the expected version number will be used:
+
+```bash
+(master)$ git pull
+(master)$ npm run release -- --dry-run
+```
```
eszthoff: What if the dry-run gives an undesired result?

CONTRIBUTING.md
```diff
@@ -95,7 +95,7 @@ OneUI's [release procedure](RELEASING.md) and [changelog](CHANGELOG.md) automati
 [optional footer]
 ```
 
-Commit types include:
+Commit types (required) include:
```
eszthoff: I find it a bit confusing... we have the PR title set like this. Do we need to repeat the PR title in the squash commit message? I guess only one of these is indeed necessary. Would be good to clarify which.

RELEASING.md
carlobernardini: Then you should investigate why

CONTRIBUTING.md
carlobernardini: When squashing and merging a PR in Github the title is transferred to the squash and merge dialog. So if your PR title already contains what you want your (squashed) commit title to be you should be good (apart from the body / footer). Still, you need to be careful and double check that yourself, as with everything. We can document every single thing but in general you should always be careful about what you do.

CONTRIBUTING.md
carlobernardini: It says exactly how you should write it in the example so assuming everyone actually reads the text I think it's sufficiently described. The title does not serve as an example, otherwise I would also have to uppercase it.

CONTRIBUTING.md
eszthoff: I don't think that's correct. See my PR yesterday. It transferred the text of the commit to the squash, which wasn't named according to convention.

RELEASING.md
eszthoff: I meant it would be good to add a link or something how can you manually change the changelog or the version number if needed.

CONTRIBUTING.md
carlobernardini: I'm pretty sure that's correct because it has worked for me that way each time I squashed and merged PRs.

squash-merge

So something else must have gone wrong in your particular case. Either way we should always be careful with what we do no matter to which extent we document things.

eszthoff: please note that you have the same message for your commit as for the PR title.

carlobernardini: Well, that's exactly what I expect

RELEASING.md
carlobernardini: We should really *not* do that, everyone should make sure that commit messages are carefully formatted. If the version number is incorrect that is because some commits have not been formatted appropriately.

PR #130, opened 2019-07-05
feat: add container variables to make it a bit responsive

eszthoff:
BREAKING CHANGE: variable `--site-container-size` is deprecated in favor of `--site-container-min-size`, `--site-container-max-size` and `--site-container-padding`. Recommended new usage  through `<PageWidthRestrictor>`.


eszthoff:
@antipin done as recommended

PR #140, opened 2019-07-17
1460 keyboard navigation

igorarkhipenko:
* Add keyboard navigation
* Add tests coverage
* Fix color shades for the selected and highlighted items in the list

### DIFF
--- a/src/components/List/List.js
+++ b/src/components/List/List.js
@@ -1,8 +1,12 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import bem from 'bem';
 import ListItem from './ListItem';
 import styles from './List.scss';
+import { LIST_NAVIGATION_DIRECTIONS, ENTER_KEY } from '../../constants';
 
 const { block, elem } = bem({
     name: 'List',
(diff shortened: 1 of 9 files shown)


src/themes/oneui/base-variables/_colors.scss
```diff
@@ -67,6 +67,6 @@ $color-bad: #f42534 !default;
     --border-color-light: var(--color-neutral-25);
     --border-color-strong: var(--color-neutral-40);
 
-    --color-hover-background: #{mix($color-background, $color-brand, 90)};
-    --color-highlight-background: #{mix($color-background, $color-brand, 90)};
+    --color-highlight-background: #{var(--color-info-10)};
```
antipin: sees like unnecessary mix of sass and css syntax here
can be:
```
    --color-highlight-background: var(--color-info-10);
```

src/themes/oneui/base-variables/_colors.scss
```diff
@@ -67,6 +67,6 @@ $color-bad: #f42534 !default;
     --border-color-light: var(--color-neutral-25);
     --border-color-strong: var(--color-neutral-40);
 
-    --color-hover-background: #{mix($color-background, $color-brand, 90)};
-    --color-highlight-background: #{mix($color-background, $color-brand, 90)};
+    --color-highlight-background: var(--color-info-10);
```
antipin: why have you removed `mix` function here?

igorarkhipenko: @antipin I replaced it with a slightly brighter color so the item would have visible difference between `highlighted` and `selected` style options, but I'm going to discuss this part in details with David so let it hang here for a while please

src/components/List/List.js
```diff
@@ -1,21 +1,94 @@
...
+        // Update selectedIndex with arrow navigation and make onNavigate function callback
+        if (e.key === LIST_NAVIGATION_DIRECTIONS.UP || e.key === LIST_NAVIGATION_DIRECTIONS.DOWN) {
+            const nextSelectedIndex = getNextSelectedIndex(e.key);
+
+            if (selectedIndex !== nextSelectedIndex) {
+                e.preventDefault();
+                setSelectedIndex(getNextSelectedIndex(e.key));
+
+                if (onNavigate) {
+                    onNavigate(e.key);
```
eszthoff: It might be useful to return the nextIndex as well. Then the consuming component doesn't need to recalculate it. Perhaps can return an object with `{ navKey: e.key, nextSelectedIndex }` or just simply as the first argument

src/components/List/List.js
```diff
@@ -1,21 +1,94 @@
...
+        if (e.key === ENTER_KEY) {
+            if (
+                children[selectedIndex] &&
+                children[selectedIndex].props &&
+                children[selectedIndex].props.onClick
+            ) {
+                children[selectedIndex].props.onClick(e);
+
+                if (onSelect) {
+                    onSelect();
```
eszthoff: This function should be called with some argument (e.g. selectedIndex) so that the consuming app can find out what item has been selected.

src/components/List/List.js
```diff
@@ -1,21 +1,94 @@
...
-        <ul {...rest} ref={ref} {...block(props)}>
-            {React.Children.map(children, child =>
-                child ? React.cloneElement(child, elem('item', props)) : null
+        <ul {...rest} ref={ref} tabIndex="0" onKeyDown={handleKeyDown} {...block(props)}>
+            {React.Children.map(children, (child, index) =>
+                child
+                    ? React.cloneElement(child, {
+                          ...elem('item', props),
+                          isHighlighted: index === selectedIndex,
+                          onMouseOver: () => handleMouseOver(index)
```
eszthoff: maybe enough to select an item on mouse enter? that would reduce confusion when the mouse is just forgotten over an item

src/components/List/List.js
```diff
@@ -1,21 +1,94 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
```
eszthoff: I think it is clearer why those exceptions are made if they are placed at the line where they are needed. That can also help with linting in the future, so that these exceptions are only added after per-case consideration.

src/components/List/List.js
```diff
@@ -39,12 +112,18 @@ List.propTypes = {
         return error;
     },
     /** Adds dividing lines between the list items */
-    isDivided: PropTypes.bool
+    isDivided: PropTypes.bool,
+    /** onNavigate function callback */
```
eszthoff: maybe add to the comments what arguments will be passed to this callback. As well as to the next one.

src/themes/oneui/base-variables/_colors.scss
igorarkhipenko: @antipin can we mark it as resolved?

antipin: @igorarkhipenko 👍

PR #141, opened 2019-07-18
feat(component): new Modal component

eszthoff:
Story: [ONEUI-67](https://jira.textkernel.nl/browse/ONEUI-670)

### DIFF
--- a/src/components/Modal/Modal.scss
+++ b/src/components/Modal/Modal.scss
@@ -10,11 +10,11 @@
         transition: opacity var(--transition-duration) ease-in-out;
     }
 
-    &__overlayEnter {
+    &__overlay--entered {
         opacity: 1;
     }
 
-    &__overlayExit {
+    &__overlay--exited {
         opacity: 0;
     }
 
@@ -35,11 +35,11 @@
         transition: transform var(--transition-duration);
     }
 
-    &__contentEnter {
+    &__content--entered {
         transform: translateY(0);
     }
 
-    &__contentExit {
+    &__content--exited {
         transform: translateY(50px);
     }
 }
--- a/src/components/Modal/Modal.scss
+++ b/src/components/Modal/Modal.scss
@@ -1,0 +1,45 @@
+.Modal {
+    &__overlay {
+        position: fixed;
+        top: 0;
+        left: 0;
+        right: 0;
+        bottom: 0;
+        background-color: rgba(0, 0, 0, 0.5);
+        opacity: 0;
+        transition: opacity var(--transition-duration) ease-in-out;
+    }
+
+    &__overlayEnter {
+        opacity: 1;
+    }
+
+    &__overlayExit {
+        opacity: 0;
+    }
+
+    &__content {
+        position: relative;
+        top: 15vh;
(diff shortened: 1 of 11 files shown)


src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,45 @@
...
+        transition: opacity var(--transition-duration) ease-in-out;
+    }
+
+    &__overlayEnter {
+        opacity: 1;
+    }
+
+    &__overlayExit {
+        opacity: 0;
+    }
```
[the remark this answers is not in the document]
eszthoff: renamed it

mukiienko:
For me this Modal component doesn't work in IE11 :(

eszthoff:
does work for me. There are other errors in storybook though...

src/utils/OneUI/OneUI.js
```diff
@@ -83,6 +84,14 @@ class OneUI {
             head.appendChild(styleElement);
         });
     }
+
+    /**
+     * Inits ReactModal and sets global params for it.
+     * @param {string} selector - css selector for the app element
+     */
+    static modalInit(selector) {
```
SebastiaanPasterkamp: Would it make sense to make this selector an optional parameter for the `OneUI.init(...)` function? Required if `Modal` is used.  Otherwise you'll have to

```
OneUI.init(initParams);
OneUI.modalInit(appSelector);
OneUI.futureInits(...);
```

antipin: Hm, I probably missed it, but why `modalInit` lives in OneUI namespace, not in the `Modal` component itself?
it's not good for the following reasons:

- if a developer will open `Modal` component, he/she will hardly find out that there is also some implicit logic somewhere else.
- If a project (assuming that it has tree shaking set up) doesn't use `Modal` but using `OneUI.init()` then it will have the whole unnecessary `react-modal` in it's bundle.

Can it be a static method on `Modal` component itself?

eszthoff: This is something you need to run once in your app for optimising modal for screen readers. Hence I put it as part of the OneUI init. However tree shaking is a good point. Could this be solved in having OneUI build modularly (which we should do anyway?)

As for the developer not being aware of it, that's would not change, regardless on where this init function lives. Hence there is a note in Storybook about it.

antipin: > Could this be solved in having OneUI build modularly (which we should do anyway?)

I don't see how it can be solved if we require `react-modal` in `src/utils/OneUI/OneUI.js`

If we have this function in `Modal` as `Modal.init()` then it can look like this:

```js
import React from 'react';
import { render } from 'react-dom';
import OneUI from '@textkernel/oneui';
import App from 'components/App';
import { Module } from '@textkernel/oneui';

const appContainerSelector = 'main';

Module.init(appContainerSelector);

OneUI.init()
  .then((): void => {
    render(<App />, document.getElementsByTagName(appContainerSelector)[0]);
  })
  .catch(...)



```

eszthoff: done

PR #143, opened 2019-07-19
feat: renamed OneUI.startCssVarsPonyfill to OneUI.applyCssVarsPonyfill()

antipin:
Initial idea was to call OneUI.init() once on page load (that used to call
OneUI.startCssVarsPonyfill underthegood). But with code splitting
`cssVarsPonyfill` should be appled after any route navigation.
So it should be exposed with a more meaningfull name.

BREAKING CHANGE: if you used `OneUI.startCssVarsPonyfill()` directly in your app
rename it to `OneUI.applyCssVarsPonyfill()`.

### DIFF
--- a/src/utils/OneUI/__tests__/OneUI.spec.js
+++ b/src/utils/OneUI/__tests__/OneUI.spec.js
@@ -21,7 +21,7 @@ describe('OneUI loader that starts the ponyfill and attach the theme to DOM', ()
             configurable: true
         });
 
-        OneUI.startCssVarsPonyfill();
+        OneUI.applyCssVarsPonyfill();
 
         expect(cssVarsPonyfill).toBeCalledTimes(1);
     });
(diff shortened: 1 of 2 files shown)


src/utils/OneUI/__tests__/OneUI.spec.js
```diff
@@ -45,18 +45,18 @@ describe('OneUI loader that starts the ponyfill and attach the theme to DOM', ()
         const ponyfillOptions = { onlyLegacy: false };
 
         const mockedApplyTheme = jest.spyOn(OneUI, 'applyTheme');
-        const mockedStartCssVarsPonyfill = jest.spyOn(OneUI, 'startCssVarsPonyfill');
+        const mockedAtartCssVarsPonyfill = jest.spyOn(OneUI, 'applyCssVarsPonyfill');
```
eszthoff: cool mock name, but... `mockApplyCssVarsPonyfill` perhaps?

antipin: good catch :)

PR #145, opened 2019-07-24
fix(List): Fix scroll to highlighted component with keyboard navigation 

igorarkhipenko:


src/components/List/List.js
```diff
@@ -48,6 +56,11 @@ const List = React.forwardRef((props, ref) => {
                 e.preventDefault();
                 setSelectedIndex(nextSelectedIndex);
 
+                // Scroll view with a delay so this function would refer to the next highlighted item
+                setTimeout(() => {
+                    highlightedListItem.current.scrollIntoView(SCROLL_INTO_VIEW_SETTINGS);
```
antipin: 1. IE11 doesn't support scrollIntoView :(

2. What if component was unmounted before this timeout is executed?

igorarkhipenko: @antipin
1. Yeah, that's a pity, it doesn't get `SCROLL_INTO_VIEW_SETTINGS ` param but the scroll by itself is still working so we decided to stick with that.
2. I'll add `clearTimeout` function call before unmounting to be sure it will never happen

src/components/List/List.js
```diff
@@ -17,9 +17,33 @@ const NAVIGATION_STEP_VALUES = {
...
+            highlightedListItem.current &&
+            highlightedListItem.current.scrollIntoView;
+
+        if (highlightedWithKeyboard && hasScrollIntoViewFunction) {
+            highlightedListItem.current.scrollIntoView(SCROLL_INTO_VIEW_SETTINGS);
+            setHighlightedWithKeyboard(false);
+        }
+    }, [selectedIndex]);
 
     const getNextSelectedIndex = keyCode => {
```
antipin: Looks good. But I've just noticed that all the event handlers in this component are being recreated on every render, which is not efficient. Could you wrap them into `useCallback` in order to memoize it?

PR #148, opened 2019-07-26
feat(component): new Map component

eszthoff:
Story: [ONEUI-68](https://jira.textkernel.nl/browse/ONEUI-68)


src/components/Map/MapRenderer/MapRenderer.js
```diff
@@ -0,0 +1,132 @@
...
+    radius,
+    zIndex: 1
+});
+
+const MapRenderer = React.forwardRef((props, ref) => {
+    const { center, zoom, markers, mapContainerStyle, ...rest } = props;
+    const positions = markers.map(marker => ({ lat: marker.lat, lng: marker.lng }));
+    const mapRef = ref || React.createRef();
+
+    const fitBounds = () => {
```
antipin: Needs to be memoized with `useCallback`

src/components/Map/GoogleLoader/GoogleLoader.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useLoadScript } from '@react-google-maps/api';
+import { LoadingSpinner } from '../../../index';
+
+const GoogleLoader = props => {
```
antipin: Name is bit weird :) it doesn't load Google, it probably loads `google maps sdk`?

eszthoff: GoogleMapsAPILoader?

src/components/Map/MapRenderer/MapRenderer.js
eszthoff: ok. there are a few things to polish still...

src/components/Map/MapRenderer/MapRenderer.js
```diff
@@ -0,0 +1,131 @@
...
+    draggable: false,
+    editable: false,
+    visible: true,
+    radius,
+    zIndex: 1,
+});
+
+const MapRenderer = React.forwardRef((props, ref) => {
+    const { center, zoom, markers, mapContainerStyle, ...rest } = props;
+    const positions = markers.map(marker => ({ lat: marker.lat, lng: marker.lng }));
```
antipin: Below in the code there are few places where we iterate `markers` array, but in a loop's body we're getting position of a marker by referencing `positions[i]`. That looks a bit "fragile", cause it assumes that you have to have 2 arrays of the same length where elements relates to each other by index :)

It could be done in a less complicated way just by enriching `markers` array with `position` property in a following way:

```js
const positionedMarkers = markers.map(marker => ({
  ...marker,
  position: {
    lat: marker.lat,
    lng: marker.lng
  }
}));
```

Furthermore, this markers "processing" can be done higher in the `Map` component once. Or, we can even change `markers` prop interface from
```
markers: PropTypes.arrayOf(
    PropTypes.shape({
        lng: PropTypes.number.isRequired,
        lat: PropTypes.number.isRequired,
        radius: PropTypes.number,
    })
),
```
to 

```
markers: PropTypes.arrayOf(
    PropTypes.shape({
        position: PropTypes.shape.isRequired({
            lng: PropTypes.number.isRequired,
            lat: PropTypes.number.isRequired,
        }),
        radius: PropTypes.number,
    })
),
```
and then we don't need to do this processing at all.

src/components/Map/MapRenderer/MapRenderer.js
antipin: still missing `useCallback `

stories/Map.js
```diff
@@ -0,0 +1,59 @@
...
+
+            return (
+                <div
+                    style={{
+                        width: text('Container width', '800px'),
+                        height: text('Container height', '400px'),
+                    }}
+                >
+                    <Map
+                        apiKey="Please provide a key"
```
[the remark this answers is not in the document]
eszthoff: You need a valid key for it to work. What behaviour would you expect?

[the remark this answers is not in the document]
antipin: Commit a valid key is not an option. My point is that it works without a key (empty string key) in dev mode. With empty key I can see and interact with the map (though it has "DEVELOPMENT MODE" watermark and a popup that needs to be closed). When key has `Please provide a key` value, map doesn't work completely.

src/components/Map/MapRenderer/MapRenderer.js
```diff
@@ -0,0 +1,134 @@
...
+                }
+                map.fitBounds(bounds);
+            });
+        } else {
+            map.setCenter(center);
+            map.setZoom(zoom);
+        }
+    };
+
+    useEffect(() => fitBounds());
```
antipin: Can be `useEffect(fitBounds)`

PR #149, opened 2019-07-26
feat(LocationCard): new LocationCard component

igorarkhipenko:


src/components/LocationCard/LocationCard.js
```diff
@@ -0,0 +1,80 @@
...
+        onRadiusChange,
+        onDelete,
+        ...rest
+    } = props;
+
+    return (
+        <div {...rest} {...block(props)}>
+            <div className="header">
+                <Heading className="title">{locationTitle}</Heading>
+                <Heading className="delete-button" onClick={onDelete}>
```
eszthoff: can we make it a button, so that the html is semantic?

src/constants/index.js
```diff
@@ -18,6 +18,8 @@ export const LIST_NAVIGATION_DIRECTIONS = {
     DOWN: 'ArrowDown',
 };
 
+export const CROSS_CHAR = '×';
```
eszthoff: maybe we should have it in a separate file with commonly used chars. Also, it makes sense to replace current occurrences with this constant (e.g. in clear button in FieldWrapper)

mukiienko:
Screenshot 2019-07-30 at 17 32 26
A little bit broken HTML markup in IE11 and EDGE

mukiienko:
Screenshot 2019-07-30 at 17 36 25
In IE11 delete button is copyable, I think better to use svg icon for delete button

mukiienko:
In the documentation, I noticed prop **locationToString**, have we abandoned this prop?

igorarkhipenko:
> In the documentation, I noticed prop **locationToString**, have we abandoned this prop?

seems like there's no need in it for now since `locationTitle` is the only location property which is passed to this component

igorarkhipenko:
> Screenshot 2019-07-30 at 17 36 25
> In IE11 delete button is copyable, I think better to use svg icon for delete button

I'll check with Carlo if we should replace it with separate button component which has an svg inside, but for now it's fine I guess

src/components/List/List.js
```diff
@@ -61,45 +61,39 @@ const List = React.forwardRef((props, ref) => {
...
-    const handleKeyDown = useCallback(
-        e => {
-            // Update selectedIndex with arrow navigation and make onNavigate function callback
-            if (
-                e.key === LIST_NAVIGATION_DIRECTIONS.UP ||
-                e.key === LIST_NAVIGATION_DIRECTIONS.DOWN
-            ) {
-                const nextSelectedIndex = getNextSelectedIndex(e.key);
+    const handleKeyDown = e => {
+        // Update selectedIndex with arrow navigation and make onNavigate function callback
```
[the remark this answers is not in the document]
igorarkhipenko: @mukiienko accoring to [this article](https://kentcdodds.com/blog/usememo-and-usecallback) it makes things even worse here. `useCallback` wrapping is not a must-have for each callback in function components

src/components/LocationCard/LocationCard.js
```diff
@@ -0,0 +1,99 @@
...
+        sliderLabel,
+        minRadius,
+        maxRadius,
+        radiusStep,
+        onRadiusChange,
+        onDelete,
+        ...rest
+    } = props;
+
+    const getTruncatedTitle = () => {
```
mukiienko: Why do you use for truncate function?
The simplest way to truncate by CSS 
```
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

PR #150, opened 2019-07-26
feat(List): Update item's onClick calling

igorarkhipenko:
BREAKING CHANGE: `onNavigate` and `onSelect` callbacks are now deprecated in favour of `doSelectOnNavigate` prop


src/components/List/List.js
```diff
@@ -149,19 +141,15 @@ List.propTypes = {
     },
     /** Adds dividing lines between the list items */
     isDivided: PropTypes.bool,
-    /** onNavigate function callback. (selectedIndex: number, key: 'ArrowUp' || 'ArrowDown') */
-    onNavigate: PropTypes.func,
-    /** onSelect function callback. (selectedIndex: number) */
-    onSelect: PropTypes.func,
+    doSelectOnNavigate: PropTypes.bool,
```
eszthoff: this needs a description

PR #151, opened 2019-07-31
fix(Autosuggest): styles and behaviour adjustments

eszthoff:
Story: [JF-1708](https://jira.textkernel.nl/browse/JF-1708)

### DIFF
--- a/src/components/Autosuggest/Autosuggest.js
+++ b/src/components/Autosuggest/Autosuggest.js
@@ -126,11 +126,15 @@ class Autosuggest extends React.Component {
     };
 
     handleWrapperClick = openMenu => () => {
-        this.focus(openMenu);
+        const { focused } = this.state;
+        if (!focused) {
+            this.focus(openMenu);
+        }
     };
 
     handleWrapperKeyDown = openMenu => e => {
-        if (e.key === ENTER_KEY) {
+        const { focused } = this.state;
+        if (!focused && e.key === ENTER_KEY) {
             this.focus(openMenu);
         }
     };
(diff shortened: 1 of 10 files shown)


mukiienko:
I am not sure this relates to this task, but for me strangely when we delete an item and will have this item in suggestions. In some cases, it can be just one item in suggestions and clear input.
Screenshot 2019-07-31 at 14 48 58

mukiienko:
Screenshot 2019-07-31 at 14 40 51
Placeholder is not aligned in IE11 :)

mukiienko:
I have an empty input (without selected items) after selected several items and then click outside the browser. But probably it is an issue with the _wrapper_ component

eszthoff:
> I am not sure this relates to this task, but for me strangely when we delete an item and will have this item in suggestions. In some cases, it can be just one item in suggestions and clear input.

I think that is sensible. it is available again for selection. but that's possibly a business logic question and is part of the example implementation.

eszthoff:
> I have an empty input (without selected items) after selected several items and then click outside the browser. But probably it is an issue with the _wrapper_ component

hhmm... couldn't reproduce. but that might be also a problem with the example implementation. maybe it is a mistake after all to add it to the story, as it distracts the reviewer from the actual component (wondering face)

eszthoff:
> Placeholder is not aligned in IE11 :)

I don't mind that :-) It is because of a known bug with flexbox with min-height, but no explicit height (I forgot about it when implementing). If you ok to live with it, I rather do, that go through all the alignment issues with, close/open/many tags etc states...

@carlobernardini ?

mukiienko:
> > Placeholder is not aligned in IE11 :)
> 
> I don't mind that :-) It is because of a known bug with flexbox with min-height, but no explicit height (I forgot about it when implementing). If you ok to live with it, I rather do, that go through all the alignment issues with, close/open/many tags etc states...
> 
> @carlobernardini, @mukiienko ?

Seems we have a lot of users that use IE11, and I think it will be a more used and more noticeable component.

eszthoff:
> > > Placeholder is not aligned in IE11 :)
> > 
> > 
> > I don't mind that :-) It is because of a known bug with flexbox with min-height, but no explicit height (I forgot about it when implementing). If you ok to live with it, I rather do, that go through all the alignment issues with, close/open/many tags etc states...
> > @carlobernardini, @mukiienko ?
> 
> Seems we have a lot of users that use IE11, and I think it will be a more used and more noticeable component.

I changed it to be aligned ok when the component is closed, hover the IE bug is still in effect when the component is open. @carlobernardini is that ok?

carlobernardini:
@eszthoff I don't get alerts for mentions and I don't have IE11 at hand right now. Can you please send me some screenshots of what you mean e.g. on Slack so I can answer your question?

PR #152, opened 2019-08-01
feat(storybook): added withStore decorator for Storybook

mukiienko:


stories/Autosuggest.js
```diff
@@ -52,85 +65,79 @@ storiesOf('Organisms|Autosuggest', module)
...
+        return (
+            <Autosuggest
+                selectedSuggestions={
+                    boolean('Add selectedSuggestions', true)
+                        ? store.get('selectedSuggestions')
+                        : undefined
+                }
+                selectedPlaceholder={
+                    boolean('Add selectionPlaceholder', true) ? getSelectedPlaceholder() : undefined
+                }
```
eszthoff: Maybe we can add an icon with this same trick to render or not depending on the knob.

mukiienko: are you mean something like this?
`
iconNode={
                        boolean('Add iconNode', true)
                            ? <IconMatch />
                            : undefined
                    }
`

eszthoff: yes. Then all props are covered

stories/Autosuggest.js
```diff
@@ -51,86 +58,81 @@ storiesOf('Organisms|Autosuggest', module)
...
+            store.set({ inputValue: '' });
         };
 
-        Implementation.displayName = 'Implementation';
+        const onClearAllSelected = () => {
+            console.log('onClearAllSelected was called');
+            store.set({ selectedSuggestions: [] });
+        };
 
-        return <Implementation style={{ width: '650px' }} />;
```
eszthoff: It looks better in storybook if we keep the style. you can add it directly to Autosuggest

mukiienko: Are you mean to add this style just in the last story  **Example implementation** or to all **Autosuggest**?

PR #154, opened 2019-08-05
feat(component): new Location Autocomplete component

eszthoff:
Story: [ONEUI-72](https://jira.textkernel.nl/browse/ONEUI-72)


src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
```diff
@@ -0,0 +1,136 @@
...
+                {
+                    input: debouncedInputValue,
+                    types: placeTypes,
+                    componentRestrictions: { country },
+                },
+                predictions => {
+                    setSuggestionsList(predictions);
+                    setIsLoading(false);
+                }
+            );
```
mukiienko: We can't 100% rely on external API, so better to wrap it in **try {} catch(err){}**

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
```diff
@@ -0,0 +1,136 @@
...
+            setIsLoading(true);
+            setInputValue(value);
+        } else {
+            setIsLoading(false);
+            resetSuggestionsList();
+        }
+    };
+
+    // eslint-disable-next-line react/display-name
+    const renderListPoweredByGoogle = ({ listInputValue, getItemProps, highlightedIndex }) => (
```
mukiienko: A little bit strange name and I'd prefer to split up this method on one render list other insert img

src/components/LocationAutocomplete/LocationAutocompleteRenderer/__tests__/LocationAutocompleteRenderer.spec.js
```diff
@@ -0,0 +1,13 @@
...
+// import toJson from 'enzyme-to-json';
+import stabGoogleApi from '../../../../__mocks__/googleApiMock';
+// import LocationAutocompleteRenderer from '../LocationAutocompleteRenderer';
+
+stabGoogleApi();
+
+describe('<MapRendered/> that renders a Map with markers', () => {
+    it('should render with default props', () => {
+        // const wrapper = mount(<LocationAutocompleteRenderer />);
+        expect(1).toBe(1);
```
mukiienko: :)

stories/LocationAutocomplete.js
```diff
@@ -0,0 +1,41 @@
...
+        'LocationAutocomplete',
+        () => {
+            const onSelectionChange = value => {
+                console.log('onSelectionChange was called with:');
+                console.log(value);
+            };
+            return (
+                <LocationAutocomplete
+                    apiKey=""
+                    inputPlaceholder={text('Input field placeholder', 'Enter a city or region...')}
```
mukiienko: I see to the ways how we can dynamically insert  _apiKey_ .
1) Use **window.prompt()**. This method block JS flow so google doesn't init without the key.
2) Add into **additionalGoogleProps** id with apiKey, when id changes cause reinit google module

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
```diff
@@ -0,0 +1,136 @@
...
+    placeTypes: PropTypes.arrayOf(PropTypes.string),
+};
+
+LocationAutocomplete.defaultProps = {
+    onSelectionChange: () => null,
+    country: null,
+    placeTypes: ['(regions)'],
+};
+
+export default LocationAutocomplete;
```
mukiienko: I don't see the reason to use separate component **LocationAutocompleteRenderer**. We can implement this logic in **LocationAutocomplete**. Now LocationAutocompleteRenderer component is very coupled with LocationAutocomplete, and we can't reuse it. Or to implement all google logic (init/retrieve suggestions, etc) in one component LocationAutocomplete and then LocationAutocompleteRenderer will not aware about google, just takes care about render suggestions, debounce, clear, etc.

package.json
```diff
@@ -46,12 +46,12 @@
     "homepage": "https://github.com/textkernel/oneui#readme",
```
mukiienko: Screenshot 2019-08-06 at 11 17 59
Prop errors

mukiienko: Screenshot 2019-08-06 at 11 45 35
Missing pin icon on the left of the input

mukiienko: Screenshot 2019-08-06 at 10 26 44
Make sense implement google module that will be responsible for google features(init/get suggestions, etc), because now we can init this component just once, the second init throw the error _mulitiply init_. This error also appears when to want to use _Map_ component with _Location Autocomplete_ component. Google module probably should be a singleton, that allows us to init google just one time.

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -0,0 +1,52 @@
...
+
+    return (
+        <LoadScriptNext
+            googleMapsApiKey={apiKey}
+            language={language}
+            region={region}
+            loadingElement={<LoadingSpinner centerIn="parent" />}
+            libraries={LOAD_SCRIPT_LIBRARIES}
+            {...additionalGoogleProps}
+        >
```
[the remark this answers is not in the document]
eszthoff: You can pass that through the `additionalGoogleProps`.

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
eszthoff: There are 2 reasons for this split: 
1. to make it testable
2. the wrapper that loads google can be easily extracted. So in LocationSelector, we can have one loader that loads all APIs at once

we cannot extract google completely as it's the main point of this component - managing info from google. The rest of the logic is already extracted to Autosuggest

package.json
eszthoff: We also thought so in the beginning, but basically that would mean to reimplement LoadScriptNext, so it would be a waste of time. I haven't looked at multiple occurrences on one page, but will check

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -0,0 +1,52 @@
...
+import { LoadingSpinner } from '../../index';
+import LocationAutocompleteRenderer from './LocationAutocompleteRenderer';
+
+const LOAD_SCRIPT_LIBRARIES = ['places'];
+
+const LocationAutocomplete = props => {
+    const { apiKey, language, region, additionalGoogleProps, ...rest } = props;
+
+    return (
+        <LoadScriptNext
```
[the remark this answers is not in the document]
eszthoff: We would just reimplement this LoadScriptNext. I don't really see the point. What return values would you need?

scripts/test/jest.config.js
```diff
@@ -35,4 +35,9 @@ module.exports = {
     // An array of regexp pattern strings that are matched against all file paths before executing
     // the test. If the file path matches any of the patterns, coverage information will be skipped.
     coveragePathIgnorePatterns: ['/dummy-components/', '/node_modules/', '/packages/bem/'],
+
+    // A map from regular expressions to module names that allow to stub out resources with a single module
+    moduleNameMapper: {
+        '.+\\.(svg|png|jpg)$': '<rootDir>/src/__mocks__/testFileMock.js',
```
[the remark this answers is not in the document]
eszthoff: ok. I'll leave it for now then. If there are other files need mocking, we can use this library later

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
```diff
@@ -0,0 +1,136 @@
...
+import styles from '../LocationAutocomplete.scss';
+
+const { elem } = bem({
+    name: 'LocationAutocomplete',
+    classnames: styles,
+});
+
+const DEBOUNCE_DELAY = 350;
+
+const LocationAutocomplete = props => {
```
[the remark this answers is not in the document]
eszthoff: done

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { LoadScriptNext } from '@react-google-maps/api';
+import { LoadingSpinner } from '../../index';
+import LocationAutocompleteRenderer from './LocationAutocompleteRenderer';
+
+const LOAD_SCRIPT_LIBRARIES = ['places'];
```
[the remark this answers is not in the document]
eszthoff: renamed it to GOOGLE_API_LIBRARIES

package.json
eszthoff: added

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -0,0 +1,52 @@
...
+
+const LocationAutocomplete = props => {
+    const { apiKey, language, region, additionalGoogleProps, ...rest } = props;
+
+    return (
+        <LoadScriptNext
+            googleMapsApiKey={apiKey}
```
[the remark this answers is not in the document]
eszthoff: not sure what do you mean. it is a required prop, that should be enough, no?

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
eszthoff: added some provisional error handling

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
eszthoff: the list is rendered by autosuggest, this method overwrites the default behaviour of autosuggest. I think this is the simplest way to get everything into the proper place in the dom structure.

src/components/LocationAutocomplete/LocationAutocomplete.js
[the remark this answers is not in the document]
mukiienko: We can use hooks without returned values, like useEffect. For me, it is more consistent, if we use hooks so why we didn't use it here. Other points, when I use a component in most cases I expect it will return some UI, now it is just the extra wrapper. The advantage of hooks is also that we get rid of nested hell :) when one component inside another component and so on.. But I don't sink that will blocker, it is just a proposal.

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
mukiienko: if Google doesn't expose on some reasons window.google.... object, or it will be inited with a delay so in this case, UI can go down and we got the white screen.

package.json
```diff
@@ -46,15 +46,18 @@
...
         "downshift": "3.2.10",
         "fast-memoize": "2.5.1",
+        "prop-types": "^15.7.2",
+        "rc-slider": "8.6.13",
         "react-modal": "3.9.1"
     },
     "peerDependencies": {
+        "@fortawesome/fontawesome-svg-core": "^1.2.21",
+        "@fortawesome/free-solid-svg-icons": "^5.10.1",
+        "@fortawesome/react-fontawesome": "^0.1.4",
```
mukiienko: Alex recommended me to use in NextGent UI `react-icons`. For it doesn't need to add 3 dependencies and it easy to use.

package.json
```diff
@@ -46,15 +46,18 @@
...
-        "rc-slider": "8.6.13",
         "downshift": "3.2.10",
         "fast-memoize": "2.5.1",
+        "prop-types": "^15.7.2",
+        "rc-slider": "8.6.13",
         "react-modal": "3.9.1"
     },
     "peerDependencies": {
+        "@fortawesome/fontawesome-svg-core": "^1.2.21",
+        "@fortawesome/free-solid-svg-icons": "^5.10.1",
```
mukiienko: why did you put `fortawesome` into peerDependencies

src/components/LocationAutocomplete/LocationAutocomplete.js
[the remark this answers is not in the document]
eszthoff: you can always mess things up if you want. you can also give a key that doesn't work. As for empty key, it actually does work for map for instance, but it renders it in dev mode, with an overlay and warning popup (see map in storybook)

package.json
eszthoff: ok, I'll have a look

package.json
eszthoff: because we need to use it in NGUI too, so better not to load it twice

src/components/LocationAutocomplete/LocationAutocompleteRenderer/LocationAutocompleteRenderer.js
```diff
@@ -0,0 +1,174 @@
...
+                            onError(status);
+                        }
+                    }
+                    setIsLoading(false);
+                }
+            );
+        } else {
+            resetSuggestionsList();
+        }
+    }, [country, debouncedInputValue, onError, placeTypes]);
```
mukiienko: Tell me please why we need `onError` in this array?

eszthoff: because it is used within the effect function

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -0,0 +1,176 @@
...
+
+            service.getPlacePredictions(
+                {
+                    input: debouncedInputValue,
+                    types: placeTypes,
+                    componentRestrictions: { country },
+                },
+                (predictions, status) => {
+                    if (ACCEPTABLE_API_STATUSES.includes(status)) {
+                        setSuggestionsList(predictions);
```
antipin: **Consider slow connection scenario:**

1. Connection is not super fast
2. User types `Ams`, it triggers request `A`
3. While request `A` is still in progress, user deletes `Ams` and types `Lon`, that triggers request `B`.
4. At this point request `A` is completed and user sees suggestions for `Ams`,  having `Lon` in the input. That's inconsistent state of UI (in this scenario it's temporary).
5. Later request `B` is completed and only now user sees suggestions for `Lon`

**Worse scenario:**

1. Connection is not very stable
2. User types `Ams`, it triggers request `A`
3. While request `A` is still in progress, user deletes `Ams` and types `Lon`, that triggers request `B`.
4. At this point request `B` is completed and user sees suggestions for `Lon` for a moment
5. Moment later request `A` is completed and user sees suggestions for `Ams` having `Lon` in the input. That's also inconsistent state of UI (and in this scenario it's already a bug).

scripts/build/utils/utils.js
```diff
@@ -38,7 +38,19 @@ function getRuleCSS({ styleLoader, localIdentName, includePaths, context }) {
     };
 }
 
+function getRuleFiles({ fileLoader }) {
+    return {
+        test: /\.(png|jpe?g|gif)$/i,
```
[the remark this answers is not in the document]
eszthoff: Adding svg is a good point. But having it already defined for other img types had no negative effect, as far as I know. But it will make things smoother if we ever happen to add something like that. Or am I missing something?

package.json
```diff
@@ -46,16 +46,17 @@
...
         "@babel/polyfill": "7.0.0",
-        "@react-google-maps/api": "1.4.2",
+        "@react-google-maps/api": "1.5.1",
         "css-vars-ponyfill": "1.16.4",
-        "prop-types": "^15.7.2",
-        "rc-slider": "8.6.13",
         "downshift": "3.2.10",
         "fast-memoize": "2.5.1",
+        "prop-types": "^15.7.2",
+        "rc-slider": "8.6.13",
```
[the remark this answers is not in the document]
eszthoff: might had been Alexey :-P
Ihor uses `npm i` lol

src/components/Autosuggest/Autosuggest.scss
```diff
@@ -66,9 +66,10 @@
     }
 
     &__list--focused {
-        max-height: 200px;
+        max-height: 215px;
```
antipin: This hard-coded `max-height` number breaks the idea of `listRenderer` property

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -0,0 +1,191 @@
...
+                        ? suggestionToString(item)
+                        : item.structured_formatting.main_text}
+                </MarkedText>
+            </ListItem>
+        ));
+
+        if (!hidePoweredByGoogleLogo) {
+            elems.unshift(
+                // TODO: replace this element with correct resolution version
+                // for all available images see: https://developers.google.com/maps/documentation/images/powered_by_google.zip
```
antipin: will this be addressed? I don't see any blockers for this

eszthoff: It isn't 100% straight forward, and since we want to finish this PR and move ahead, and not complicate it even more then what it is now, we said we gonna do it in a separate task.

src/components/Autosuggest/Autosuggest.scss
eszthoff: I'm not really sure how it relates to `listRenderer`, that one is responsible for each item in the list, while this one is for the max-height of the dropdown. The dropdown is scrollable, and hence the items will show up regardless of their height.

eszthoff: but we can move it to a variable if that is what you meant

src/components/LocationAutocomplete/LocationAutocomplete.js
antipin: What's the exact difficulty with it? It's just about extracting this piece of code into component and add  a media query for hd display that reference @2x and @3x images from the archive provided.

And if we skip it now, chances high that we'll never have time to come back to it.

src/components/Autosuggest/Autosuggest.scss
antipin: I mean, that `215px` is not just some random value, it plays perfectly with the current height of the item, which is `43px` (`43 * 5 = 215`).
And `listRenderer` allows to render any external component as an item. So it would be logical that max-height is calculated based on Item's height.

src/components/LocationAutocomplete/LocationAutocomplete.js
eszthoff: I can create a ticket now, if it isn't there yet. but there are a lot more then 3 images in that zip folder, and what media query is relevant for each is not that clear

antipin: I think it's more than enough to add just this one so that it doesn't look blurry on retina.
image

src/components/Autosuggest/Autosuggest.scss
eszthoff: perhaps when we have a usecase for it, we can add a function for figuring it out.

src/components/LocationAutocomplete/LocationAutocomplete.js
eszthoff: I added retina

src/components/Autosuggest/Autosuggest.scss
antipin: then let's also remove `listRenderer` property, cause now it's hafly done then.

eszthoff: we use it to render the poweredByGoogle img in an absolute position within the dropdown. It is fully done. 
BTW: there was no requirement to adjust the height to show 5 items. If the items are very short (in height) that would not look good...

PR #156, opened 2019-08-09
feat:(Map): rename Map components to match LocationAutocomplete structure

eszthoff:
Story: [ONEUI-83](https://jira.textkernel.nl/browse/ONEUI-83)

BREAKING CHANGE: Map component is renamed to MapWithGoogleLoader and MapRenderer to Map to better reflect their function and to match the LocationAutocomplete API


carlobernardini:
@eszthoff please note that breaking change is spelled without underscore, otherwise it probably won't be picked up by conventional changelog generator.

PR #157, opened 2019-08-12
Add SelectorComponent API (WIP)

igorarkhipenko:


src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+const DEFAULT_LOCATION_CIRCLE_RADIUS = 20;
+
+function LocationSelector(props) {
+    /**
+     * Stores an array of selected location objects
+     *
+     * locationShape = {
+     *  id: PropTypes.string.isRequired,
+     *  title: PropTypes.string.isRequired,
+     *  coordinates: PropTypes.shapeOf({ lng: number, lan: number }).isRequired,
```
eszthoff: maybe call it center to be consistent with other components

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+
+const DEFAULT_LOCATION_CIRCLE_RADIUS = 20;
+
+function LocationSelector(props) {
+    /**
+     * Stores an array of selected location objects
+     *
+     * locationShape = {
+     *  id: PropTypes.string.isRequired,
+     *  title: PropTypes.string.isRequired,
```
eszthoff: description...

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+     *  coordinates: PropTypes.shapeOf({ lng: number, lan: number }).isRequired,
+     *  radius: PropTypes.number.isRequired,
```
eszthoff: also add `structured_formatting`, `radiusUnits`

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+    return <div />;
+}
+
+LocationSelector.displayName = 'LocationSelector';
+
+LocationSelector.propTypes = {
+    /** country where search can take place */
+    country: PropTypes.string.isRequired,
+    /** locale code in which suggestions should be displayed */
+    locale: PropTypes.string.isRequired,
```
eszthoff: lets call it language

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+    /** placeholder for LocationAutocomplete field */
+    autocompletePlaceholder: PropTypes.string.isRequired,
+    /** label for the Done button */
+    doneLabel: PropTypes.string.isRequired,
+    /** label to be used on the clear all button */
+    removeAllLabel: PropTypes.string.isRequired,
+    /** callback function for closed modal */
+    onBlur: PropTypes.func,
+
```
eszthoff: missing props:
radiusUnits: oneOf['km', 'mi']; 
radiusUnitDisplayText: string (e.g. km, miles or other languages), 
minRadius, 
maxRadius, 
radiusStep, 
center (for the map if nothing yet selected), 
zoom, 
apiKey, 
region (for localising map),  
noSuggestionsPlaceholder, 
country (for place prediction restriction), 
placeTypes (for place prediction),  
showCountryInSuggestions (for autocomplete, but perhaps should be also applied to location description on the card)

please check again that these are all the props required by the children components.

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const DEFAULT_LOCATION_CIRCLE_RADIUS = 20;
```
eszthoff: should be 1, but maybe better to move it to a prop, with default value.

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /** Delete a selected location by id */
+    function removeLocation(locationId) {
+        setSelectedLocations(selectedLocations.filter(location => location.id !== locationId));
+    }
+
+    /** callback function for the Clear button click */
+    function resetSelectedLocations() {
+        setSelectedLocations([]);
```
eszthoff: this should also call `onBlur([])`

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+     * Stores an array of selected location objects
+     *
+     * locationShape = {
+     *  id: PropTypes.string.isRequired,
+     *  title: PropTypes.string.isRequired,
+     *  coordinates: PropTypes.shapeOf({ lng: number, lan: number }).isRequired,
+     *  radius: PropTypes.number.isRequired,
+     * };
+     */
+    const [selectedLocations, setSelectedLocations] = React.useState([]);
```
eszthoff: maybe it is too early to note, but will have to have a state to manage the modal's state: isOpen

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,80 @@
...
+    function removeLocation(locationId) {
+        setSelectedLocations(selectedLocations.filter(location => location.id !== locationId));
+    }
+
+    /** callback function for the Clear button click */
+    function resetSelectedLocations() {
+        setSelectedLocations([]);
+    }
+
+    return <div />;
```
eszthoff: maybe we can already outline it: 
```
<FieldWrapper />
{isOpen && 
  <Modal>
    <LocationAutocomplete />
    <Button context="brand">{doneLabel}</Button>
    {selectedLocations.map(location => <LocationCard />}
    <Map />
  </Modal>
}
```

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,140 @@
...
+    /** string to be displayed in FieldWrapper when the modal is closed, but locations are selected */
+    selectionPlaceholder: PropTypes.string.isRequired,
+    /** placeholder for FieldWrapper when there's no locations selected */
+    mainPlaceholder: PropTypes.string.isRequired,
+    /** label for the Done button */
+    doneLabel: PropTypes.string.isRequired,
+    /** show country name in autocomplete suggestions */
+    showCountryInSuggestions: PropTypes.bool,
+    /** label to be used on the clear all button */
+    removeAllLabel: PropTypes.string.isRequired,
```
eszthoff: maybe we should call it the same as in `FieldWrapper` to be consistent: `clearLabel`. Or rename it there as well

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,140 @@
...
+    /** function with a location object as an argument to be added to the selectedLocations array */
+    addLocation: PropTypes.func.isRequired,
```
eszthoff: normally we name callbacks as `onAddLocation` and inside the component we use `handleXXX`. Same for the next 3 props

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,140 @@
...
+import {
+    Modal,
+    Map,
+    FieldWrapper,
+    LocationAutocomplete,
+    Button,
+    LocationCard,
+} from '@textkernel/oneui';
+
+const DEFAULT_LOCATION_CIRCLE_RADIUS = 1;
```
eszthoff: I still think this should be a prop, but not required, so that it defaults back to 1.

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+}
+
+LocationSelector.displayName = 'LocationSelector';
+
+LocationSelector.propTypes = {
+    /** Google api key */
+    apiKey: PropTypes.string.isRequired,
+    /** stores an array of selected location objects */
+    selectedLocations: PropTypes.arrayOf(
```
eszthoff: will it work ok without this array? Maybe should be required if this to be implemented as a controlled component

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,140 @@
...
+        PropTypes.shape({
+            id: PropTypes.string.isRequired,
+            description: PropTypes.string.isRequired,
+            structuredFormatting: PropTypes.string.isRequired,
+            center: PropTypes.shapeOf({
+                lng: PropTypes.number.isRequired,
+                lan: PropTypes.number.isRequired,
+            }).isRequired,
+            radius: PropTypes.number.isRequired,
```
eszthoff: this is optional in Map. Maybe should be here too. 

Also I wonder in what units this should be? For ease of use maybe the same as `radiusUnits`. Is it ok, that they aren't defined at the same place? Or can it cause confusion later on?

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,140 @@
...
+    /** radius label suffix (e.g. km, miles or other language) */
+    radiusUnitDisplayText: PropTypes.string.isRequired,
+    /** min radius value of the slider component */
+    minRadius: PropTypes.number.isRequired,
+    /** max radius value of the slider component */
+    maxRadius: PropTypes.number.isRequired,
+    /** radius step value of the slider component */
+    radiusStep: PropTypes.number.isRequired,
+    /** map center coordinates if nothing has been selected yet */
+    mapCenter: PropTypes.shape({
```
eszthoff: this could be optional. there is a fallback in Map. Same for the next 2 props

src/components/Autosuggest/__tests__/Autosuggest.spec.js
```diff
@@ -163,7 +163,8 @@ describe('Autosuggest', () => {
             wrapper.find('.Autosuggest__wrapper').simulate('click');
 
             expect(wrapper.state('focused')).toBeTruthy();
-            expect(focusSpy).toHaveBeenCalledTimes(1);
+            // FIXME: focusSpy doesn't get called despite 'focused' value of the wrapper state is true
```
carlobernardini: Something still to be fixed here?

src/components/LocationSelector/LocationSelector.js
carlobernardini: Is this still an open discussion or is it already resolved?

src/components/Map/Map.js
```diff
@@ -105,8 +106,8 @@ Map.propTypes = {
             center: PropTypes.oneOfType([
                 PropTypes.arrayOf(PropTypes.number),
                 PropTypes.shape({
-                    lng: PropTypes.number.isRequired,
-                    lat: PropTypes.number.isRequired,
+                    lng: PropTypes.func.isRequired,
+                    lat: PropTypes.func.isRequired,
```
antipin: Could you please give some more details on why do we need `lng` and `lat` to be functions?

package.json
```diff
@@ -109,6 +109,7 @@
         "react-dom": "16.8.6",
         "react-icons": "3.7.0",
         "react-router-dom": "4.3.1",
+        "regenerator-runtime": "0.13.3",
```
antipin: Seems like `regenerator-runtime` is being used only for testing, might be makes sense to move it to `devDependensies`?

src/components/LocationSelector/LocationSelectorWithGoogleLoader.js
```diff
@@ -0,0 +1,242 @@
...
+    showCountryInSuggestions: true,
+    onBlur: () => null,
+    additionalGoogleProps: {},
+    region: undefined,
+    selectionPlaceholder: null,
+    placeTypes: LocationAutocomplete.defaultProps.placeTypes,
+    onLocationAutocompleteError: LocationAutocomplete.defaultProps.onError,
+};
+
+export default LocationSelectorWithGoogleLoader;
```
antipin: `LocationSelector` and `LocationSelectorWithGoogleLoader` are mixed together now in the same directory.
I think they could be better separated into two different components living in different folders:
`LocationSelector` is kind of a `Molecule`
`LocationSelectorWithGoogleLoader` is an `Organism`.

Also, as Oleksii @mukiienko noticet, both `LocationSelector` and `LocationSelectorWithGoogleLoader` are referencing the same `scss` which shouldn't be the case.

mukiienko:
Screenshot 2019-09-06 at 11 43 56
A lot of errors in the console. Would be nice to keep the console in Storybook clear as will.
When I choose LocationSelector > Basic Component and then click on input locationSelector the Storybook is crashed (white screen).

src/components/Autosuggest/Autosuggest.js
```diff
@@ -21,17 +21,21 @@ const { block, elem } = bem({
...
         this.state = {
             inputValue: '',
             inputValueRecall: '',
-            focused: false,
+            focused: props.isFocused,
         };
 
+        if (props.isFocused) {
+            this.focus();
+        }
```
mukiienko: I suggest to move this code in **componentDidMount** and then is possible to get rid of **setTimeout** in focus method

src/components/Autosuggest/__tests__/Autosuggest.spec.js
mukiienko: If I got right problem with **setTimeout** in component. Above I suggested how possible to fix it

src/components/LocationSelector/LocationSelector.js
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem';
+import { Map, Button, LocationCard, LocationAutocomplete } from '../../index';
+import styles from './LocationSelector.scss';
+
+const { elem } = bem({
+    name: 'LocationSelector',
+    classnames: styles,
+    propsToMods: ['muted'],
```
mukiienko: It is not used in LocationSelector component, so can be deleted.

src/components/LocationSelector/LocationSelectorWithGoogleLoader.js
```diff
@@ -0,0 +1,242 @@
...
+                clearLabel={clearLabel}
+                onClick={handleOpenModal}
+                onClear={onRemoveAllLocations}
+                {...rest}
+            >
+                <FaMapMarkerAlt {...elem('icon', props)} />
+                <Text {...elem('mainTextInput', { ...props, muted: !selectionPlaceholder })}>
+                    {selectionPlaceholder || inputPlaceholder}
+                </Text>
+            </FieldWrapper>
```
mukiienko: Would be nice for more transparency create separate component like 'FaKeLocationInput'  and put this there.

src/components/LocationCard/LocationCard.js
```diff
@@ -74,6 +77,8 @@ LocationCard.propTypes = {
     onRadiusChange: PropTypes.func,
     /** Location card delete callback */
     onDelete: PropTypes.func,
+    /** Name of the wrapping component */
+    As: PropTypes.string,
```
mukiienko: I suggested a little bit change comment for transparency `/** HTML tag to be used to for the wrapping component */`

src/components/Map/Map.js
igorarkhipenko: @antipin because it's the shape of the 'center' object which Google returns, but I think it would be better to call them once to get proper values and then use numbers instead so I'll fix it

src/components/LocationSelector/LocationSelectorWithGoogleLoader.js
igorarkhipenko: @mukiienko I think it's a quite unique case for now and it won't be reused anywhere else. Besides that I find it pretty self-descriptive, short so I'm not sure if we should do it right now

package.json
igorarkhipenko: @antipin it is in devDependencies already

antipin: Ah.. Sorry, I was confused by `react-dom`,  `react-icons` and `react-router-dom`, that looks like non-dev deps.

PR #158, opened 2019-08-15
 feat(Map): add option to fit map to address as default view

eszthoff:
Story: [ONEUI-85](https://jira.textkernel.nl/browse/ONEUI-85)

BREAKING CHANGE: `center` and `zoom` props have been deprecated in favour of `defaultArea` that supports also fitting map boundaries based on address lookup

### DIFF
--- a/src/components/Map/Map.js
+++ b/src/components/Map/Map.js
@@ -17,7 +17,7 @@ const circleOptions = radius => ({
 });
 
 const Map = React.forwardRef((props, ref) => {
-    const { center, zoom, markers, mapContainerStyle, ...rest } = props;
+    const { defaultArea, markers, mapContainerStyle, ...rest } = props;
     const mapRef = ref || React.createRef();
 
     const fitBounds = React.useCallback(() => {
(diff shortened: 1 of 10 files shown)


src/components/Map/Map.js
```diff
@@ -40,20 +40,29 @@ const Map = React.forwardRef((props, ref) => {
                 }
                 map.fitBounds(bounds);
             });
+        } else if (defaultArea.address) {
+            const { Geocoder } = window.google.maps;
+            const geocoder = new Geocoder();
+
+            geocoder.geocode({ address: defaultArea.address }, (result, status) => {
+                if (status === 'OK') {
+                    map.fitBounds(result[0].geometry.viewport);
```
igorarkhipenko: is there any chance of either `result[0]` or `geometry` being falsy values or it's 100% safe to do that without any additional checking?

eszthoff: if status is 'OK' then as far as I understand the docs, it should be ok

PR #159, opened 2019-08-16
fix(bem): add support to preserve existing classes in 'elem'

eszthoff:
Story: [ONEUI-82](https://jira.textkernel.nl/browse/ONEUI-82)


src/packages/bem/bem.js
```diff
@@ -99,20 +99,25 @@ function bemStateless(blockDecl) {
...
-                extraMods,
-                classnamesMap: classnames,
-            });
+            const classes = [
+                buildBemProps({
+                    block: name,
+                    elem: elemName,
+                    props,
+                    propsToMods,
+                    extraMods: {},
```
antipin: Now it doesn't make sense to leave `extraMods` in ` buildBemProps`, should be removed from there as well.

src/packages/bem/bem.js
```diff
@@ -99,20 +99,25 @@ function bemStateless(blockDecl) {
...
+                    classnamesMap: classnames,
+                }).className,
+                classesToKeep,
+            ].filter(c => c);
+
+            return classes.length ? { className: classes.join(' ') } : null;
```
antipin: Now return values of `block` and `elem` functions are inconsistent:

* `block` always return `{ className: '...' }`
* `elem` can return `null` or `{ className: '...' }`

I'd suggest that both functions return an object and if `className` is empty, then it's an empty object (so it can be safely spreaded).

Assuming this, it's more logical to move this logic into `buildBemProps` function from here.

src/packages/bem/README.md
```diff
@@ -239,14 +240,42 @@ Button.scss
 }
 ```
 
-Examples of outcome
--------------------
+### Using elem to enrich existing elements
+
+If you wish to enrich an existing HTML element (e.g. child of the component) with extra classes, you need to make sure to preserve already existing classes on that element. To achieve that you can list existing classes as the 3rd argument for `elem`. For example:
```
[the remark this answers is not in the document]
eszthoff: yes. I will remove HTML

src/packages/bem/__tests__/dummy-components/Unstyled/Unstyled.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import bem from '../../..';
+// import styles from './styles.json';
```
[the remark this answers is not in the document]
eszthoff: deleted

src/packages/bem/bem.js
eszthoff: removed from everywhere

src/packages/bem/bem.js
eszthoff: all this logic is now in benUtils

PR #162, opened 2019-08-30
feat: new TwoPaneView component

eszthoff:


src/components/TwoPaneView/TwoPaneView.js
```diff
@@ -0,0 +1,113 @@
...
+        setDoDisplayRightPane(doDisplay);
+    };
+
+    const handleResize = e => {
+        const widthEl = blockRef.current;
+        const { width: fullWidth } = widthEl.getBoundingClientRect();
+        const padding = +window
+            .getComputedStyle(widthEl, null)
+            .getPropertyValue('padding-right')
+            .match(/\d.?/)[0];
```
igorarkhipenko: Can we have a short comment what exactly is happening here? Or maybe it would be even better to move it to a separate function.

carlobernardini:
Screenshot 2019-08-30 at 14 27 50

Screenshot 2019-08-30 at 14 28 14

It seems to me like the pane(s) aren't properly sizing on initial load.

1. When I initially load the story the right pane is narrower than the one on the left. When I scroll or resize my viewport, the right pane adjusts itself.
2. When I set the left pane height to e.g. 100px, the right pane height is not in sync (again, until I scroll or resize).

Also, I see a whole lot of inline styles applied to both left and right pane, many of which have invalid property values. What are these for? See below:

Screenshot 2019-08-30 at 14 38 28

Last: perhaps for the story it makes sense to add sample result items on the left and some content on the right, to make the example a bit more relatable?

eszthoff:
> Last: perhaps for the story it makes sense to add sample result items on the left and some content on the right, to make the example a bit more relatable?
I think that is a lot of overhead to code, with little benefits. In addition it has the potential to take the focus away from the point of this component, and move it to the content of the panes, instead of how they work.

carlobernardini:
Some more feedback

- Define `minHeight` as a prop so it’s explicitly documented?
- Would it make sense to define `minHeight` on the TwoPaneView component since it kind of affects the layout as a whole?
- The `children` prop on TwoPaneView isn't marked as required but when I remove LeftPane and RightPane in the story source it throws `Cannot read property 'getBoundingClientRect' of null`.

PR #164, opened 2019-09-09
feat(component) new Drawer component

mukiienko:
- a mobile version will be implemented later

### DIFF
--- a/src/components/Drawer/Drawer.js
+++ b/src/components/Drawer/Drawer.js
@@ -1,0 +1,57 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import bem from 'bem';
+import { IoIosArrowUp } from 'react-icons/io';
+import styles from './Drawer.scss';
+
+const { block, elem } = bem({
+    name: 'Drawer',
+    classnames: styles,
+    propsToMods: ['isOpen'],
+});
+
+const Drawer = props => {
+    const { title, initialIsOpen, children, ...rest } = props;
+    const [isOpen, setIsOpen] = useState(initialIsOpen);
+
+    function handleClick() {
+        setIsOpen(!isOpen);
+    }
+
+    return (
+        <div {...block({ isOpen })} {...rest}>
+            <header {...elem('head', props)}>
+                <h4 {...elem('headTitle', props)}>{title}</h4>
+                <div {...elem('headButtons', props)}>
+                    <button
+                        {...elem('extendButton', { isOpen })}
+                        type="button"
+                        onClick={handleClick}
+                    >
+                        <IoIosArrowUp {...elem('extendIcon', props)} />
+                    </button>
+                </div>
+            </header>
+            <section {...elem('body', props)} aria-hidden={!isOpen}>
+                {children}
+            </section>
+        </div>
+    );
+};
+
+Drawer.propTypes = {
+    /** Drawer title */
+    title: PropTypes.string.isRequired,
+    /** Set open/close status for initial state */
+    initialIsOpen: PropTypes.bool,
+    /** Node(s) to be rendered inside the container */
+    children: PropTypes.node.isRequired,
+};
+
+Drawer.defaultProps = {
+    initialIsOpen: false,
+};
+
+Drawer.displayName = 'Drawer';
+
+export default Drawer;
--- a/src/components/Drawer/Drawer.scss
+++ b/src/components/Drawer/Drawer.scss
@@ -1,0 +1,85 @@
+$drawer-head-height: 40px;
+$open-transition-duration: 170ms;
+$icon-size: 26px;
+
+.Drawer {
+    display: flex;
+    flex-direction: column;
+    position: fixed;
+    bottom: 0px;
+    right: 100px;
+    width: var(--drawer-width);
+    max-height: 100vh;
+    box-shadow:  0 5px 10px rgba(0,0,0,.1);
+    z-index: var(--drawer-z-index);
+    transform: translateY(100%) translateY($drawer-head-height * -1);
+    transition-duration: $open-transition-duration;
+    transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+
+    &--isOpen {
+        transform: translateY(0%);
+    }
+
+    &__head {
+        padding: 0 var(--spacing-5x);
+        display: flex;
+        align-items: center;
+        height: $drawer-head-height;
+        background-color: var(--drawer-head-background-color);
+        border-radius: 4px 4px 0 0;
+    }
+
+    &__headTitle {
+        font-weight: 200;
+        color: var(--color-neutral-10);
+        white-space: nowrap;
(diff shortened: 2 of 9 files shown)


src/components/Drawer/Drawer.js
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+        <div {...block({ isOpen })} {...rest}>
+            <header {...elem('head', props)}>
+                <h4 {...elem('head-title', props)}>{title}</h4>
+                <section {...elem('head-buttons', props)}>
+                    <button
+                        {...elem('extend-button', { isOpen })}
+                        type="button"
+                        onClick={() => setIsOpen(!isOpen)}
```
igorarkhipenko: using of the arrow function can be avoided here

src/components/Drawer/Drawer.js
```diff
@@ -0,0 +1,53 @@
...
+});
+
+const Drawer = props => {
+    const { title, initialIsOpen, children, ...rest } = props;
+    const [isOpen, setIsOpen] = useState(initialIsOpen);
+
+    return (
+        <div {...block({ isOpen })} {...rest}>
+            <header {...elem('head', props)}>
+                <h4 {...elem('head-title', props)}>{title}</h4>
```
igorarkhipenko: we're using camelCase style for all of the styling purposes, right?

src/components/Drawer/Drawer.scss
```diff
@@ -0,0 +1,87 @@
...
+        text-overflow: ellipsis;
+    }
+
+    &__head-buttons {
+        flex-shrink: 0;
+        margin-left: auto;
+
+        button {
+            height: 26px;
+            width: 26px;
```
igorarkhipenko: maybe it would be better to use a variable instead?

src/components/Drawer/Drawer.scss
```diff
@@ -0,0 +1,87 @@
...
+    position: fixed;
+    bottom: 0px;
+    right: 100px;
+    width: var(--drawer-width);
+    max-height: 100vh;
+    box-shadow:  0 5px 10px rgba(0,0,0,.1);
+    z-index: 100;
+    transform: translateY(100%) translateY(calc(-1 * var(--drawer-head-height)));
+    transition: var(--transition-duration);
+    transition-timing-function: cubic-bezier(.4,0,.2,1);
```
igorarkhipenko: I think it's better to replace it with `(0.4, 0, 0.2, 1)` for better readability

src/components/Drawer/Drawer.scss
```diff
@@ -0,0 +1,87 @@
...
+
+    &--isOpen {
+        transform: translateY(0%);
+    }
+
+    &__head {
+        padding: 0 15px;
```
igorarkhipenko: `var(--spacing-5x)`

src/components/Drawer/Drawer.js
mukiienko: For me not clear how we can avoid the arrow function. Could you please give an example?

igorarkhipenko:
```
function handleClick() {
  setIsOpen(!isOpen);
}
. . .
onClick={handleClick}
```

mukiienko: what is the profit use this approach and create an extra variable?

src/components/Drawer/Drawer.scss
mukiienko: Good point!

src/components/Drawer/Drawer.js
mukiienko: Probably yes. I was confused while I was implementing Drawer component because I looked at some last components (LocationCard), and some of them also are mixed style. So I thought for _block_  we use camelCase, for _element_ hyphen, for _modificator_  camelCase. I agree with you, we should stick one style
Screenshot 2019-09-10 at 10 19 52

src/components/Drawer/Drawer.js
antipin: In functional components, technically, it doesn't make much sense, cause this handler function will be recreated on every render. But for the sake of code style and better readability I'd still stick to the separate declaration of event handlers. Furthermore, with `useCallback` handler can be created only once.

```
function Foo() {
  const handleClick = useCallback(
    () => console.log('Click happened'),
    [],
  );
  return <Button onClick={handleClick}>Click Me</Button>;
}
```

igorarkhipenko: @mukiienko just to be consistent and transparent all over the code, so you would always expect a handler for an event inside the component. Also it's easier to extend the functionality inside the handler function this way.
It doesn't make any performance profit or whatsoever, but just unifies all of the handlers to follow the same pattern

igorarkhipenko: @antipin I think `useCallback` would be an overkill in a case like this but only makes it worse.
[Here](https://kentcdodds.com/blog/usememo-and-usecallback/) is an article which explains why

src/components/Drawer/Drawer.js
```diff
@@ -0,0 +1,57 @@
...
+
+    function handleClick() {
+        setIsOpen(!isOpen);
+    }
+
+    return (
+        <div {...block({ isOpen })} {...rest}>
+            <header {...elem('head', props)}>
+                <h4 {...elem('headTitle', props)}>{title}</h4>
+                <section {...elem('headButtons', props)}>
```
antipin: `<section>` tag probably is not the best choice here, as far as it assues header and content inside.

image

src/components/Drawer/Drawer.scss
```diff
@@ -0,0 +1,90 @@
...
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+    &__headButtons {
+        flex-shrink: 0;
+        margin-left: auto;
+
+        button {
```
antipin: Using nested selectors is not recommended practice in BEM, especially if there is no strong reason for it.
In this case `&__headButtons button` can be selected just by `&__extendButton`

src/components/Drawer/Drawer.js
```diff
@@ -0,0 +1,57 @@
...
+    classnames: styles,
+    propsToMods: ['isOpen'],
+});
+
+const Drawer = props => {
+    const { title, initialIsOpen, children, ...rest } = props;
+    const [isOpen, setIsOpen] = useState(initialIsOpen);
+
+    function handleClick() {
+        setIsOpen(!isOpen);
```
antipin: In IE11 on browserstack Drawer doesn't collapse or expand. If I click on the `extendButton` – nothing happens :(

src/components/Drawer/Drawer.js
mukiienko: Good point!

src/components/Drawer/Drawer.js
mukiienko: Fixed, thanks!

PR #170, opened 2019-09-18
feat: popup component design with some examples

antipin:


stories/Popup.js
```diff
@@ -0,0 +1,35 @@
...
+        </div>
+    );
+});
+Popover.propTypes = {
+    setPopupVisibility: PropTypes.func.isRequired,
+};
+Popover.displayName = 'Popover';
+
+storiesOf('Atoms|Popup', module).add('Popup', () => (
+    <Popup
```
eszthoff: using render-props pattern it would look like:
``` 
<Popup>
        {({ setPopupVisibility, isOpened, anchorRef, contentRef, contentProps }) => (
            <React.Fragment>
                <Button onClick={() => setPopupVisibility(!isOpened)} ref={anchorRef}>
                    Click me!
                </Button>
                {ReactDom.createPortal(
                    <Popover
                        setPopupVisibility={setPopupVisibility}
                        ref={contentRef}
                        {...contentProps}
                    />,
                    document.body
                )}
            </React.Fragment>
        )}
    </Popup>
```

src/components/Popup/Popup.js
```diff
@@ -0,0 +1,167 @@
...
+        const contentElem = content(this.getArgs());
+        const contentElemWithRef = React.cloneElement(contentElem, {
+            ref: this.contentRef,
+            'data-popup': 'true',
+        });
+        return ReactDOM.createPortal(contentElemWithRef, document.body);
+    }
+
+    render() {
+        return (
```
eszthoff: or using render-props pattern, the `renderAnchor` and `renderContent` functions above could be removed and this render replaced with:
```
render() {
        const { children } = this.props;
        return children({
            ...this.getArgs(),
            anchorRef: this.anchorRef,
            contentRef: this.contentRef,
            contentProps: { 'data-popup': true },
        });
    }
```

stories/Popup.js
carlobernardini: If I may challenge your thoughts: it seems to me like this approach is more flexible but may not be very explicit with lesser separation of concerns (children serves for rendering both anchor and content - that's probably hard to properly document). Looking at this example it also seems to me like it would introduce more overhead / complexity in using the component (wrap in fragment, create portal, pass refs).

What if you for example do keep `anchorRenderer` and use `children` for content instead (both as render props) - would that perhaps be an idea?

eszthoff: I'm not sure either if this is better, but I think it is good to look at this approach too. 
As for passing refs and props, that could be further refined. For example it could be put together into a `getAnchorProps` function that can take params you want to use and combine with defaults from Popup. That would allow for example passing an external ref so that the parent component also has access to it if needed. The Fragment and Portal could also be managed by the Popup if we always assume that the children function returns an array of 2 nodes.
In conclusion I mainly tried to have a look at how it can work with the render-props pattern and see if what API we prefer. I'm actually open for both.

antipin: Thanks for feedback!
I'm trying to understand what problem you're trying to solved by the proposal above, but seems like it only adds more complexity. Or I'm I thinking too simple?

About render props pattern. [React docs](https://reactjs.org/docs/render-props.html) give the following example on this:
```jsx
<DataProvider
  render={data => (
    <h1>Hello {data.target}</h1>
  )}
/>
```

and popup's interface looks exactly the same with the only difference that it has 2 render props instead of one, cause it handles 2 components.

```jsx
<Popup
  placement="down"
  anchor={({ setPopupVisibility, isOpened }) => (
    <AnchorComponent setPopupVisibility={setPopupVisibility} isOpened={isOpened} />
  )}
  content={({ anchorWidth }) => (
      <SomeContentComponent anchorWidth={anchorWidth}
  )}
/>
```

But indeed make sense to rename `anchor` to `anchorRenderer` and `content` to `contentRenderer`

PR #171, opened 2019-09-19
feat(Drawer): improvements Drawer component

mukiienko:


src/components/Drawer/Drawer.js
```diff
@@ -7,19 +7,29 @@ import styles from './Drawer.scss';
...
+    propsToMods: ['isShowAndClose', 'isShowAndOpen', 'isOpen'],
 });
 
 const Drawer = props => {
-    const { title, initialIsOpen, children, ...rest } = props;
-    const [isOpen, setIsOpen] = useState(initialIsOpen);
+    const { title, isShow, isOpen: isOpenProps, initialIsOpen, onClick, children, ...rest } = props;
+    const [isOpen, setIsOpen] = useState(isOpenProps || initialIsOpen);
+    // eslint-disable-next-line no-nested-ternary
+    const status = isShow && isOpen ? 'isShowAndOpen' : isShow ? 'isShowAndClose' : '';
```
igorarkhipenko: I would replace this one-liner with a separate function in order to avoid using of double ternary operator and undeclared string constants
```
function getStatus() {
    const STATUS_LIST = {
        IS_SHOW_AND_OPEN: 'IS_SHOW_AND_OPEN',
        IS_SHOW_AND_CLOSE: 'IS_SHOW_AND_CLOSE',
        EMPTY: '',
    };

    if (isShow) {
        return isOpen ? STATUS_LIST.IS_SHOW_AND_OPEN : STATUS_LIST.IS_SHOW_AND_CLOSE;
    }

    return STATUS_LIST.EMPTY;
}
```

mukiienko: Thanks, nice improvements.

src/components/Drawer/Drawer.js
```diff
@@ -1,55 +1,112 @@
...
 };
 
 Drawer.propTypes = {
     /** Drawer title */
     title: PropTypes.string.isRequired,
     /** Set open/close status for initial state */
-    initialIsOpen: PropTypes.bool,
+    initialIsExpand: PropTypes.bool,
+    /** Control visibility Drawer component */
+    isShow: PropTypes.bool,
```
eszthoff: `isShow` and `isExpand` is bad English, and hence hard to read. It should be `doShow` and `doExpand` or `isShown` and is `Expanded` instead. That applies to several places, not just here (e.g. intialIsExpand should be initiallyDoExpand, etc).

src/components/Drawer/Drawer.js
```diff
@@ -1,55 +1,112 @@
...
+const TRANSITION_DURATION = {
+    enter: 10,
+    exit: 300,
+};
+
 const { block, elem } = bem({
     name: 'Drawer',
     classnames: styles,
-    propsToMods: ['isOpen'],
+    propsToMods: ['isShownAndExpanded', 'isShownAndClosed', 'isExpanded'],
```
eszthoff: seems to me that isShownAndExpanded is always equal isExpanded. Can they be merged? Also maybe the naming can be simplified then

src/components/Drawer/Drawer.js
```diff
@@ -1,55 +1,112 @@
...
 });
 
 const Drawer = props => {
-    const { title, initialIsOpen, children, ...rest } = props;
-    const [isOpen, setIsOpen] = useState(initialIsOpen);
+    const {
+        title,
+        isShow,
```
eszthoff: do we really need this? If the component doesn't need to be shown, why to render it at all? Cannot it be part of the business logic, and simply assume that if the component is there, it should be shown indeed (as it is with any other component)? That would simplify the API as well as the state calculation.

package.json
```diff
@@ -52,7 +52,8 @@
         "fast-memoize": "2.5.1",
         "prop-types": "^15.7.2",
         "rc-slider": "8.6.13",
-        "react-modal": "3.9.1"
+        "react-modal": "3.9.1",
+        "react-transition-group": "4.3.0"
```
eszthoff: if we are planning to use transitions in JF and Search, maybe this should be a peer dependency so that we don't load it twice on the page.

src/components/Drawer/Drawer.js
```diff
@@ -1,55 +1,112 @@
...
+                            <button
+                                {...elem('expandButton', { isExpanded: isExpand })}
+                                type="button"
+                                onClick={handleClick}
+                            >
+                                <IoIosArrowUp {...elem('expandIcon', props)} />
+                            </button>
+                        </div>
+                    </header>
+                    <section {...elem('body', props)} aria-hidden={!(isShow && isExpand)}>
```
eszthoff: shouldn't the body have some default padding around it? It looks a but funny now.

src/components/Drawer/__tests__/Drawer.spec.js
```diff
@@ -2,50 +2,128 @@ import React from 'react';
...
+        wrapper.find('.Drawer__expandButton').simulate('click');
+
+        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(true);
+        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(false);
+        expect(
+            wrapper.find('.Drawer__expandButton').hasClass('Drawer__expandButton--isExpanded')
+        ).toBe(false);
+    });
+
+    it('should expand and close correctly', () => {
```
eszthoff: is this test the same as the previous one?

src/components/Drawer/Drawer.scss
```diff
@@ -56,21 +60,21 @@ $icon-size: 26px;
...
 
-        &--isOpen {
+        &--isExpanded {
             transform: rotate(180deg);
         }
 
         &:hover,
         &:focus {
-            background-color: var(--drawer-extent-hover-background-color);
+            background-color: var(--drawer-expand-hover-background-color);
```
eszthoff: this style is sticky after the drawer is opened until you click somewhere else. Is that intended?

src/components/Drawer/Drawer.js
mukiienko: Yep, I first did this logic in Search, but for the animation we need to keep this in the component. It allows us to make a smooth transition between show/hide states. And I don't fully agree with that is a part of the business logic. It is like a Modal component, it also has the same behavior.

package.json
mukiienko: Make sense, thanks

src/components/Drawer/Drawer.js
mukiienko: I didn't think so. Always inner components have some padding/margin. If we add now some padding it will be very hard to align text inside inner components with Drawer title. But I agree with you, now it looks a little bit funny.

src/components/Drawer/__tests__/Drawer.spec.js
mukiienko: Thanks!

src/components/Drawer/Drawer.js
eszthoff: I missed the part where there is an animation on the initial render. Thanks

src/components/Drawer/Drawer.js
eszthoff: I don't know the usecases this component solves, but if it is normally text (as in the example) that is shown inside the drawer, then adding padding makes sense. If it is usually something more complex, then there is no need to style it indeed.

src/components/Drawer/Drawer.scss
mukiienko: Yes, this behavior on purpose. It shows that element has the focus, and we can use keyboard navigation. It is not a part of the requirement. For me, better to keep this style, maybe make it not so bright. What do you think?

src/components/Drawer/Drawer.js
mukiienko: Yes, it is more complex :). In Search, we added notification and list of the queue. So padding here  will be redundant
Screenshot 2019-09-23 at 15 08 14

eszthoff: I see. Thanks.

src/components/Drawer/Drawer.scss
eszthoff: The color works fine for me. Only if you don't need to engage with the content of the drawer only look at it, there is no hover effect anymore after you opened it first. Maybe there is no need though, and it is fine as it is.

src/components/Drawer/Drawer.js
mukiienko: I changed names for status `IS_EXPANDED`, `IS_CLOSED`, `IS_HIDDEN` and now it looks more clear, thanks. For CSS I would like to keep old names because it reflects real visual status for the Drawer. For example, isExpanded can be not just in shown mode.

src/components/Drawer/Drawer.scss
mukiienko: I think we don't need change focus from other elements when Drawer is opened or showed automatically.

src/components/Drawer/Drawer.js
```diff
@@ -1,55 +1,112 @@
...
+        <Transition appear unmountOnExit in={isShown} timeout={TRANSITION_DURATION}>
+            {state => (
+                <div
+                    {...block({ [getStatus()]: state !== 'entering' })}
+                    aria-hidden={!isShown}
+                    {...rest}
+                >
+                    <header {...elem('head', props)}>
+                        <h4 {...elem('headTitle', props)}>{title}</h4>
+                        <div {...elem('headButtons', props)}>
```
eszthoff: is this div really needed? cannot we put the css all on the button itself?

src/components/Drawer/Drawer.js
```diff
@@ -1,55 +1,112 @@
...
+        document.body
     );
 };
 
 Drawer.propTypes = {
     /** Drawer title */
     title: PropTypes.string.isRequired,
     /** Set open/close status for initial state */
-    initialIsOpen: PropTypes.bool,
+    initialIsExpand: PropTypes.bool,
```
eszthoff: forgot to rename it here `initialIsExpanded`

src/components/Drawer/__tests__/Drawer.spec.js
```diff
@@ -2,50 +2,107 @@ import React from 'react';
...
+
+        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(false);
+        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(false);
     });
 
-    it('should open and close correctly', () => {
-        const wrapper = shallow(<Drawer title="some title">some text</Drawer>);
-        // Open Drawer
-        wrapper.find('.Drawer__extendButton').simulate('click');
+    it('should be still hidden when drawer expented and isShown is false', () => {
```
eszthoff: typo: expented => expanded

stories/Drawer.js
```diff
@@ -1,18 +1,21 @@
...
-            title={text('Drawer title', '12/13 processed (2 need review, 1 failed)')}
-        >
-            {text(
-                'Content',
-                `
+    .add(
+        'Drawer',
+        () => (
+            <Drawer
+                isShown={boolean('isShown', true)}
```
eszthoff: can we add `initialIsExpanded` as a knob too?

stories/Drawer.js
```diff
@@ -1,18 +1,21 @@
...
+                isExpanded={boolean('isExpanded', true)}
```
eszthoff: can we demonstrate how the component behaves with `onClick` prop defined? We can add that prop conditionally: 
```
onClick={
  boolean('Use onClick prop', false)
    ? () => console.log('onClick callback has been called')
    : undefined
}
```

src/components/Drawer/Drawer.js
mukiienko: With div wrapper easy to add new buttons, now it is not the case, but in future it possible. We have in Search customer CSS, and often this CSS relies on HTML structure, so it is very painful to change HTML structure.

stories/Drawer.js
mukiienko: I don't see sense to add ```initialIsExpanded``` into knob. ```initialIsExpanded``` just impacts on initial render. Drawer doesn't reflect on this change  if we will change ```initialIsExpanded``` after render

src/components/Drawer/Drawer.js
eszthoff: I thought that the plan is not to allow custom css in the future but only change the theme variables we expose. Otherwise we will be very limited in many other places as well. Not important for this div, but for other places and general planning of component I would be concerned.

stories/Drawer.js
eszthoff: good point :-)

src/components/Drawer/__tests__/Drawer.spec.js
eszthoff: expanted => expanded ;-)

I have this spell checker extension in VSCode now. I'm usually terrible with this too, but it helps a lot. Checkout: Code Spell Checker
streetsidesoftware.code-spell-checker. It manages all the CamerCase words too.

mukiienko: Awesome! I will try it. Thanks!

src/components/Drawer/Drawer.js
```diff
@@ -49,14 +50,27 @@ const Drawer = props => {
         return STATUS_LIST.IS_HIDDEN;
     }
 
-    function handleClick() {
+    function triggeredIsExpanded() {
```
eszthoff: maybe call it `toggleDrawer` or `handleToggleCall` or `toggleIsExpanded`. The current name is not clear

mukiienko: Good point!

PR #172, opened 2019-09-23
fix: escape button press doesn't reset the state of Autosuggest component

igorarkhipenko:


src/components/Autosuggest/Autosuggest.js
```diff
@@ -75,20 +74,11 @@ class Autosuggest extends React.Component {
...
-                this.inputRef.current.parentElement.focus();
-                event.stopPropagation();
-            } else {
-                this.inputRef.current.blur();
-                this.handleBlur();
-            }
+            // prevents key propagation and sets the focus on parent component
+            this.inputRef.current.blur();
+            this.handleBlur();
+            this.inputRef.current.parentElement.focus();
```
eszthoff: what is the logic here of shifting focus to the div inside the FieldWrapper of Autosuggest?

igorarkhipenko: @eszthoff once the focus was removed from the input, the app doesn't handle events like key press anymore until you set it somewhere by your own. Just to make it more convenient I decided to move the focus to the parent element. For now t's used only in LocationSelector and the issue behind it was impossibility of closing the modal window by pressing on escape after closing the suggestions dropdown was closed by the same escape press.

eszthoff: I see. thanks

PR #173, opened 2019-09-25
fix(LocationSelector): rename components and fix unit tests

eszthoff:
Story: [ONEUI-95](https://jira.textkernel.nl/browse/ONEUI-95)

BREAKING CHANGE: LocationSelector and related components has been renamed. For top level use LocationSelector (instead of LocationSelectorWithGoogleLoader which has been deprecated). For sub components see documentation.

### DIFF
--- a/src/components/LocationSelector/LocationSelector.js
+++ b/src/components/LocationSelector/LocationSelector.js
@@ -1,21 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import bem from 'bem';
-import { LoadScriptNext } from '@react-google-maps/api';
 import { FaMapMarkerAlt } from 'react-icons/fa';
-import { Text, Modal, FieldWrapper, LoadingSpinner, LocationSelector } from '../../../index';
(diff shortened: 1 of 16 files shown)


src/components/LocationSelector/LocationSelector.js
```diff
@@ -83,13 +81,13 @@ function LocationSelectorWithGoogleLoader(props) {
...
-        findCenter(geocoder, location.place_id)
+        return findCenter(geocoder, location.place_id)
             .then(center => {
                 const locationToAdd = {
                     ...location,
                     center: {
-                        lng: center.lng(),
-                        lat: center.lat(),
+                        lng: typeof center.lng === 'function' ? center.lng() : center.lng,
+                        lat: typeof center.lat === 'function' ? center.lat() : center.lat,
```
antipin: I forgot, what's the point of having `lat ` `lng` as functions?

eszthoff: Sometimes google returns it as a component/function. I'm also a bit confused when is which, but this way we are sure to capture it.

PR #175, opened 2019-10-02
fix(LocationSelector): move rest props to the main wrapper

igorarkhipenko:


carlobernardini:
Tip: you can define a scope for your fix by using `fix(LocationSelector): title` in your PR / commit title.

PR #177, opened 2019-10-08
fix: refactor and arrange components z-index prop

igorarkhipenko:


src/components/LocationSelector/LocationSelector.js
```diff
@@ -12,6 +12,7 @@ const { block, elem } = bem({
     classnames: styles,
     propsToMods: ['muted'],
 });
+const modalStyleObject = { overlay: { zIndex: 500 } };
```
eszthoff: shouldn't we set this in the modal component to start with?

src/themes/oneui/base-variables/_misc.scss
```diff
@@ -1,5 +1,14 @@
 $transition-duration: .3s;
 
+/**
```
eszthoff: this is good, but we will never remember this is here. Can it be moved to storybook?

src/themes/oneui/base-variables/_misc.scss
```diff
@@ -1,5 +1,14 @@
...
+ * Layering of components by z-index guide
+ * [0-100] — regular components
```
eszthoff: maybe make this and the next range bigger. It is general practice to use steps of 10 for these things, so that it is easy to add things in between in the future, for that reason larger ranges are useful. Could even move the overall layover up to 700, as there are not much coming after that I suppose

src/components/Autosuggest/Autosuggest.scss
```diff
@@ -9,7 +9,7 @@
         transition-property: box-shadow, transform, border-color;
         transition-duration: var(--transition-duration);
         border-radius: var(--border-radius);
-        z-index: 10;
```
eszthoff: this is actually an overlay, the dropdown of autosuggest, isn't it? Then it should be in the second range.

src/components/Buttons/Button/Button.scss
```diff
@@ -27,7 +27,7 @@
 
     &:focus,
     &:active {
-        z-index: 100;
+        z-index: 2;
```
eszthoff: would 10 here to won't work? if it has to be higher then the autosuggest for some reason, maybe use 20.

src/components/Autosuggest/Autosuggest.scss
eszthoff: since we changed the range, this should go even higher :-|

PR #178, opened 2019-10-08
feat: new PopupBase component

eszthoff:
Story: [ONEUI-98](https://jira.textkernel.nl/browse/ONEUI-98)

### DIFF
--- a/src/__mocks__/popper.js.js
+++ b/src/__mocks__/popper.js.js
@@ -1,0 +1,12 @@
+import PopperJs from 'popper.js';
+
+export default class Popper {
+    static placements = PopperJs.placements;
+
+    constructor() {
+        return {
+            destroy: () => {},
+            scheduleUpdate: () => {},
+        };
+    }
+}
--- a/src/components/PopupBase/PopupBase.js
+++ b/src/components/PopupBase/PopupBase.js
@@ -5,20 +5,11 @@ import PopperJS from 'popper.js';
 import { POPUP_PLACEMENTS, ESCAPE_KEY } from '../../constants';
 
 class PopupBase extends React.Component {
-    static checkDOMPath(domElem, checkIsTarget) {
-        let currentElem = domElem;
-        while (currentElem && currentElem !== document) {
-            if (checkIsTarget(currentElem)) return true;
-            currentElem = currentElem.parentElement;
-        }
-        return false;
-    }
-
     constructor(props) {
         super(props);
         const { anchorRef, popupRef } = props;
 
-        this.state = { isOpened: false };
+        this.state = { isOpen: false };
         this.popper = undefined;
         this.anchorRef = anchorRef || React.createRef();
         this.popupRef = popupRef || React.createRef();
@@ -34,11 +25,11 @@ class PopupBase extends React.Component {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        const { isOpened } = this.state;
-        if (isOpened && !prevState.isOpened) {
+        const { isOpen } = this.state;
+        if (isOpen && !prevState.isOpen) {
             this.createPopperInstance();
         }
-        if (!isOpened && prevState.isOpened) {
+        if (!isOpen && prevState.isOpen) {
             this.destroyPopperIstance();
         }
     }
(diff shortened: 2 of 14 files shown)


src/__mocks__/popper.js.js
```diff
@@ -0,0 +1,12 @@
...
+export default class Popper {
+    static placements = PopperJs.placements;
+
+    constructor() {
+        return {
+            destroy: () => {},
+            scheduleUpdate: () => {},
+        };
+    }
+}
```
igorarkhipenko: Is duplicated extension in `popper.js.js` made on purpose?

src/components/PopupBase/PopupBase.js
```diff
@@ -0,0 +1,172 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import PopperJS from 'popper.js';
+import { POPUP_PLACEMENTS, ESCAPE_KEY } from '../../constants';
+
+class PopupBase extends React.Component {
+    static checkDOMPath(domElem, checkIsTarget) {
+        let currentElem = domElem;
+        while (currentElem && currentElem !== document) {
```
igorarkhipenko: I don't quite understand why exactly `while` loop is used here. It can be replaced with `if` since it's called with a single `domElem` each time or I'm missing something?

src/components/PopupBase/PopupBase.js
```diff
@@ -0,0 +1,172 @@
...
+
+    handleKeyPress(event) {
+        if (event.key === ESCAPE_KEY) {
+            this.close();
+        }
+    }
+
+    close() {
+        const { isOpened } = this.state;
+        if (isOpened === true) {
```
igorarkhipenko: can be simplified to `if (isOpened)`

src/components/PopupBase/PopupBase.js
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    createPopperInstance() {
+        if (!this.anchorRef.current || !this.popupRef.current) return;
+        const { placement } = this.props;
+        this.destroyPopperIstance();
+        this.popper = new PopperJS(this.anchorRef.current, this.popupRef.current, {
+            placement,
+        });
+    }
```
igorarkhipenko: I think it would be better to refactor it to
```
if (this.anchorRef.current && this.popupRef.current) {
        const { placement } = this.props;
        this.destroyPopperIstance();
        this.popper = new PopperJS(this.anchorRef.current, this.popupRef.current, {
            placement,
        });
}
```
just to remove an unnecessary exit point

src/components/PopupBase/PopupBase.js
```diff
@@ -0,0 +1,172 @@
...
+
+    destroyPopperIstance() {
+        if (!this.popper) return;
+        this.popper.destroy();
+        this.popper = undefined;
+    }
```
igorarkhipenko:
```
if (this.popper) {
    this.popper.destroy();
    this.popper = undefined;
}

File: src/components/PopupBase/PopupBase.js [hunk 6]
```diff
@@ -0,0 +1,172 @@
...
+        const { isOpened } = this.state;
+        if (!isOpened) return null;
+        const { popupRenderer } = this.props;
+        const popupElem = popupRenderer(this.getArgs());
+        const popupElemWithProps = React.cloneElement(popupElem, {
+            ref: this.popupRef,
+            'data-popup': 'true',
+        });
+        return ReactDOM.createPortal(popupElemWithProps, document.body);
+    }
```
igorarkhipenko: if (isOpened) {
        const { popupRenderer } = this.props;
        const popupElem = popupRenderer(this.getArgs());
        const popupElemWithProps = React.cloneElement(popupElem, {
            ref: this.popupRef,
            'data-popup': 'true',
        });

        return ReactDOM.createPortal(popupElemWithProps, document.body);
}

return null;

src/__mocks__/popper.js.js
eszthoff: yes. the package called `popper.js`. for Jest to see use it as a mock automatically each time it needs the package, we need to add `.js` after the package name. hence `popper.js.js`

src/components/PopupBase/PopupBase.js
```diff
@@ -0,0 +1,172 @@
...
+            elem => elem.dataset && !!elem.dataset.popup
+        );
+        const wasOutsideClicked = !this.popupRef.current.contains(event.target);
+
+        const wasAnchorClicked =
+            this.anchorRef.current && this.anchorRef.current.contains(event.target);
+        if (!wasPopupClicked && wasOutsideClicked && !wasAnchorClicked) {
+            this.close();
+        }
+    }
```
igorarkhipenko:
```
if (this.popupRef.current) {
        const wasPopupClicked = PopupBase.checkDOMPath(
            event.target,
            elem => elem.dataset && !!elem.dataset.popup
        );
        const wasOutsideClicked = !this.popupRef.current.contains(event.target);
        const wasAnchorClicked =
            this.anchorRef.current && this.anchorRef.current.contains(event.target);

        if (!wasPopupClicked && wasOutsideClicked && !wasAnchorClicked) {
            this.close();
        }
}
```

src/components/PopupBase/PopupBase.js
eszthoff: it is called with a single element, but then it cycles through all its parents and checks if that is the one we are targeting. E.g. there is an `a` in a `p` inside the popup component. When clicked, we don't want to close the popup. In order to see that it was indeed this popup that was clicked we cycle through all the parents and apply the `checkIsTarget` to them.

src/components/PopupBase/PopupBase.js
eszthoff: simplified the whole thing

PR #180, opened 2019-10-14
fix(List): keyboard navigation bugs

eszthoff:
Story: [JF-1780](https://jira.textkernel.nl/browse/JF-1780)

### DIFF
--- a/src/components/List/List.js
+++ b/src/components/List/List.js
@@ -1,9 +1,11 @@
 import React from 'react';
+// import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import bem from 'bem';
 import ListItem from './ListItem';
 import styles from './List.scss';
 import { LIST_NAVIGATION_DIRECTIONS, ENTER_KEY } from '../../constants';
+import scrollIntoViewIfNeeded from '../../utils/scrollIntoViewIfNeeded';
 
 const { block, elem } = bem({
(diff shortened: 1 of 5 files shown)


src/components/List/List.js
```diff
@@ -17,36 +19,37 @@ const NAVIGATION_STEP_VALUES = {
...
+        if (!children) return;
+
+        // eslint-disable-next-line no-plusplus
+        for (let i = 0; i < children.length; i++) {
+            const child = children[i];
+            if (!child.props[NOT_LIST_CHILD] && child.props.isSelected) {
+                setSelectedIndex(i);
+                break;
+            }
+        }
```
igorarkhipenko: just to avoid an additional function end point and `for` cycle we can refactor it to
```
if (children) {
    children.forEach((child, index) => {
        if (!child.props[NOT_LIST_CHILD] && child.props.isSelected) {
            setSelectedIndex(i);
        }
    });
}
```

eszthoff: using forEach doesn't allow you to break the loop. So early exit is not possible.

src/components/List/List.js
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
+// import ReactDOM from 'react-dom';
```
igorarkhipenko: should it be deleted?

PR #182, opened 2019-10-30
feat: new RadioButtonGroup component

eszthoff:
Story: [JF-1868](https://jira.textkernel.nl/browse/JF-1868)


src/components/RadioButton/__tests__/RadioButtonGroup.spec.js
```diff
@@ -0,0 +1,25 @@
...
+
+    it('should render correctly', () => {
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+    it('should pass name prop to children', () => {
+        wrapper
+            .find('RadioButton')
+            .forEach(option => expect(option.prop('name')).toEqual('group_name'));
+    });
+});
```
igorarkhipenko: Shall we add a test for checked item changes as well here?

eszthoff: That's a native behaviour, not part of our code. So I think it is ok to leave it out. In Radio Button I added that test mostly to see that rest props are passed correctly.

src/components/RadioButton/RadioButtonGroup.js
```diff
@@ -0,0 +1,49 @@
...
+    const { children, name, onChange, ...rest } = props;
+
+    const childProps = { name };
+    if (onChange) {
+        childProps.onChange = onChange;
+    }
+
+    return (
+        <div ref={ref} {...rest} {...block(props)}>
+            {React.Children.map(children, child => {
```
mukiienko: I just wondering why we need here {...block(props)} if the component doesn't have styles for block

eszthoff: I think this is just the standard of how we use OneUI. It will also ensure that className for consuming application will be passed correctly.

PR #186, opened 2019-11-05
WIP: build(typescript): migrated build pipeline to typescript

mukiienko:


.eslintrc.json
```diff
@@ -26,7 +35,37 @@
             }
         ],
         "react-hooks/rules-of-hooks": "error",
-        "react-hooks/exhaustive-deps": "warn"
+        "react-hooks/exhaustive-deps": "warn",
+        "@typescript-eslint/explicit-function-return-type": "off",
+        "@typescript-eslint/no-unused-vars": "off",
```
eszthoff: why to turn it off?

.eslintrc.json
```diff
@@ -26,7 +35,37 @@
...
+        "@typescript-eslint/interface-name-prefix": ["error", { "prefixWithI": "never" }],
+        "@typescript-eslint/ban-ts-ignore": "error"
```
eszthoff: we use eslint ignores when needed. why not ts ignore?

.eslintrc.json
```diff
@@ -26,7 +35,37 @@
...
+                "@typescript-eslint/no-var-requires": "off"
+            }
+        },
+        {
+            "files": [
+                "**/__tests__/**/*",
+                "**/__mocks__/**/*"
+            ],
+            "rules": {
+                "@typescript-eslint/camelcase": "off",
```
eszthoff: please explain

.eslintrc.json
```diff
@@ -26,7 +35,37 @@
...
+                "@typescript-eslint/no-empty-function": "off",
+                "@typescript-eslint/class-name-casing": "off"
```
eszthoff: this also unclear for OneUI

package.json
```diff
@@ -6,6 +6,7 @@
     "version": "8.2.1",
     "description": "Library of reusable React components with theming support",
     "main": "dist/oneui.min.js",
+    "types": "dist/bem.d.ts",
```
eszthoff: shouldn't this correspond to what is declared in tsconfig.js?

package.json
```diff
@@ -70,15 +71,19 @@
...
-        "@storybook/addon-info": "^5.1.8",
-        "@storybook/addon-knobs": "^5.1.8",
-        "@storybook/react": "^5.1.8",
+        "@storybook/addon-info": "5.1.8",
+        "@storybook/addon-knobs": "5.1.8",
+        "@storybook/react": "5.1.8",
         "@storybook/storybook-deployer": "2.8.1",
+        "@types/jest": "24.0.21",
+        "@types/node": "12.12.5",
+        "@types/react": "16.9.11",
```
eszthoff: should we get types for react-dom too?

src/components/Autosuggest/Autosuggest.js
```diff
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types';
 import Downshift from 'downshift';
 import memoize from 'fast-memoize';
 import bem from '../../utils/bem';
-import { List, ListItem, ContentPlaceholder, Text, MarkedText, FieldWrapper } from '../../index';
```
eszthoff: why cannot we import everything together? seems to be easier to type and easier to refactor

src/components/Input/Input.tsx
```diff
@@ -1,12 +1,26 @@
 import React, { forwardRef } from 'react';
-import PropTypes from 'prop-types';
 import bem from '../../utils/bem';
 import styles from './Input.scss';
-import { CONTEXTS, INPUT_TYPES, SIZES } from '../../constants';
+import { ContextTypes, InputTypes, SizesTypes } from '../../constants';
+
+// Any other attributes (onChange, onKeyUp etc.) are
+// supported although not defined in props type definition
+interface Props extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'> {
```
eszthoff: At our meeting we said that as part of this ticket there will be some document as to say what the developer guidelines are. Like when to use `interface` and when `types`. Could you address it?

src/components/Input/Input.tsx
```diff
@@ -1,12 +1,26 @@
...
+    /** The input field context (e.g. brand, primary, bad, good etc. - defaults to brand) */
+    context: ContextTypes;
```
eszthoff: in Storybook these now show up as `type: unknown`. Is it something we could remedy?

stories/Modal.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { text, withKnobs } from '@storybook/addon-knobs';
-import { withInfo } from '@storybook/addon-info';
+import { withInfo } from '@storybook/addon-info'; // eslint-disable-line import/no-unresolved,
```
eszthoff: I don't think we need this ignore

stories/Input.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import { storiesOf } from '@storybook/react';
-import { boolean, select, text, withKnobs } from '@storybook/addon-knobs';
+import { storiesOf } from '@storybook/react'; // eslint-disable-line import/no-extraneous-dependencies
+import { boolean, select, text, withKnobs } from '@storybook/addon-knobs'; // eslint-disable-line import/no-extraneous-dependencies
```
eszthoff: I don't think we need these ignores

package.json
```diff
@@ -124,6 +130,9 @@
         "stylelint-scss": "3.3.1",
         "stylelint-webpack-plugin": "0.10.5",
         "terser-webpack-plugin": "1.2.3",
+        "ts-jest": "24.1.0",
+        "ts-loader": "6.2.1",
+        "typescript": "3.6.4",
```
eszthoff: 3.7 has just been released. let's use that.

mukiienko: Done! I Added extension "*.tsx" to .eslintrc.json for storybook.

stories/Input.tsx
mukiienko: Done! I Added extension "*.tsx" to .eslintrc.json for storybook.

src/components/Input/Input.tsx
eszthoff: That was not my understanding from the meeting. If we have no guidelines on things that are easy to write in different ways, we will end up with a mess pretty soon. It is not even sure then if this is something we will need to change right after it is submitted. I don't mean we have to have a comprehensive guide to TS, only some basic standards that we already run into, such as using type or interface and when.

src/components/Input/Input.tsx
eszthoff: I found it a quite important regression and this should be brought up to the whole team and we need to all agree that on if/how to go ahead with it.

src/components/Input/Input.tsx
antipin: For `interfaces vs types` I'd suggest to stick with interfaces, cause `interface` allows to extend Props from 3rd party libs and it can be useful sometimes (as we see above). So for the sake of consistency `interface` would be nicer from my pov.

eszthoff: Sounds ok to me. Honestly, I had no time to look into the differences between them, but I see that this PR already uses both. So it would be good to know when which.

antipin: Hmm, I'd rather go with one option for consistency and less time spent on choosing what to use for a particular cause

eszthoff: Maybe you can discuss it with Alexey and come up with a convention. Currently src/constants/style-related.ts uses `type`.

src/components/Input/Input.tsx
```diff
@@ -22,29 +36,12 @@ const Input = forwardRef((props, ref) => {
...
-};
-
 Input.defaultProps = {
     context: 'brand',
     disabled: false,
     isBlock: false,
-    size: 'normal',
     type: 'text',
+    size: 'normal',
     value: undefined,
```
eszthoff: Please remove this ;-)

CONTRIBUTING.md
```diff
@@ -56,6 +56,25 @@ Button.propTypes = {
 };
 ```
 
+### TypeScript
+
+### Type definitions for Props and States
+Always use `interface` for type definitions Props and States. Interfaces are more “extendable” due to the support of declaration merging. It is useful when components supports native HTML props, for example:
```
carlobernardini: Little grammar check

Always use `interface` when defining types for props and states. Interfaces are more "extendable" because of support for declaration merging. This is useful when components support native HTML props, for example:

CONTRIBUTING.md
```diff
@@ -56,6 +56,25 @@ Button.propTypes = {
...
+
+```javascript
+interface Props extends <HTMLInputElement> {
+    /** Description of this props  */
+    context: string;
+}
+```
+
+### Using Type
+Types are useful for union types. So its are more “composable” :
```
carlobernardini: Little grammar check 2

Types are useful for union types, meaning they are more "composable":

src/components/Input/Input.tsx
mukiienko: Done

CONTRIBUTING.md
mukiienko: Done

src/types/modules.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module '*.scss' {
+  interface IClassNames {
```
antipin: Should be without `I` prefix

PR #187, opened 2019-11-11
fix(storybook): fix error in Autosuggest story

eszthoff:
Story: [ONEUI-90](https://jira.textkernel.nl/browse/ONEUI-90)

### DIFF
--- a/src/packages/storybook/withStore/withStore.js
+++ b/src/packages/storybook/withStore/withStore.js
@@ -54,10 +54,12 @@ class StoreInjector {
 
 const storeInjectors = [];
 
-export { StoreInjector };
-export default function(injectedStore) {
+const injectStore = injectedStore => {
     const injector = new StoreInjector();
     const index = storeInjectors.length;
     storeInjectors.push(injector);
     return storeInjectors[index].withStore(injectedStore);
-}
+};
+
+export { StoreInjector, storeInjectors };
+export default injectStore;
--- a/src/packages/storybook/withStore/withStore.js
+++ b/src/packages/storybook/withStore/withStore.js
@@ -52,14 +52,14 @@ class StoreInjector {
     }
 }
 
-const storeInjectors = [];
+StoreInjector.stores = [];
 
-const injectStore = injectedStore => {
+StoreInjector.initStore = injectedStore => {
     const injector = new StoreInjector();
-    const index = storeInjectors.length;
-    storeInjectors.push(injector);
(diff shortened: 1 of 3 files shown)


stories/Autosuggest.js
```diff
@@ -75,7 +75,8 @@ storiesOf('Organisms|Autosuggest', module)
         };
 
         const getSelectedPlaceholder = () => {
-            const numOfItems = store.get('selectedSuggestions').length;
+            const items = store.get('selectedSuggestions');
+            const numOfItems = items ? items.length : 0;
```
carlobernardini: Matter of 'taste' but I think this could be even simpler like this:

`const numOfItems = items.length || 0`

Either way it fixes the issue.

stories/Autosuggest.js
```diff
@@ -75,7 +75,8 @@ storiesOf('Organisms|Autosuggest', module)
...
             if (!numOfItems) {
```
igorarkhipenko: maybe it would be better to change the condition to `numOfItems === 0` just to be more transparent and consistent with the next one?

stories/Autosuggest.js
eszthoff: if we use this construct you would get the same error as before, when `items` is undefined

carlobernardini: Ah true

src/packages/storybook/withStore/withStore.js
```diff
@@ -52,9 +52,14 @@ class StoreInjector {
     }
 }
 
-const storeInjector = new StoreInjector();
+const storeInjectors = [];
```
mukiienko: Good improve! I have a small proposal.
`storeInjectors` is lools like a good candidate for a static variable of `StoreInjector` class.
```
StoreInjector.stores = []
```
What do you think?

src/packages/storybook/withStore/withStore.js
```diff
@@ -52,9 +52,14 @@ class StoreInjector {
...
 
-export { StoreInjector };
-export default function(injectedStore) {
-    return storeInjector.withStore(injectedStore);
-}
+const injectStore = injectedStore => {
+    const injector = new StoreInjector();
+    const index = storeInjectors.length;
+    storeInjectors.push(injector);
+    return storeInjectors[index].withStore(injectedStore);
```
mukiienko: Actually I think to make sense to extend class `StoreInjector` with this logic and move this code to like static method `StoreInjector.create()` for example.

PR #188, opened 2019-11-11
build: implemented script that checks npm access before publishing

antipin:
Resolves ONEUI-60

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
         "lint:styles": "stylelint 'src/**/*.s(a|c)ss'",
         "test": "jest --config scripts/test/jest.config.js",
         "test:coverage": "jest --config scripts/test/jest.config.js --coverage",
-        "prepublishOnly": "npm test && npm run lint:ci && npm run lint:styles && npm run build:prod",
+        "check-npm-publish-access": "sh scripts/check-npm-publish-access/check-npm-publish-access.sh",
+        "prepublishOnly": "npm run check-npm-publish-access && npm run test && npm run lint:ci && npm run lint:styles && npm run build:prod",
         "coveralls": "cat ./coverage/lcov.info | node node_modules/.bin/coveralls",
         "release": "standard-version -t ''"
     },
--- a/scripts/check-npm-publish-access/check-npm-publish-access.sh
+++ b/scripts/check-npm-publish-access/check-npm-publish-access.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Exit immediately if any command exits with a non-zero status
+set -e;
+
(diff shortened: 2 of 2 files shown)


scripts/check-ownership/check-ownership.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
eszthoff: maybe we can put it in a folder `script/npm` so we can use it for other things in the future if we have? or just put the file right under `script`?

antipin: > maybe we can put it in a folder script/npm so we can use it for other things in the future if we have

Why can't we use it from `scripts/check-ownership/check-ownership.sh` in the future if we need?

I'd prefer to keep `scripts` folder in a more well defined structure:
```
└── scripts
    ├── <action-name> // directoery
    │   └── // one or more files related to the action
    └── <anither-action-name> // directoery
        └── // one or more files related to the action
```

E.g. in next-gen-search project it looks like this: http://tkgit-01.textkernel.net/searchbox/search-ui/tree/dev/scripts
```
└── scripts
    ├── build
    ├── lint
    ├── lint-commits
    ├── lint-styles
    └── test
```

eszthoff: Those were just my 2 cents. If we don't have 100s of scripts, it is not really important. I don't think we ever discussed any guidelines about it. The PR is approved, you can merge it.

package.json
```diff
@@ -17,7 +17,7 @@
         "lint:styles": "stylelint 'src/**/*.s(a|c)ss'",
         "test": "jest --config scripts/test/jest.config.js",
         "test:coverage": "jest --config scripts/test/jest.config.js --coverage",
-        "prepublishOnly": "npm test && npm run lint:ci && npm run lint:styles && npm run build:prod",
+        "prepublishOnly": "sh scripts/check-ownership/check-ownership.sh && npm test && npm run lint:ci && npm run lint:styles && npm run build:prod",
```
mukiienko: Can we move `sh scripts/check-ownership/check-ownership.sh` to separate command?
Profit: 
- we can run check ownership without the whole build process;
- `prepublishOnly` command will be shorter;
What do you think?

scripts/check-ownership/check-ownership.sh
antipin: ok, thanks,  I'll do. Btw, what do you think in general about keeping all the scripts under `scripts/<action_name>` directory? Just in order not to keep root pf the project clean and all this stuff more organized?
image

package.json
antipin: good idea

PR #189, opened 2019-11-12
feat(SearchButton): Extend SearchButton with an optional label

igorarkhipenko:
Story [JF-1907](https://jira.textkernel.nl/browse/JF-1907)

### DIFF
--- a/src/components/Buttons/SearchButton/SearchButton.js
+++ b/src/components/Buttons/SearchButton/SearchButton.js
@@ -8,16 +8,11 @@ const { block, elem } = bem('SearchButton', styles);
 
 const SearchButton = forwardRef((props, ref) => {
     const { disabled, children, type, ...rest } = props;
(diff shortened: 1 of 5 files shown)


src/components/Buttons/SearchButton/SearchButton.js
```diff
@@ -4,14 +4,21 @@ import { FaSearch } from 'react-icons/fa';
...
 const SearchButton = forwardRef((props, ref) => {
-    const { disabled, type, ...rest } = props;
+    const { disabled, children, type, ...rest } = props;
 
     return (
-        <button {...rest} {...block(props)} ref={ref} type={type} disabled={disabled}>
-            <FaSearch />
+        <button
+            {...rest}
+            {...block({ ...props, withLabel: !!children })}
```
eszthoff: maybe can extract `{ ...props, withLabel: !!children }` to a const (e.g. `propsForBem`) so that it is only in one place

PR #190, opened 2019-11-19
fix(LocationSelector): fix input focus

igorarkhipenko:
Story [ONEUI-99](https://jira.textkernel.nl/browse/ONEUI-99)


src/components/Autosuggest/Autosuggest.js
```diff
@@ -14,23 +14,27 @@ import {
...
-        this.rootRef = React.createRef();
-        this.listRef = React.createRef();
+
+        this.inputRef = props.inputRef || React.createRef();
+        this.rootRef = props.rootRef || React.createRef();
+        this.listRef = props.listRef || React.createRef();
+
+        this.handleTagDeleteClick = memoize(this.handleTagDeleteClick);
+        this.handleWrapperClick = memoize(this.handleWrapperClick);
+        this.handleWrapperKeyDown = memoize(this.handleWrapperKeyDown);
```
mukiienko: Seems memoization doesn't work correctly here. I typed the same in input several times but function `this.handleWrapperKeyDown` was invoked on every change.

src/components/Autosuggest/Autosuggest.js
```diff
@@ -374,6 +412,12 @@ Autosuggest.propTypes = {
     noSuggestionsPlaceholder: PropTypes.string.isRequired,
     /** to be shown as clear button title */
     clearTitle: PropTypes.string,
+    /** input field ref */
+    inputRef: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+    /** root wrapper ref */
+    rootRef: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+    /** suggestions list ref */
+    listRef: PropTypes.object, // eslint-disable-line react/forbid-prop-types
```
mukiienko: Can we try to define `PropTypes` in this way: https://stackoverflow.com/questions/48007326/what-is-the-correct-proptype-for-a-ref-in-react

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.js
```diff
@@ -34,15 +36,26 @@ const LocationSelectorDialog = props => {
         onCloseModal,
     } = props;
 
+    function handleAddLocation(location) {
+        setTimeout(() => locationInputRef.current.focus());
+        onAddLocation(location);
+    }
+
+    function handleRemoveLocation(locationId) {
+        setTimeout(() => locationInputRef.current.focus());
```
mukiienko: Can we skip timeout? Maybe add some hooks on `locationId` if it changes then focus input?

src/components/Autosuggest/Autosuggest.js
eszthoff: I wouldn't waste much time on it, if we are going to move it to TS soon

src/components/Autosuggest/Autosuggest.js
igorarkhipenko: @mukiienko I didn't quite understand why it should not be invoked on each change, could you clarify it please? And also it seems to be a class component which we should split onto a few separate ones with rewriting it to hooks logic, so memoization will go away soon anyway.

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.js
igorarkhipenko: @mukiienko is there something wrong with focusing input wrapped in setTimeout? We don't have an access to locationId here since it's an argument for event handling functions.

src/components/Autosuggest/Autosuggest.js
mukiienko: I thought it needs for memoization result, but it memoizes reference on the function. I was mistaken :)

src/components/Autosuggest/Autosuggest.js
mukiienko: Yes, it makes sence

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.js
mukiienko: Not transparent why the focus needs a timeout. It looks like a workaround for some problems with focusing. That code can cause problems in the future, for instance, in below code execution we need to add focus for another element, but it will not work because async focus rewrites sync focus, so we need for another focus add a timeout as well.

PR #191, opened 2019-11-22
Move components to TypeScript

igorarkhipenko:
Story: [ONEUI-106](https://jira.textkernel.nl/browse/ONEUI-106)
Move Button, CandidateAvatar and Callout components to TypeScript


src/components/Buttons/Button/Button.tsx
```diff
@@ -1,12 +1,32 @@
...
 import bem from '../../../utils/bem';
 import styles from './Button.scss';
-import { CONTEXTS, SIZES } from '../../../constants';
+import { ButtonTypes, ContextTypes, SizesTypes } from '../../../constants';
+
+// Any other attributes (onClick, onFocus etc.) are
+// supported although not defined in propTypes
+interface Props extends Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'size'> {
+    /** The label of the button */
+    children: [string, Node];
```
antipin: `[string, Node]` declares a tuple that accepts string as a first element and Node as second.
Syntax of "OneOf" logic is `A | B`.

Also, `Node` is not `ReactNode`, it's a [DOM Node](https://developer.mozilla.org/en-US/docs/Web/API/Node)

Should be
```
children: React.ReactNode;
```

This applies to all the other places where `Node` is being used.

src/components/Buttons/Button/Button.tsx
```diff
@@ -1,12 +1,32 @@
-import React, { forwardRef } from 'react';
-import PropTypes from 'prop-types';
+import React from 'react';
 import bem from '../../../utils/bem';
 import styles from './Button.scss';
-import { CONTEXTS, SIZES } from '../../../constants';
+import { ButtonTypes, ContextTypes, SizesTypes } from '../../../constants';
```
antipin: I'd put `ButtonTypes, ContextTypes, SizesTypes` (and all the other types) in singular form.

src/components/CandidateAvatar/CandidateAvatar.tsx
```diff
@@ -1,21 +1,31 @@
...
 const { block, elem } = bem('CandidateAvatar', styles);
 
-const context = percentage => {
-    if (percentage <= 33) {
-        return 'bad';
-    }
-    if (percentage <= 66) {
-        return 'warning';
+const CandidateAvatar = (props: Props) => {
+    function getContext(percentage: number): string {
```
antipin: Here, return type can be narrowed down from `string` to `'bad' | 'warning' | 'good'`

src/constants/component-specific.ts
```diff
@@ -3,6 +3,11 @@
 /** Autocomplete */
 export const NUMBER_OF_SUGGESTION_LOADING_PLACEHOLDERS = 5;
 
+/** Button */
+export const BUTTON_TYPES = ['submit', 'button'] as const;
+
+export type ButtonTypes = typeof BUTTON_TYPES[number];
```
antipin: Might be here enum would fit better, cause it can be both: a type and a set of values:
```ts
enum ButtonType  {
    submit = 'submit',
    button = 'button',
}
```

src/components/Buttons/Button/Button.tsx
eszthoff: What about using `TButton`, `TContext`, etc.?

igorarkhipenko: I think we should stick with `Type` suffix (prefix?)
https://github.com/microsoft/TypeScript/blob/master/lib/typescript.d.ts#L2388

Do you guys follow any regularities in `Type` part moving around the variable name? Please help us if you do

antipin: We started with prefixes but later decided not to use any prefixes and suffixes – it's too much noise and seems like it doesn't add much value.

antipin: In this particular case `ButtonType` makes sense cause it's actually about a type of a button. But `ContextType` and `SizesType` are redundant and can be written as `Size` and `Context`. Leading capital indicates that it's a type or enum.

eszthoff: Did you never have a problem with name clashes? e.g. there is a variable size, but also the var that describes it called the same? I guess more relevant for classes and such that are PascalCased

antipin: That was my initial concern (conflicts of names) when we decided to use prefixes, but in real life for us that never happened. The only thing that I can imagine is `Class` thet implements some `Interface` with the same name. But again, never happend to us :)

src/constants/style-related.ts
```diff
@@ -13,5 +13,5 @@ export const SIZES = ['small', 'normal', 'large'] as const;
 export const HEADING_SIZES = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] as const;
 export const CHECKBOX_VIEWBOX = [0, 0, 10, 10] as const;
 
-export type ContextType = typeof CONTEXTS[number];
-export type SizeType = typeof SIZES[number];
+export type Context = typeof CONTEXTS[number];
+export type Size = typeof SIZES[number];
```
antipin:
```
typeof SIZES[number];
```
this is not the most obvious TS construction. Might be makes sense to add a comment here.

PR #194, opened 2019-11-29
refactor(TS migration): migrate WidthRestrictor, Header, Footer, Navigation and Link

eszthoff:
Storey: [ONEUI-109](https://jira.textkernel.nl/browse/ONEUI-109)


package.json
```diff
@@ -34,7 +34,7 @@
     },
     "husky": {
         "hooks": {
-            "pre-commit": "npm run test && npm run lint:ci && npm run lint:styles"
+            "pre-commit": "npm run tsc && npm run test && npm run lint:ci && npm run lint:styles"
```
mukiienko: It seems time to use here `npm-run-all`. The code will be mode shorter and readable.

src/components/Buttons/Button/Button.tsx
```diff
@@ -7,19 +7,19 @@ import { ButtonType, Context, Size } from '../../../constants';
 // supported although not defined in propTypes
 interface Props extends Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'size'> {
     /** The label of the button */
-    children: string | React.ReactNode;
+    children: string | React.ReactElement;
```
mukiienko: Why we can not use here `React.ReactNode`?

src/components/Buttons/Button/Button.tsx
```diff
@@ -7,19 +7,19 @@ import { ButtonType, Context, Size } from '../../../constants';
...
+    isInline?: boolean;
     /** Should button be disabled or not */
-    disabled: boolean;
+    disabled?: boolean;
     /** Type of the button */
-    type: ButtonType;
+    type?: ButtonType;
     /** Providing an href will render an <a> element, styled as a button. */
     href?: string;
 }
```
mukiienko: seems you forgot to move to typeScript `Button/index.js`, `Callout/index.js`, `CandidateAvatar/index.ts`

src/components/Callout/Callout.tsx
```diff
@@ -6,14 +6,14 @@ import styles from './Callout.scss';
...
 interface Props {
     /** The Callout context (e.g. brand, primary, bad, good etc. - defaults to info) */
-    context: Context;
+    context?: Context;
     /** A property specifies the height of a line. */
-    lineHeightStyle: number;
+    lineHeightStyle?: number;
     /** Content to be rendered inside the container */
-    children: React.ReactNode;
+    children: string | React.ReactElement;
```
mukiienko: The same about React.ReactNode?

src/components/CandidateAvatar/CandidateAvatar.tsx
```diff
@@ -24,11 +24,11 @@ const CandidateAvatar = (props: Props) => {
             return 'warning';
         }
         return 'good';
-    }
+    };
 
     const { imageUrl, matchPercentage, showPercentageOnHover, size, ...rest } = props;
 
-    const constrainedSize = Math.round(Math.max(32, size));
+    const constrainedSize = Math.round(Math.max(32, size!));
```
mukiienko: I have typescript tip here in my IDE `Forbidden non-null assertion.`.
Can we use here Number(size)? That will be more clear.

src/components/Icon/IconBase.tsx
```diff
@@ -33,7 +50,7 @@ const IconBase = props => {
                 aria-labelledby={title ? 'title' : null}
                 xmlns="http://www.w3.org/2000/svg"
                 viewBox={viewBox}
-                style={adjustSize(size, preserveAspectRatio)}
+                style={adjustSize(preserveAspectRatio!, size)}
```
mukiienko: Can we use here Boolean(preserveAspectRatio)?

src/components/Icon/IconTextkernel/IconTextkernel.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import IconBase, { Props as BaseProps } from '../IconBase';
+
+interface Props extends Omit<Omit<BaseProps, 'children'>, 'viewbox'> {}
```
mukiienko: Should we remove `viewbox`? It can be used like `IconBase` prop  ?

src/components/Callout/Callout.tsx
```diff
@@ -6,14 +6,14 @@ import styles from './Callout.scss';
...
     /** A function to be called when close button was clicked */
-    onRequestClose: () => void;
-};
+    onRequestClose?: () => void;
+}
 
 const { block, elem } = bem('Callout', styles);
```
mukiienko: React.FC<Props>  was missed for Callout

src/components/Buttons/Button/Button.tsx
eszthoff: we had a whole discussion about it on slack, didn't we? `React.Node` can be `null | undefined`, so you practically putting it as optional, even though that's not the intention

src/components/Buttons/Button/Button.tsx
eszthoff: Yes, they were converted in a different PR, but I can approve on it...

CONTRIBUTING.md
```diff
@@ -58,8 +66,23 @@ Button.propTypes = {
...
+```javascript
+// GOOD:
+import * as React from 'react';
+```
+
+```javascript
+// BAD:
+import React from 'react';
+```
+
```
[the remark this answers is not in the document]
eszthoff: :-)

src/components/Icon/IconTextkernel/IconTextkernel.tsx
eszthoff: I don't understand what do you mean. 'Viewbox' is hard coded based on the svg path. You should not change it through the props. You can however change the size of the icon, using `size`

CONTRIBUTING.md
```diff
@@ -68,35 +91,55 @@ interface Props extends <HTMLInputElement> {
...
 
+-   Define **optional params** in `defaultProps` only if they have a meaningful value. In FunctionalComponents also mark them with `?` as in `propName?: type`. Without setting the prop to optional with `?` we have 2 problems:
+
+    -   TS will think it is required when the component is used (_last updated: TS 3.7_).
+
+    -   It will show us as required in storybook even if its type can be undefined.
+
+    _Note_: this can lead to annoying situations where you have to add additional type checking in the code, if you really have to make sure that a prop is not null
+
+-   For **optional node** props use `propName?: React.Node` rather then setting the prop to `null` in `defaultProps`.
```
mukiienko: React.Node => ReactNode

PR #195, opened 2019-11-29
refactor(TS migration): migrate LoadingSpinner, Heading and Text

eszthoff:
Story: [ONEUI-110](https://jira.textkernel.nl/browse/ONEUI-110)


src/components/Text/Text.tsx
```diff
@@ -1,12 +1,22 @@
...
-import PropTypes from 'prop-types';
+import * as React from 'react';
 import bem from '../../utils/bem';
 import styles from './Text.scss';
-import { SIZES, CONTEXTS } from '../../constants';
+import { Size, Context } from '../../constants';
+
+interface Props extends React.HTMLAttributes<HTMLElement> {
+    /** Text content */
+    children: React.ReactElement;
```
mukiienko: Maybe make sense write here `React.ReactNode`?. The component will be more flexible and can apply null, undefined... I think it is a use case. What do you think?

eszthoff: I did not change the way components are, just migrated them. If we allow for no children, we should that and not render empty html elements. I would rather do these improvements in separate PR. Do you want to make a ticket for it?

mukiienko: ...hmm, for me not obviously why for `Text` we don't able apply to pass null, undefined, false. Some example:
```
<Text>
{isShow && 'some text' }
</Text>
```
What do you think it can be use case?

mukiienko: Anyway, `Text` should support simple text node like this `<Text>hello</Text>` I think.

mukiienko: Text can be `number`?

PR #196, opened 2019-12-04
feat(Tooltip): add tooltip component

igorarkhipenko:


stories/Tooltip.js
```diff
@@ -0,0 +1,21 @@
...
+import { storiesOf } from '@storybook/react';
+import { withKnobs, select } from '@storybook/addon-knobs';
+import { Tooltip } from '@textkernel/oneui';
+import { POPUP_PLACEMENTS } from '@textkernel/oneui/constants';
+
+storiesOf('Molecules|Tooltip', module)
+    .addDecorator(withKnobs)
+    .add('Tooltip', () => (
+        <div>
+            <div style={{ display: 'table', margin: '0 auto' }}>
```
eszthoff: why do we need these divs, and `display: table`? It is unlikely that we will ever use it in this way in a real application. Can we demonstrate the function by e.g. rendering it within `Text`?

stories/Tooltip.js
```diff
@@ -0,0 +1,21 @@
...
+                <Tooltip
+                    placement={select('Placement', POPUP_PLACEMENTS, 'bottom')}
+                    content="this is my tooltip text"
```
eszthoff: lets make this and children changeable in the props

stories/Tooltip.js
igorarkhipenko: I added this style to move the component to the middle of the screen in order to all of the placement options be viable

eszthoff: `margin: 0 auto` makes sense to me. but not `display: table`. and also don't see why we need 2 divs instead of just one.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,59 @@
...
+const { elem } = bem('Tooltip', styles);
+
+interface Props {
+    children: React.ReactNode;
+    content: React.ReactNode;
+    placement?: PopupPlacement;
+}
+
+const Tooltip: React.FC<Props> = props => {
+    const { placement, content, children } = props;
```
eszthoff: missing `...rest`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </div>
+    );
+
+    const renderPopup = () => (
+        <div>
+            <div {...elem('container', props)}>
+                {content}
+                <div {...elem('arrow', props)} />
+            </div>
+        </div>
```
mukiienko: Can we get rid of div wrapper? I tried to delete it locally, but this change brokes `Tooltip`. This div applies styles `top: 0, left: 0` that go probably from  `PopperJS` and when I deleted  `top: 0, left: 0`  popup appears correctly. That fixed `Edge` and `IE11` as well.
Screenshot 2019-12-06 at 11 25 33

Screenshot 2019-12-06 at 12 06 09

PR #201, opened 2019-12-16
refactor(TypeScript): migrate ButtonGroup, Icons, ContentPlaceholder, Chip, TwoPaneView, Pagination

eszthoff:
Story: [ONEUI-111](https://jira.textkernel.nl/browse/ONEUI-111)


eszthoff:
> I see this warning when I do `npm run test` (in dev it doesn't exist)
> image

Yes, it is there since a long time. I think it is because of JSDom doesn't have the actual width and such defined, but we never investigated it... We should make a separate ticket for it.

PR #204, opened 2019-12-20
feat(Tabs): Rewrite Tabs API

eszthoff:
Story: [ONEUI-126](https://jira.textkernel.nl/browse/ONEUI-126)

BREAKING CHANGE: Tabs, Tab, TabMenu, TabContent  were deprecated along with some of their features in favour of TabsBar and TabItem


carlobernardini:
Few comments in advance:

* Please make border color below tablist color-neutral-25
* Please give tab labels font-weight-bold
* When I hover over disabled tabs they have their normal hover styles, please make those stick to their normal styles (no difference on hover except for the not-allowed cursor)
* Does it already allow having 'secondary text' (like the `(3)` next to the label as per designs)?

mukiienko:
I suggested small UI/UX improvements:
1)  to add immediate feedback after a click. I mean add some styles `:active` to the button.
2) when I use keyboard navigation Tab doesn't highlight, add `:focus`
3) Active Tab has default `cursor`, I'd rather keep 'pointer'. Now It looks strange when all  tabs have a pointer but active tab `default`.

mukiienko:
For IE11 tabs don't switch :(

src/components/Tabs/TabsBar/TabsBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+const TabsBar: React.FC<Props> = props => {
+    const { activeTabId, children, onSelect, isBlock, ...rest } = props;
+
+    const extendWithProps = (tab: Props['children']) => {
+        if (React.isValidElement(tab) && tab.type === TabItem) {
+            const { tabId } = tab.props;
+            const extendProps: {
+                isBlock?: boolean;
+                isActive?: boolean;
+                onSelect?: (selectedTabID: number | string) => void;
```
mukiienko: This TypeScript assigning a bit confused me, can we move it to the top of the file?

carlobernardini: Tiny detail but should this perhaps be `selectedTabId` (lowercase `d`) to be consistent with other places that mention `tabId`?

src/components/Tabs/TabItem/TabItem.scss
```diff
@@ -1,49 +1,45 @@
...
+    &:focus {
+        border-bottom: 4px solid var(--color-neutral-25);
+        color: var(--link-color-hover);
     }
 
-    &:first-child {
-        margin-left: 0;
+    &--isActive,
+    &:active {
+        color: var(--color-foreground);
```
mukiienko: I think this incorrect:
```
&:active {
 &:hover {}
}
```
It will be gerate CSS:
```.TabItem--20c4ZfdZd9:active:focus,.TabItem--20c4ZfdZd9:active:hover,.TabItem--isActive--1NIM9nd7JM```

PR #210, opened 2020-01-02
fix(Footer): add year prop to Footer component to make it determenistic

antipin:
Hardcoded `new Date().getFullYear();` caused test failure


carlobernardini:
Checks failed only because the coverage supposedly dropped too low, but test coverage for this component is 100% so this seems to be a false negative...

carlobernardini (approved):

PR #213, opened 2020-01-03
feat(SelectButtonGroup): new components

eszthoff:
Story: [ONEUI-116](https://jira.textkernel.nl/browse/ONEUI-116)


src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    children: string;
+    /** the value associated with this button */
+    value: string;
+    /**
+     * A function to be called if a button is pressed.
+     * This prop, if defined on the parent SelectButtonGroup will overwrite this one
+     */
+    onSelect?: (value: string) => void;
+    /** if this button should be in a selected state */
+    isSelected?: boolean;
```
carlobernardini: Is this a suitable interface in case of a single select (radios)? Can you make sure there is only one value selected or is that up to the integrating application? For single selects what you typically see is e.g. `value` (for specifying selected value) on the outer component (`SelectButtonGroup` in this case) but on the other hand, that's probably less suitable for multi select.

eszthoff: in native HTML you have `checked` both on radio button and checkbox. So I think it will work here too

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+
+interface Props {
+    /** the label to be displayed on this button */
+    children: string;
```
mukiienko: I'd in favor to pass any react element for more flexibility.

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    /** the value associated with this button */
+    value: string;
```
mukiienko: I think value also can be number or boolean.
For example, FE retrieves some boolean (true/false) value from BE and renders this value like two switch box buttons disabled/enabled. In the current implementation, we need to convert it to string and then back to boolean.

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * A function to be called if a button is pressed.
+     * This prop, if defined on the parent SelectButtonGroup will overwrite this one
+     */
+    onSelect?: (value: string) => void;
+    /** if this button should be in a selected state */
+    isSelected?: boolean;
+    /** the color context to be applied in the selected state */
+    selectedContext?: 'neutral' | 'brand';
```
mukiienko: I think better pass the entire `Context`.  Maybe in the future, we will decide to use other color schemas

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * is this button part of a multiselect group
+     * when yes will render as checkbox otherwise as radioButton
+     * This prop will be overwritten by the parent <SelectButtonGroup>
+     */
+    isMultiselect?: boolean;
```
mukiienko: I think we don't need `isMultiselect` for SelectButton component. It already exists in the parent. It can cause problems. For example, if set for one `isMultiselect=true`, and for others `isMultiselect=false` so what will be output?

src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+
+interface Props {
+    /** SelectButton components */
+    children: NotEmptyReactNode;
+    /**
+     * A function to be called if a button is pressed.
+     * This prop, if defined will overwrite the onSelect on children
+     */
+    onSelect?: (value: string) => void;
```
mukiienko: If it is `isMultiselect` i think better to pass all selected values.
Not clear if we uncheck value so what `onSelect` should return?
Passed all selected values solve this problem.

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
eszthoff: I think we can extend it later if we have design for it. The ticket currently describes only these 2.

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
eszthoff: We need it because that what determines if it will render as a radio button or as a checkbox. It is not really meant to be specified by the user of the API but passed down from the parent, as says in the comments above. However, we do need a way to pass it down, so we need this prop to exist on the button itself.

src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx
eszthoff: do you mean: `onSelect?: (value: string | number | boolean) => void | (selection: [name: string, value: string | number | boolean]) => void` where the first typing is for single select, and the second typing is for multiselect?

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
eszthoff: I double checked and the native HTML supports only string. So if we allow for other types, we would need to convert them for rendering and then reconvert them. The reconverting will be quite tricky as we have to guess what it was sent as. I think this is too much risk to add this flexibility to a UI component, and would rather keep this logic in the application.

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
carlobernardini: Can't it perhaps be passed on through `...rest`? Just like we do in other components, without having validation? This way they're not documented in Storybook as "public" props.

eszthoff: It is not about how it is passed, it is about how it is received.

carlobernardini: I understand that, but how come we need to have these props be explicitly defined (while they are not meant to be applied on this level by the user) while in other components we use `rest` to apply any other props without having such explicit definitions or validation?

I.e., can you explain why we cannot apply those props that are passed down from the parent using `rest`, without defining them explicitly on child level (again, since they're not part of the actual component API)?

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
mukiienko: Now `SelectButton` component tied for some particular ticket, it should be more flexible and reusable. 
And OneUI it is open source and anyone can use it  :), thus not cleat why others Context values are not supported.

src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx
mukiienko: The `name` for selection will be the same, correct?
Originally the idea was implemented  kind of `select/option` that supports `multiply`
https://mdbootstrap.com/docs/jquery/forms/multiselect/
So maybe better:
```
onChange?: ({ name: string, values:  [] }) => void
```

mukiienko: For `radio` it will be the mostly the same
```
onChange?: ({ name: string, value:  string }) => void
```

mukiienko: but I think maybe better pass arguments, not object:
```
 onChange?: (value: [string] |  string, name: string) => void
```
Then in external callback function we can use any variable names:
```
handleChange(selectedValues) {

}

<SelectButtonGroup
...
onChange={handleChange}
/>

File: src/components/SelectButtonGroup/SelectButton/SelectButton.tsx (hunk 5 above)
Comment (mukiienko) [reply src/components/SelectButtonGroup/SelectButton/SelectButton.tsx]: I think we don't need to use `checkbox`/`radio`. It can be:
```
<select  multiple>
  <option selected>Open this select menu</option>
  <option value="1">One</option>
  <option value="2">Two</option>
  <option value="3">Three</option>
</select>

```
this cover two scenarios and we can rid of `isMultiselect` for item

File: src/components/SelectButtonGroup/SelectButton/SelectButton.tsx (hunk 4 above)
Comment (carlobernardini) [reply src/components/SelectButtonGroup/SelectButton/SelectButton.tsx]: If the effort is the same it's not a problem to support more contexts, provided that it does at least include the ones required by the ticket and those match the designs. If it's unclear how other contexts are to be styled due to lack of designs, then let's not implement them.

Comment (eszthoff) [reply src/components/SelectButtonGroup/SelectButton/SelectButton.tsx]: ok, I will look at it specifically when it comes to css. If it is the same for all contexts, I will add them, and if designs are not clear, I will leave only the ones defined

File: src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx (hunk 6 above)
Comment (eszthoff) [reply src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx]: `name` is the same for all of them if they are radio buttons but not if they are checkboxes. Hence my suggestion to send it back in a similar format to the native elements. Your example relates to a select box, not a group of checkboxes. 

Native behaviour for radio is: 
`name=value` where `name` is the same for all radio buttons in the group and `value` is unique for each individual button.
So here I suggest to return `onSelect: (value: string) => void`

Native behaviour for checkbox is: 
`name=value` where `name` is unique to every checkbox on the page and `value=true` if not otherwise defined. This will be returned for individual checked checkbox on the form. 
So in multiselect mode I suggest `onSelect: (selection: [name: string, value: string][]) => void`. Alternatively, to be consistent with the radio buttons we can use `onSelect: (selection: string[]) => void` which will return an array of the values of the selected options. So name is not a required prop.

File: src/components/SelectButtonGroup/SelectButton/SelectButton.tsx (hunk 5 above)
Comment (eszthoff) [reply src/components/SelectButtonGroup/SelectButton/SelectButton.tsx]: That's a very good idea!

File: src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx (hunk 6 above)
Comment (eszthoff) [reply src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx]: To sum it up, as suggested the implementation will be based on select. So the callback can have the following signature: `onChange: (selection: string | string[]) => void`. Do you agree?

Comment (mukiienko) [reply src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx]: So maybe let's add the `name` to  `SelectButtonsGroup` as well.
Take a look at the case where it will useful:
```
handleChange(value, name) {
  data[name] = value;
}

<SelectButtonsGroup name="nameOne" onChange={handleChange}>....
<SelectButtonsGroup name="nameTwo" onChange={handleChange}>....
```
In the example above we can easily distingue what exactly part of data was changed.
Otherwise, without the `name` prop we need to write a distinct handler for each `SelectButtonsGroup`:
```
<SelectButtonsGroup onChange={handleNameOneChange}>....
<SelectButtonsGroup onChange={handleNameTwoChange}>....
```

File: src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx [hunk 7]
```diff
@@ -0,0 +1,85 @@
...
+        selectedContext,
+        onChange,
+        ...rest
+    } = props;
+
+    const initiallySelectedValues: string[] = [];
+    children.forEach(child => {
+        const { value, isSelected } = child.props;
+        if (isSelected) {
+            initiallySelectedValues.push(value);
```
igorarkhipenko: can be replaced with `initiallySelectedValues = [... initiallySelectedValues, value];` to avoid using of mutating `push` method

eszthoff: Why is it wrong to use push here, when this array was created only a few lines above only for the reason of holding these values?

src/components/SelectButtonGroup/SelectButtonGroup/SelectButtonGroup.tsx
```diff
@@ -0,0 +1,85 @@
...
+        isBlock,
+        selectedContext,
+        onChange,
+        ...rest
+    } = props;
+
+    const initiallySelectedValues: string[] = [];
+    children.forEach(child => {
+        const { value, isSelected } = child.props;
+        if (isSelected) {
```
igorarkhipenko: should be renamed to `isInitiallySelected` or something

mukiienko:
Small UX/UI improvements:
* I'd in favor to add transition animations for colors and bg colors
* add  styles for  `:active`
* centralize label text

src/components/SelectButtonGroup/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        onChange,
+        isEqualWidth,
+        isBlock,
+        selectedContext,
+        isSelected,
+        ...rest
+    } = props;
+
+    const handleClick = () => {
+        onChange?.(value);
```
mukiienko: Cool 👍 . For supporting this feature do we need to setup webpack accordingly, correct? Nice to have it for external search.

eszthoff: no need to change webpack. TS loader does it if you have TS >=3.7. I did need to update prettier though, as that got confused. Hence https://github.com/textkernel/oneui/pull/217

PR #218, opened 2020-01-14
WIP: Select POC

eszthoff:
Story: [ONEUI-136](https://jira.textkernel.nl/browse/ONEUI-136)

### DIFF
--- a/src/components/Select/SelectBase.tsx
+++ b/src/components/Select/SelectBase.tsx
@@ -12,8 +12,6 @@ export interface Props extends React.HTMLAttributes<HTMLDivElement> {
     suggestions: object[];
     /** suggestionToString(suggestion) should return a string to be displayed in the UI. e.g.: suggestion => suggestion.name */
     suggestionToString: (suggestions?: object | null) => string;
-    /** trigger of the initial focus of the input field */
-    isInitiallyFocused?: boolean;
     /** to be shown when no suggestions are available */
     noSuggestionsPlaceholder: string;
     /** to be shown as clear button title */
@@ -58,7 +56,6 @@ const SelectBase: React.FC<Props> = props => {
     const {
         suggestions,
         suggestionToString,
-        isInitiallyFocused,
         clearTitle,
         showClearButton,
         noSuggestionsPlaceholder,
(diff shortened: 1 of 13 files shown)


src/components/Select/SelectBase.tsx
```diff
@@ -0,0 +1,273 @@
...
+import { Props as ListRendererProps } from './SuggestionsList';
+import styles from './SelectBase.scss';
+
+export interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** an array of objects that will be used to render the suggestions list. */
+    suggestions: object[];
+    /** suggestionToString(suggestion) should return a string to be displayed in the UI. e.g.: suggestion => suggestion.name */
+    suggestionToString: (suggestions?: object | null) => string;
+    /** trigger of the initial focus of the input field */
+    isInitiallyFocused?: boolean;
```
antipin: The idea of setting focus by a prop can lead to unexpected behavior. There can be multiple `Select` components on a page, and if each of them has `isInitiallyFocused` set to `true`, which one would get focus in the end?

src/components/Select/SelectBase.tsx
```diff
@@ -0,0 +1,273 @@
...
+import bem from '../../utils/bem';
+import FieldWrapper from '../FieldWrapper';
+import { List } from '../List';
+import { ENTER_KEY } from '../../constants';
+import { Props as ListRendererProps } from './SuggestionsList';
+import styles from './SelectBase.scss';
+
+export interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** an array of objects that will be used to render the suggestions list. */
+    suggestions: object[];
```
antipin: I think it makes sense to put some effort here and make `Props` interface a generic one. It would allow us to express type logic better and probably catch some errors in the future.

It would looks like:

```ts
interface Props<S> extends React.HTMLAttributes<HTMLDivElement> {
    /** an array of objects that will be used to render the suggestions list. */
    suggestions: S[];
    suggestionToString: (suggestions?: S | null) => string;
    onSelectionChange: (item: S) => void;
```

src/components/Select/SelectBase.tsx
eszthoff: good point. I copied this prop from Autofocus, but this can be achieved in the application using refs and so on. I will delete it.

src/components/Select/SelectBase.tsx
eszthoff: This is really good. I didn't know how to do this, hence I used this object thing

antipin: There is [an article about JSX and generics](https://mariusschulz.com/blog/passing-generics-to-jsx-elements-in-typescript) that I googled yesterday, can be also helpful

PR #222, opened 2020-01-21
refactor(Typescript): move List, ListItem, ListAction and Teaser to TS

eszthoff:
Story: [ONEUI-112](https://jira.textkernel.nl/browse/ONEUI-112)


src/components/List/List.tsx
```diff
@@ -1,27 +1,38 @@
...
 const NAVIGATION_STEP_VALUES = {
     [LIST_NAVIGATION_DIRECTIONS.UP]: -1,
     [LIST_NAVIGATION_DIRECTIONS.DOWN]: 1,
 };
 
-export const NOT_LIST_CHILD = 'data-list-exception';
-export const LIST_CHILD = 'data-list-child';
-
-const List = React.forwardRef((props, ref) => {
+const List: React.FC<Props> = React.forwardRef((props, ref) => {
```
antipin: `React.forwardRef<T, P = {}>` is a generic function which accepts 2 arguments:

* T is type of our DOM element
* P is type of our props

That gives a better type of `ref.current` value in the end.

So here it can be:
```ts
const List: React.FC<Props> = React.forwardRef<HTMLUListElement, Props>((props, ref) => {
```

src/components/List/ListActions/ListActions.tsx
```diff
@@ -1,11 +1,15 @@
...
+interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** Actions to be pushed to the left side of a List Item */
+    children?: ReactNode;
+}
+
 const { block } = bem('ListActions', styles);
 
-const ListActions = props => {
+const ListActions: React.FC<Props> = props => {
     const { children, ...rest } = props;
```
antipin: Shouldn't we add default property for `children` here as long as we removed it from `defaultProps`?

Like:
```
const { children = null, ...rest } = props;
```

antipin: Just as a side note, the whole component can be more "compact".

```tsx

const ListActions: React.FC<Props> = ({ children, ...rest }) => (
    <div {...rest} {...block(rest)}>
        {children}
    </div>
);
```

eszthoff: What would be the reason to define children as `null` instead of undefined?

src/components/List/List.tsx
eszthoff: I think we tried to use that standard, but had some problems with it. Hence we went with defining the `ref` type in the interface. This approach also has the advantage that it shows up in storybook, so it becomes more transparent what the component expects and accepts

src/components/List/ListActions/ListActions.tsx
antipin: Yeah, should be  `undefined` :)

src/components/List/List.tsx
antipin: ok then

PR #224, opened 2020-01-22
refactor(eslint): remove default exports

igorarkhipenko:
Story [ONEUI-134](https://jira.textkernel.nl/browse/ONEUI-134)


carlobernardini:
@igorarkhipenko Is the (main) scope of this PR really about eslint? If the change doesn't have one specific scope you could also only specify a type.

README.md
```diff
@@ -38,7 +38,7 @@ To enable support for older browsers that don’t support CSS variables natively
 To apply a specific theme on top of the library components, the OneUI utility can be used to parse the provided theme file:
 
 ```javascript
-import OneUI from '@textkernel/oneui';
+import { OneUI } from '@textkernel/oneui';
```
eszthoff: I wonder if this should not be an exception to this rule. If we change this export it would be a breaking change, and also it would need an other name if we want to be named export.

src/components/Alert/Alert.js
```diff
@@ -1,13 +1,13 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import bem from '../../utils/bem';
+import { bem } from '../../utils/bem';
```
eszthoff: if we change this to named export, maybe we could also create an `index.ts` file and require it from `../utils`

src/components/List/List.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import bem from '../../utils/bem';
-import styles from './List.scss';
+import { bem } from '../../utils/bem';
 import { LIST_NAVIGATION_DIRECTIONS, ENTER_KEY } from '../../constants';
-import scrollIntoViewIfNeeded from '../../utils/scrollIntoViewIfNeeded';
+import { scrollIntoViewIfNeeded } from '../../utils/scrollIntoViewIfNeeded';
```
eszthoff: maybe we should create a `index.ts` in utils and export this one from there as well

src/components/LocationAutocomplete/LocationAutocomplete.js
```diff
@@ -1,12 +1,12 @@
...
-import bem from '../../utils/bem';
+import { bem } from '../../utils/bem';
 import { Autosuggest } from '../Autosuggest';
-import Alert from '../Alert';
-import ListItem from '../List/ListItem';
+import { Alert } from '../Alert';
+import { ListItem } from '../List/ListItem';
 import { MarkedText } from '../Text';
-import useDebounce from '../../hooks/useDebounce';
+import { useDebounce } from '../../hooks/useDebounce';
```
eszthoff: shall we add a `index.ts` to hooks too?

README.md
carlobernardini: I'm inclined to agree with that

PR #226, opened 2020-01-28
feat(ComboboxMulti): Add new component

igorarkhipenko:


src/components/Select/ComboboxMulti/ComboboxMulti.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const inputRef = inputRefFromProps || React.createRef<HTMLInputElement>();
+
+    // eslint-disable-next-line react/display-name
+    const renderFocused = ({ getInputProps, getToggleButtonProps, onBlur: blur }) => {
+        const handleInputKeyDown = event => {
+            if (event.key === TAB_KEY) {
+                inputRef.current?.blur();
+                blur();
+            } else if (event.key === ESCAPE_KEY) {
+                // prevents key propagation and sets the focus on parent component
```
eszthoff: I'm not sure if setting the focus on the parent is correct behaviour here. The focus would still be in the component, so it would still be open. Can you clarify the requirements about it?

src/components/Select/ComboboxMulti/ComboboxMulti.tsx
```diff
@@ -0,0 +1,112 @@
...
+                />
+                <div {...elem('buttons', { ...props })}>
+                    <IoIosArrowUp {...getToggleButtonProps()} />
+                </div>
+            </div>
+        );
+    };
+
+    // eslint-disable-next-line react/display-name
+    const renderBlurred = ({ getInputProps, getToggleButtonProps }) => (
```
eszthoff: These functions look almost the same. I wonder if it couldn't be simplified into 1 function with some params (mainly for which arrow is displayed) instead. But maybe that would mess up the keyboard things..

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
+import React from 'react';
+import toJson from 'enzyme-to-json';
+import { ComboboxMulti } from '../ComboboxMulti';
+import { SUGGESTIONS, SUGGESTION_TO_STRING } from '../../../Autosuggest/__mocks__/suggestions';
+
+describe('ComboboxMulti', () => {
+    const suggestionToString = SUGGESTION_TO_STRING;
+    const inputPlaceholder = 'type here...';
+    const noSuggestionsPlaceholder = 'No suggestions...';
+    const clearTitle = 'Clear';
```
eszthoff: ?

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
...
+    const mockOnSelectionChange = jest.fn();
+    const mockOnInputValueChange = jest.fn();
+    const mockOnClearAllSelected = jest.fn();
```
eszthoff: ?

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
...
+        );
+        inputNode = wrapper.find('input').getDOMNode();
+    });
+
+    describe('rendering', () => {
+        it('should initially render empty component correctly', () => {
+            expect(toJson(wrapper)).toMatchSnapshot();
+        });
+        it('should set focus on the input field', () => {
+            wrapper.setProps({ suggestions: [] });
```
eszthoff: why are you deleting the suggestions here?

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
...
+            setFocusOnInput();
+            expect(wrapper.find('li')).toHaveLength(0);
```
eszthoff: how does this relate to the subject of this test?

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
...
+            expect(inputNode).toBe(document.activeElement);
+        });
+        it('should render noSuggestions placeholder when empty suggestions list is passed', () => {
+            wrapper.setProps({ suggestions: [] });
+            wrapper.find('input').simulate('change', { target: { value: 'driver' } });
```
eszthoff: is this step really needed here?

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
...
+
+            expect(inputNode).not.toBe(document.activeElement);
+            expect(focusSpy).not.toHaveBeenCalled();
+
+            setFocusOnInput();
+
+            expect(inputNode).toBe(document.activeElement);
+            expect(focusSpy).toHaveBeenCalled();
+        });
+        it('should stay focused when suggestion is selected', () => {
```
eszthoff: is this really a requirement? It sounds confusing that you click on something and next think you are focused somewhere else, no?

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -0,0 +1,181 @@
...
+            expect(mockOnBlur).toHaveBeenCalled();
+        });
+    });
+    describe('callbacks', () => {
+        describe('onSelectionChange', () => {
+            it('should be called on clicking on a suggestion', () => {
+                setFocusOnInput();
+
+                expect(mockOnSelectionChange).not.toHaveBeenCalled();
+                expect(toJson(wrapper)).toMatchSnapshot();
```
eszthoff: maybe we don't really need a snapshot here.

src/components/Select/SelectBase/SelectBase.tsx
```diff
@@ -0,0 +1,268 @@
...
+    };
+
+    const handleClearSelectedSuggestions = e => {
+        e.stopPropagation();
+        onClearAllSelected?.();
+    };
+
+    const focus = openMenu => {
+        openMenu();
+        inputRef.current?.focus();
```
eszthoff: maybe need to set the focused state as well. If you do that, there is no need to set the input in focus here, the useEffect will take care of that

stories/Combobox.tsx
```diff
@@ -0,0 +1,65 @@
...
+type TSuggestion = { name: string };
+
+storiesOf('Organisms|Select', module)
+    .addDecorator(withKnobs)
+    .addParameters(
+        StoreInjector.withStore({
+            inputValue: '',
+        })
+    )
+    .add('Simple select', ({ parameters }) => {
```
carlobernardini: Shouldn't this story be called 'ComboboxMulti'?

src/components/Select/ComboboxMulti/ComboboxMulti.tsx
```diff
@@ -0,0 +1,112 @@
...
+            <div tabIndex={0} role="searchbox" {...elem('wrapper', { ...props })}>
+                <input
+                    {...getInputProps({
+                        ref: inputRef,
+                        placeholder: inputPlaceholder,
+                        onKeyDown: handleInputKeyDown,
+                        ...elem('input', { ...props }),
+                    })}
+                />
+                <div {...elem('buttons', { ...props })}>
```
carlobernardini: This class (`.ComboboxMulti__buttons`) doesn't seem to have a definition in the component stylesheet, so it only renders a simple div that has no styles. Is it necessary?

src/components/Select/ComboboxMulti/ComboboxMulti.tsx
igorarkhipenko: @eszthoff once the focus was removed from the input, the app doesn't handle events like key press anymore until you set it somewhere by your own. Just to make it more convenient I decided to move the focus to the parent element. For now it's used only in Autosuggest of the LocationSelector component. The issue behind it was impossibility of closing the modal window by pressing on escape after closing the suggestions dropdown was closed by the same escape press.

src/components/Select/ComboboxMulti/ComboboxMulti.tsx
igorarkhipenko: @eszthoff I'm not sure if that refactoring would be a good thing there. Since those are two separate render functions maybe it would be more convenient to keep them as they are in terms of readability?

stories/Combobox.tsx
```diff
@@ -0,0 +1,65 @@
...
+        return (
+            <ComboboxMulti<TSuggestion>
+                style={{ width: '650px' }}
+                inputPlaceholder={text('Input placeholder', 'Select something...')}
+                noSuggestionsPlaceholder={text('No suggestions', 'No suggestions found...')}
+                suggestions={getSuggestions()}
+                suggestionToString={SUGGESTION_TO_STRING}
+                onBlur={onBlur}
+                onSelectionChange={onSelectionChange}
+                onClearAllSelected={onClearAllSelected}
```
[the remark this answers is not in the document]
igorarkhipenko: by clicking on 'Clear' button of SelectBase component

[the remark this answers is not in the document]
eszthoff: Combobox should not enable the SelectBase clear button. Actually it has no state, so it has no clear option at all, as per current design

src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
igorarkhipenko: I checked it with Carlo, it is a requirement

src/components/Select/SelectBase/interfaces.ts
```diff
@@ -0,0 +1,48 @@
...
+    /** function to be called if FieldWrapper clear button is clicked */
+    onClearAllSelected?: () => void;
+    /** show Clear button on hover even if there are no selectedSuggestions passed */
+    showClearButton?: boolean;
+}
+
+export interface Props<S> extends CommonPropsWithClear<S> {
+    keepExpandedAfterSelection?: boolean;
+    /** will be called when list of suggestions should be rendered */
+    listRenderer: (props: Omit<ListRendererProps<S>, 'noSuggestionsPlaceholder'>) => ReactNode;
```
eszthoff: I think it is better to list the props that will be sent here, rather then import them from listRenderer. ListRenderer is one possible implementation and we could have other ones in the future. This base component is the bases for other things, so it should not depend on them

src/components/Select/SuggestionsList/__tests__/SuggestionsList.spec.js
```diff
@@ -0,0 +1,48 @@
...
+            suggestionToString={SUGGESTION_TO_STRING}
+            noSuggestionsPlaceholder={noSuggestionsPlaceholder}
+            getItemProps={getItemProps}
+            highlightedIndex={highlightedIndex}
+        />
+    );
+});
+
+describe('SuggestionsList', () => {
+    it('should initially render empty component correctly', () => {
```
eszthoff: this is not an empty component

src/components/Select/SuggestionsList/__tests__/SuggestionsList.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+    it('should render empty component correctly when focused', () => {
```
eszthoff: what do you mean by focused here?

src/components/Select/SuggestionsList/__tests__/SuggestionsList.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        wrapper.setProps({ suggestions: [] });
+        expect(wrapper.find('li')).toHaveLength(0);
+    });
+    it('should render noSuggestions placeholder when empty suggestions list is passed', () => {
+        wrapper.setProps({ suggestions: [], inputValue: 'test' });
```
eszthoff: why are you setting `inputValue` prop?

src/components/Select/SuggestionsList/__tests__/SuggestionsList.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        expect(toJson(wrapper)).toMatchSnapshot();
+        expect(wrapper.find('li')).toHaveLength(1);
+        expect(
+            wrapper
+                .find('li')
+                .childAt(0)
+                .text()
+        ).toEqual(noSuggestionsPlaceholder);
+    });
+    it('should render all suggestions from the list', () => {
```
eszthoff: is this different from the first test?

src/components/Select/SelectBase/__tests__/SelectBase.spec.js
```diff
@@ -0,0 +1,162 @@
...
+        });
+        // Despite everything is working good in real-case scenario,
+        // 'blur' event can not be simulated in the way Downshift component to understand it.
+        it.skip('should call onBlur', () => {
+            setFocusOnInput();
+            wrapper.find('SelectBase').simulate('blur');
+
+            expect(mockOnBlur).toHaveBeenCalled();
+        });
+        it('should stay focused when suggestion is selected with keepExpandedAfterSelection set to true', () => {
```
eszthoff: this test seems to belong to the previous describe

PR #227, opened 2020-01-29
fix(FieldWithValidation): change logic when using tooltip

eszthoff:
Story: [ONEUI-146](https://jira.textkernel.nl/browse/ONEUI-146)

Changes to behaviour:
* always render tooltip in FieldWithValidation when `useTooltip` is set to true
* always show tooltip if underlying component is focused
* allow `null` as result of renderPopup() in PopupBase
* do not render tooltip popup if it has no content
* add 'always visible' option to Tooltip

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -21,6 +21,14 @@ export const Tooltip: React.FC<Props> = props => {
     // eslint-disable-next-line react/prop-types
     const { placement, content, children, alwaysVisible, ...rest } = props;
 
+    let openCall: (isOpen: boolean) => void = () => null;
+
+    React.useEffect(() => {
+        if (alwaysVisible) {
+            openCall(!!content);
+        }
+    }, [content, alwaysVisible]);
+
     const createMouseOverHandler = setPopupVisibility => () => {
         setPopupVisibility(true);
     };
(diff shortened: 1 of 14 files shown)


stories/FieldWithValidation.tsx
```diff
@@ -3,6 +3,34 @@ import { storiesOf } from '@storybook/react';
...
+                    onChange={e => {
+                        const { value } = e.target;
+                        setValue(value);
+                    }}
+                />
+            </FieldWithValidation>
+        </div>
+    );
+};
+
```
antipin: Found couple more weird things in this example, sent you the demo in slack.

antipin: Found couple more weird things in this example, sent you the demo in slack.

carlobernardini:
@eszthoff there is a typo in the PR title / scope

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { PopupPlacement } from '../../constants';
 import { PopupBase } from '../PopupBase';
+import { TooltipAnchor } from './TooltipAnchor';
```
antipin: Nice extraction!

eszthoff: haha. thanks. It was really hard to read when they were in one file with `React.UseEffect` needing access to `setPopupVisibility`.

PR #230, opened 2020-02-07
build(deps): migrate popper.js to v2

igorarkhipenko:


src/components/Tooltip/Tooltip.scss
```diff
@@ -12,11 +12,11 @@
         position: absolute;
     }
 
-    [x-placement^="top"] & {
+    [data-popper-placement^="top"] & {
         margin-bottom: var(--spacing-5x);
```
eszthoff: shouldn't all the margins be removed? https://popper.js.org/docs/v2/migration-guide/#4-remove-all-css-margins

src/__mocks__/popper.js.js
```diff
@@ -1,8 +1,8 @@
-import PopperJs from 'popper.js';
+import { placements } from '@popperjs/core';
 
 // eslint-disable-next-line import/no-default-export
 export default class Popper {
-    static placements = PopperJs.placements;
+    static placements = placements;
```
eszthoff: this looks a bit redundant. Also tests seem to pass without it.

src/components/Tooltip/Tooltip.scss
igorarkhipenko: @eszthoff I removed most of the margins, but I left positioning ones like this one. Popper.js proposal is to replace them with `offset` props which is not supported by IE and Edge. Since those margins are working fine and don't make any sort of conflicts and overwriting maybe it makes sense to leave them as it is?

PR #233, opened 2020-02-19
WIP: feat(Tooltip): Migrate component to tippy.js library

igorarkhipenko:


stories/Tooltip.tsx
```diff
@@ -6,14 +6,29 @@ import { POPUP_PLACEMENTS } from '../src/constants';
...
+    .add(
+        'Tooltip',
+        () => (
+            <div style={{ position: 'relative', display: 'table', margin: '0 auto' }}>
+                <Tooltip
+                    placement={select('Placement', POPUP_PLACEMENTS, 'bottom')}
+                    content={text('Tooltip text', 'this is my tooltip text')}
+                    arrow={boolean('Arrow', true)}
+                >
+                    <div>hover here to see the tooltip</div>
```
antipin: Does that mean that the children always have to be a node? Can we use plain text?

scripts/build/utils/utils.js
```diff
@@ -18,7 +18,7 @@ function getRuleTS({ includePaths }) {
 
 function getRuleCSS({ styleLoader, localIdentName, includePaths, context }) {
     return {
-        test: /\.scss$/,
+        test: /\.s?css$/,
```
antipin: It's not related to this particular line of code, but when I run `npm run storybook`, I see this error:
image

stories/Tooltip.tsx
igorarkhipenko: @antipin yes, according to tippy.js documentation, a child of the component should have `React.ReactElement<any>` type

PR #234, opened 2020-02-24
feat(Input): Add an optional label

igorarkhipenko:
Story [ONEUI-165](https://jira.textkernel.nl/browse/ONEUI-165)

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { bem } from '../../utils';
+import { Text } from '../Text';
 import styles from './Input.scss';
 import { ValidationContext, InputType, Size } from '../../constants';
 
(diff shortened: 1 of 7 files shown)


src/components/Input/Input.tsx
```diff
@@ -20,19 +25,35 @@ interface Props extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'
...
-            {...rest}
-            {...block(props)}
-            ref={ref}
-            type={type}
-            disabled={disabled}
-            value={value}
-        />
+        <div>
+            {labelText && (
+                <Text {...elem('label', { elemClassName: labelClassName })}>{labelText}</Text>
```
mukiienko: Tell me please why we need `labelClassName`?

igorarkhipenko: @mukiienko to have an option to change label style

mukiienko: cool, I don't know that the BEM library has this option. Then I don't have any comments

PR #239, opened 2020-02-28
feat(component): listoptimizer component

mukiienko:


src/components/List/ListOptimizer/ListOptimizer.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { List, CellMeasurer, CellMeasurerCache, AutoSizer, ListRowProps } from 'react-virtualized';
+
+export interface Props {
+    rowCount: number;
+    defaultHeight?: number;
+    defaultWidth?: number;
+    children: (props: ListRowProps) => React.ReactNode;
+}
+
+const DEFAULT_HEIGHT = 30;
```
igorarkhipenko: maybe it makes sense to rename this constant to `DEFAULT_CELL_HEIGHT` or something? Current naming in combination with `defaultHeight` prop of the component below makes it a bit ambiguous.

mukiienko: Yep, good point

mukiienko: Good point

src/components/Select/SuggestionsList/SuggestionsList.tsx
```diff
@@ -36,26 +39,42 @@ export function SuggestionsList<S>(props: Props<S>) {
...
-                    key={suggestionToString(item)}
-                    {...getItemProps({
-                        item,
-                        index,
+            {useOptimizeRender ? (
+                <ListOptimizer rowCount={suggestions.length}>{renderItem}</ListOptimizer>
+            ) : (
+                <>
+                    {suggestions.map((item, index) => {
+                        const key = suggestionToString(suggestions[index]);
```
igorarkhipenko: can be replaced with `const key = suggestionToString(item);`

PR #241, opened 2020-03-24
feat(component): added new AutosuggestMulti component

mukiienko:


carlobernardini:
When I add two identical items, the item that is already selected gets removed (instead I would expect the new entry to be ignored?)

Also, the input sometimes shifts to a new row while there is still quite some space left on the first (next to selected tags). Can we improve that?

Screenshot 2020-03-24 at 17 15 13

Last, can we use normal font size and slightly decrease the spacing on the left and right of tags (by factor of 1 should be sufficient)?

carlobernardini:
There is a little alignment issue with tags that appear on subsequent rows:

Screenshot 2020-03-25 at 16 27 26

Also the label in the tags in subsequent rows seems slightly misaligned (vertically) with respect to the cross next to them, but maybe that's just me

src/components/Select/AutosuggestMulti/SuggestionTag/SuggestionTag.tsx
```diff
@@ -0,0 +1,36 @@
...
+    children: NotEmptyReactNode;
+};
+
+export const SuggestionTag: React.FC<Props> = (props) => {
+    const { children, isBounded, isStretched, onClick, ...rest } = props;
+
+    return (
+        <div {...rest} {...block({ isBounded, isStretched })}>
+            <span title={children} {...elem('label')}>
+                {children}
```
eszthoff: with the current interface `children` could be a `div` or anything else, which would result in incorrect HTML. Maybe make change `span` to `div`

src/components/Select/AutosuggestMulti/SuggestionTag/__tests__/SuggestionTag.js
```diff
@@ -0,0 +1,35 @@
...
+describe('SuggestionTag', () => {
+    it('should render correctly', () => {
+        const wrapper = mount(<SuggestionTag>tag</SuggestionTag>);
+
+        expect(toJson(wrapper)).toMatchSnapshot();
+        expect(wrapper.find('button')).toHaveLength(0);
+    });
+    it('should render correctly with styles modifier', () => {
+        const wrapper = mount(
+            <SuggestionTag isBounded isStretched>
```
eszthoff: does it actually make sense to have both this props? I feel this makes the API confusing and you can easily misuse it, no?

src/components/Select/AutosuggestMulti/SuggestionTag/SuggestionTag.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { bem } from '../../../../utils';
+import styles from './SuggestionTag.scss';
+
+const { block, elem } = bem('SuggestionTag', styles);
+
+type Props = {
+    /** Tag Stretches on full width */
+    isStretched?: boolean;
+    /** Tag has restricted width */
+    isBounded?: boolean;
```
eszthoff:
```suggestion
    isStretched?: boolean;
    isBounded?: boolean;
```

these should have a default value of `false` as usual

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
```diff
@@ -0,0 +1,161 @@
...
+        wrapper.unmount();
+    });
+
+    describe('rendering', () => {
+        it('should initially render empty component correctly', () => {
+            expect(toJson(wrapper)).toMatchSnapshot();
+        });
+        it('should render empty component correctly when focused', async () => {
+            setFocusOnAutosuggest();
+            expect(toJson(wrapper)).toMatchSnapshot();
```
eszthoff: these test can easily be wrong / snapshot changes. It would make sense to add an assertion that actually checks if the component is correctly focused. something like:
```
const inputNode = wrapper.find('input').getDOMNode();
expect(document.activeElement).not.toBe(inputNode);
```

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+        it('should render tag for each selected selection when component is focused', () => {
```
eszthoff: should have a similar test for when component is blurred?

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
```diff
@@ -0,0 +1,161 @@
...
+            it('should be called on clicking on a suggestion', () => {
+                suggestionsList = SUGGESTIONS.slice(1, 20);
+                wrapper.setProps({ suggestions: suggestionsList });
+                setFocusOnAutosuggest();
+
+                expect(mockOnSelectionChange).not.toHaveBeenCalled();
+
+                wrapper.find('input').simulate('change', { target: { value: 'a' } });
+
+                expect(toJson(wrapper)).toMatchSnapshot();
```
eszthoff: why do we have this snapshot test here?

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+        it('should blur on pressing Tab button', () => {
+            setFocusOnAutosuggest();
+            wrapper.find('input').simulate('keyDown', { key: 'Tab' });
+            expect(mockOnBlur).toHaveBeenCalled();
+        });
+    });
+    describe('callbacks', () => {
+        describe('onSelectionChange', () => {
+            it('should be called on clicking on a suggestion', () => {
```
eszthoff: I think this and the next test are covered in SelectBase. Maybe it is not needed here, what do you think?

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
```diff
@@ -0,0 +1,161 @@
...
+                wrapper.find('input').simulate('change', { target: { value: 'driver' } });
+
+                expect(mockOnSelectionChange).not.toHaveBeenCalled();
+
+                wrapper.find('input').simulate('keyDown', { key: 'Backspace' });
+
+                expect(mockOnSelectionChange).not.toHaveBeenCalled();
+            });
+        });
+        it('should call onInputValueChange when typing into input field', () => {
```
eszthoff: this and the next test are also in SelectBase, maybe we don't need to repeat them here, what do you think?

src/components/Select/SelectBase/SelectBase.tsx
```diff
@@ -86,6 +88,12 @@ export function SelectBase<S>(props: Props<S>) {
             onSelectionChange(selectedItem);
         }
 
+        if (clearInputAfterSelection) {
+            setInputValue('');
+            setInputValueRecall('');
```
eszthoff: `inputValueRecall` should not change, so that highlighting in list is still visible.

src/components/Select/SuggestionsList/SuggestionsList.tsx
```diff
@@ -15,7 +15,7 @@ export interface Props<S> {
     /** suggestionToString(suggestion) should return a string to be displayed in the UI. e.g.: suggestion => suggestion.name */
     suggestionToString: (suggestions: S) => string;
     /** to be shown when no suggestions are available */
-    noSuggestionsPlaceholder: string;
+    noSuggestionsPlaceholder?: string;
```
eszthoff: Why is this? (See CONTRIBUTE.md 'Type definitions for Props and States' section)

src/components/Text/MarkedText/MarkedText.js
```diff
@@ -11,7 +11,8 @@ export const MarkedText = (props) => {
 
     let result = children;
     if (marker) {
-        const re = new RegExp(`(${marker})`, 'gi');
+        const escapedMarker = marker.replace(/[-[\]{}()*+?.,\\^$|#\\s]/g, '\\$&');
```
eszthoff: nice one!

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
```diff
@@ -0,0 +1,150 @@
...
+            inputRef={inputRef}
+            onBlur={onBlur}
+            onSelectionChange={onSelectionChange}
+            onInputValueChange={handleInputValueChange}
+            listRenderer={(listProps) => (
+                <SuggestionsList {...listProps} useOptimizeRender={useOptimizeListRender} />
+            )}
+            focusedRenderer={renderFocused}
+            blurredRenderer={renderBlurred}
+            keepExpandedAfterSelection
```
eszthoff: When I add suggestions, the list collapses for me. Is that intended? Seems to be in contrast with this prop.

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
```diff
@@ -0,0 +1,150 @@
...
+import {
+    SelectBase,
+    CommonProps,
+    FocusedRendererHelpers,
+    BlurredRendererHelpers,
+} from '../SelectBase';
+import styles from './AutosuggestMulti.scss';
+import { BACKSPACE_KEY, ESCAPE_KEY, TAB_KEY } from '../../../constants';
+
+interface Props<S> extends CommonProps<S> {
```
eszthoff: shouldn't be able to have a `clearButton` as well? In the example implementation I cannot delete all selected items at the moment.

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
```diff
@@ -0,0 +1,150 @@
...
+    /** array of already selected suggestions */
+    selectedSuggestions: S[];
+    /** number of shown tags in blur mode */
+    numberOfShownTags?: number;
```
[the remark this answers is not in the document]
mukiienko: Done

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
```diff
@@ -0,0 +1,150 @@
...
+        } else if (event.key === ESCAPE_KEY) {
+            // prevents key propagation and sets the focus on parent component
+            inputRef.current?.blur();
+            blur();
+            inputRef.current?.parentElement?.focus();
+            event.stopPropagation();
+        } else if (
+            event.key === BACKSPACE_KEY &&
+            !inputValue &&
+            selectedSuggestions &&
```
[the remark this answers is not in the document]
mukiienko: yes, we don't need it. Deleted

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
```diff
@@ -0,0 +1,150 @@
...
+export function AutosuggestMulti<S>(props: Props<S>) {
+    const {
+        onInputValueChange,
+        onSelectionChange,
+        selectedSuggestions,
+        suggestionToString,
+        inputPlaceholder,
+        useOptimizeListRender,
+        suggestions,
+        numberOfShownTags = 3,
```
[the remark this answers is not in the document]
mukiienko: Done

src/components/Select/AutosuggestMulti/SuggestionTag/SuggestionTag.tsx
mukiienko: yes, it makes sense to use `div`

src/components/Select/AutosuggestMulti/SuggestionTag/__tests__/SuggestionTag.js
mukiienko: already discussed in Slack

src/components/Select/SelectBase/SelectBase.tsx
mukiienko: Deleted it from here

src/components/Select/SuggestionsList/SuggestionsList.tsx
mukiienko: discussed in Slack

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
mukiienko: The list collapsed because it doesn't have suggestions but it is still expanded. if we delete `keepExpandedAfterSelection` then after selection input loses the focus.

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
mukiienko: in current UI we don't have `clearButton`  and it is not clear where can place it.

eszthoff: In the current Autosuggest it shows up if you hover over it when it is closed. I think that should be ok here too, unless @carlobernardini  thinks otherwise. It is actually already implemented only have to add this prop to pass the text for the button.

src/components/Select/AutosuggestMulti/SuggestionTag/__tests__/SuggestionTag.js
```diff
@@ -10,11 +10,7 @@ describe('SuggestionTag', () => {
         expect(wrapper.find('button')).toHaveLength(0);
     });
     it('should render correctly with styles modifier', () => {
-        const wrapper = mount(
-            <SuggestionTag isBounded isStretched>
-                tag
-            </SuggestionTag>
-        );
+        const wrapper = mount(<SuggestionTag width="block">tag</SuggestionTag>);
```
eszthoff: these tests are not executed. Need to change the file name to include `spec`

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
eszthoff: I see that this clearing of input only happens in the example implementation. But that is different from current behaviour. Maybe better to show it without suggestions being cleared.

stories/Select.tsx
```diff
@@ -68,4 +73,68 @@ storiesOf('Organisms|Select', module)
...
+    })
+    // eslint-disable-next-line
+    .add('AutosuggestMulti', ({ parameters }: any) => {
+        const store = parameters.getStore();
+        searchFor.name = `Search for "${store.get('inputValue')}"`;
+        searchFor.value = store.get('inputValue');
+        const getSuggestions = () => {
+            if (!store.get('inputValue').length) return [];
+            const suggestions = SUGGESTIONS.filter(
+                (item) => !store.get('selectedSuggestions').includes(item)
```
eszthoff: checking equality for `item` doesn't work, because it is an object. You need to check for `item.name`

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
mukiienko: Yes, it true. Updated

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
mukiienko: Good point. I will add.

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
mukiienko: I will delete it

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
mukiienko: I would like to keep it. Now it looks redundant, but I think we will extend `AutosuggestMulti` and can break `onSelectionChange` in the future.

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
mukiienko: I think better to keep it.
https://github.com/textkernel/oneui/pull/241#discussion_r398752386

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
mukiienko: Screenshot 2020-03-26 at 18 18 27
Clear button moves tags to the left. We need to think up where to place it, because it can overlap tags.

stories/Select.tsx
mukiienko: it works :) It will return true if we have a link on the same object.
The same checks we have for 'Autosuggest' and `ComboboxMulti`

src/components/Select/AutosuggestMulti/SuggestionTag/__tests__/SuggestionTag.js
mukiienko: Done

src/components/Select/AutosuggestMulti/__tests__/AutosuggestMulti.spec.js
eszthoff: makes sense

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
eszthoff: @carlobernardini can you advise?

src/components/Select/AutosuggestMulti/SuggestionTag/__tests__/SuggestionTag.js
eszthoff: it is not in the commit. The whole file got deleted. :-(

stories/Select.tsx
eszthoff: Screenshot 2020-03-27 at 09 58 17

From the code, I thought that selected items should not be available to be reselected.

src/components/Select/AutosuggestMulti/AutosuggestMulti.tsx
carlobernardini: See Slack: if we can implement it without a lot of effort and make sure that the button does not flow out of the input, then I suggest we do it. The tags can be pushed to the left a little when the button appears on mouseover (possibly with a little transition?).

PR #249, opened 2020-04-07
feat(component): added HighlighterCore and StringHighlighter

mukiienko:


src/utils/HighlighterCore/HighlighterCore.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    static buildRegExp({ searchString, accuracy, ignoreCase }: RegExpBuilder) {
+        let source: string | RegExp = '';
+        let flags = 'gm';
+
+        switch (accuracy) {
+            case 'partial':
+                source = `(${searchString})`;
+                break;
+            case 'exact':
```
igorarkhipenko, line 69: using of this `case` can be omitted since it falls to `default` keyword

stories/Highlighter.tsx
```diff
@@ -0,0 +1,31 @@
...
+            .filter((i) => i !== '');
+        return (
+            <StringHighlighter
+                string={someText}
+                searchTerms={keywords}
+                accuracy={select('Accuracy', ACCURACY, ACCURACY[0]) as HighlighterCoreAccuracy}
+                ignoreCase={boolean('ignoreCase', true)}
+                ignoreDiacritics={boolean('ignoreDiacritics', true)}
+                highlightRenderer={({ substring, ...substringProps }) => (
+                    <b {...substringProps} style={{ backgroundColor: 'lightyellow' }}>
```
igorarkhipenko, line 25: 1) Shall we use `bem` instead of inline styling in this case?
2) Should we lave an option to change the color of the highlighted background?

stories/Highlighter.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import { storiesOf } from '@storybook/react';
+import { boolean, select, text, withKnobs } from '@storybook/addon-knobs';
+import { StringHighlighter, HighlighterCoreAccuracy } from '@textkernel/oneui';
+
+storiesOf('Atoms|Highlighter', module)
```
igorarkhipenko, line 6: A very minor one but it doesn't look like an atom to me, the logic and the structure is pretty complex so I'd move it at least to Molecules 😄

src/utils/HighlighterCore/HighlighterCore.ts
mukiienko: I would like to keep  `exact` because we can explicitly see which code applies for each `accuracy`. We can skip default, but it is not a good practice :)

stories/Highlighter.tsx
mukiienko: The `highlightRender` method is exposed outside, so we can pass there whatever we want: react components with BEN or any HTML tags.

src/index.ts
```diff
@@ -1,12 +1,18 @@
 // eslint-disable-next-line import/no-default-export
 export { default } from './utils/OneUI';
+export {
+    HighlighterCore,
+    HighlighterCoreOptions,
+    HighlighterCoreAccuracy,
+} from './utils/HighlighterCore';
```
antipin, line 7: do we really need to export it as a part of OneUI public functionality?

stories/Highlighter.tsx
mukiienko: I think `Molecule ` is a component that made up of other components. The current component is a standalone component and that does not matter how is code big or complex.

src/index.ts
mukiienko: Why not? :) I think the method can be useful when we need to highlight or find some text, but don't need the React component for this, on some low level. Actually I think the place for 
 `HighlighterCore` is a separate library.

antipin: Well, there is this principle "Expose as little as you can", otherwise you have to support and maintain what you have exposed. But, in our case I guess it's ok.


> Actually I think the place for HighlighterCore is a separate library.
Might be then it's better to move it to  `/src/packages`?

stories/Highlighter.tsx
mukiienko: Actually it is a small component, just 46 lines :) Probably you meant `HighlighterCore`, but it is not React component.

src/index.ts
carlobernardini: I would agree with @antipin, we can always expose it once we really need it to be exposed? Unless of course you already foresee that to be the case, I can't tell

mukiienko: Cool, I will move it to`/src/packages`.
The idea was to use some `HighlighterCore` methods for Reach.
For instance, `StringHighlighter` applies some terms ['JavaScript', 'Göttingen', 'php developer']. Every time when `StringHighlighter` component is run it does some normalization for that terms: some sorting, deleting diacritics, escaping. So if that component mounted 20 times it will do the normalization for the same terms 20 times :(.
For this reason, I added support `RegExp` for `StringHighlighter`. `HighlighterCore` exposes methods that allow to create `RegExp`. Thus we can create `RegExp` outside `StringHighlighter` and then pass that `RegExp` for every `StringHighlighter` instance. It is a kind of performance improvement.

PR #251, opened 2020-04-17
WIP: fix: update AutosuggestMulti component

igorarkhipenko:
Story [ONEUI-179](https://jira.textkernel.nl/browse/ONEUI-179)


carlobernardini:
I don't think this can be considered a bugfix, but rather an enhancement (or `feat`). Also the title can be more descriptive, consider something like `feat(AutosuggestMulti): submit on enter and select on tab press`.

PR #252, opened 2020-05-01
fix: fixed z-index of autosuggest input

igorarkhipenko:
Story [https://jira.textkernel.nl/browse/ONEUI-182](ONEUI-182)

### DIFF
--- a/src/components/Autosuggest/Autosuggest.scss
+++ b/src/components/Autosuggest/Autosuggest.scss
@@ -14,6 +14,7 @@
         &--focused {
             position: absolute;
             width: 100%;
+            z-index: 450;
         }
 
         &--isProminent {


src/components/Autosuggest/Autosuggest.scss
```diff
@@ -14,6 +14,7 @@
         &--focused {
             position: absolute;
             width: 100%;
+            z-index: 450;
```
mukiienko, line 17: Do we have some conventions about using `z-index` or seems we have some task about making `z-index` rules, correct? At least use the same `z-index` that we have for `SelectBase` and specify it in a global variable. What do think?
Screenshot 2020-05-01 at 13 43 20

igorarkhipenko: @mukiienko yes, we have a convention about arranging of z-indices which has a [page](http://localhost:9001/?path=/story/concepts-arrangment-of-z-indices--arrangment-of-z-indices) in storybook. Since that component is a popup it's z-indices range is in between 400 and 500. I'm not sure if it's really necessary to move those to since these are just guidelines and we didn't have a rule of the layers separating with constants. But we can discuss it later tho.

mukiienko: I don't think the element is the popup in a non-focus state. It has normal flow (position: relative). So I don't see any reason to keep `z-index` for `Aotosuggest__main`or maybe I miss something.

igorarkhipenko: @mukiienko just checked it in JF NextGenUI and apparently removing of `z-index: 400` from main class breaks the elements layering of the ProfessionsSearch component. Z-index set to 450 it kind of fixes the bug but during the animation of the input being overlapped over another there's a cross-sectioned part which looks broken. You can check it if you want but I'd rather leave z-indices arranging here as it used to be before.

PR #253, opened 2020-05-01
feat(component): added highlighter packages

mukiienko:


src/components/Highlighter/GroupHighlighter/GroupHighlighter.tsx
```diff
@@ -0,0 +1,129 @@
...
+ */
+const SELECTED_NODE_STYLES = {
+    boxShadow: '0 0 0 1px #000',
+    position: 'relative',
+} as CSSStyleDeclaration;
+
+/**
+ * Defines styles for cleaning selection for node
+ */
+const RESET_SELECTED_NODE_STYLES: CSSStyleDeclaration = {
```
igorarkhipenko, line 53: is it necessary to duplicate `CSSStyleDeclaration` here?

src/packages/@highlighter/control/NodeIterator/NodeIterator.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * This module allows you to iterate through dom nodes
+ * Support filtering for certain types of node
+ */
+export class NodeIterator {
+    private root: HTMLElement;
+
+    private whatToShow: number | undefined;
```
igorarkhipenko, line 20: can be replaced with `private whatToShow?: number`?

src/packages/@highlighter/node/HighlighterNode.ts
```diff
@@ -0,0 +1,65 @@
...
+        });
+    }
+
+    constructor({ highlightRenderer, highlighterCore }: HighlighterNodeOptions) {
+        this.highlightRenderer = highlightRenderer;
+        this.highlighterCore = highlighterCore;
+    }
+
+    public find(node: Node, onMatch: (node: HTMLElement) => void) {
+        if (!node.textContent || !this.highlighterCore) return;
```
igorarkhipenko, line 48: Can we replace this condition block with the one below so it would minimize the number of returns and to avoid using of negative `if` condition
```
if (node.textContent || this.highlighterCore) {
   /** the rest code of the function below **/
}
```

src/components/Highlighter/GroupHighlighter/GroupHighlighter.tsx
mukiienko: I forgot to delete it, thanks

src/packages/@highlighter/control/NodeIterator/NodeIterator.ts
mukiienko: Good point!

src/packages/@highlighter/node/HighlighterNode.ts
mukiienko: yes, I replaced it.

PR #255, opened 2020-05-13
feat(LocationSelector): Added supporting of single selection, updated radius changing interactions

igorarkhipenko:


src/components/LocationCard/LocationCard.js
```diff
@@ -13,6 +13,7 @@ export const LocationCard = (props) => {
         As,
         locationId,
         locationTitle,
+        hasRadiusSlider,
         distanceRadius,
```
mukiienko, line 17: Screenshot 2020-05-18 at 13 01 14
The height of the input is not consistent with other inputs

src/components/LocationCard/LocationCard.js
```diff
@@ -13,6 +13,7 @@ export const LocationCard = (props) => {
...
         sliderLabel,
```
mukiienko, line 18: There is no way to activate the location modal window using just a keyboard. I can select the input by pressing the tab button, but I can't open the modal window.

src/components/LocationCard/LocationCard.js
```diff
@@ -13,6 +13,7 @@ export const LocationCard = (props) => {
         As,
         locationId,
         locationTitle,
+        hasRadiusSlider,
```
mukiienko: Screenshot 2020-05-18 at 13 20 13
The selected item is inside a gray block according to the task REACH-82. I don't know that issue or not. Could @carlobernardini please confirm.

src/components/LocationCard/LocationCard.js
mukiienko, line 18: We have a contract between input and label for all our inputs.
When we click on the label we have to make input active. For our case click on the label should activate the modal window. 
Actually, I don't like the idea that we use not` input` elements for `LocationSelector`. 
For my point is better to split up logic `input` (Reach) + `LocationSelector` (OneUI).

src/components/LocationSelector/LocationSelector.tsx
```diff
@@ -1,17 +1,82 @@
...
+    /** function called with location object as an argument when it is selected from the suggestions */
+    onAddLocation: (Location) => void;
+    /** function called with a location details as an argument to be changed */
+    onUpdateLocation: (Location) => void;
+    /** function with a locationId as an argument to be removed */
+    onRemoveLocation: (Location) => void;
+    /** callback function for the Clear button click */
+    onRemoveAllLocations: () => void;
+    /** callback function for closed modal */
+    onBlur: () => void;
```
mukiienko, line 76: Do you think we need to mark that PR that has breaking changes? After increasing the OneUI version in JF can that cause some `typescript` errors?

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.scss
```diff
@@ -13,6 +14,20 @@ $location-card-width: 300px;
...
+    
+    &__slider {
+        width: 50%;
+        max-width: 400px;
+        flex-direction: row;
+        display: flex;
+        margin: var(--spacing-5x) 5px var(--spacing-5x) 25px;
+    }
+
+    &__slider-label {
```
mukiienko, line 26: I think we were agreed to use camelCasing for class names - `sliderLabel`

src/components/LocationSelector/LocationSelector.tsx
```diff
@@ -77,18 +144,19 @@ export function LocationSelector(props) {
...
      */
     function handleAddLocation(location) {
-        const { Geocoder } = window.google.maps;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const { Geocoder } = (window as any).google.maps;
         const geocoder = new Geocoder();
 
         return findCenter(geocoder, location.place_id)
-            .then((center) => {
+            .then((center: any) => {
```
mukiienko, line 152: Can you use `CenterClass` instead of `any`?
https://github.com/textkernel/oneui/pull/255/files#diff-299f591a5c6c88980bc5d24d2e8736baR26

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.js
```diff
@@ -107,6 +142,8 @@ export const LocationSelectorDialog = (props) => {
 LocationSelectorDialog.displayName = 'LocationSelectorDialog';
 
 LocationSelectorDialog.propTypes = {
+    /**  */
```
mukiienko, line 145: Missed description

src/components/Map/Map/Map.js
```diff
@@ -23,11 +23,30 @@ export const Map = React.forwardRef((props, ref) => {
...
+                geocoder.geocode({ address: marker.description }, (result, status) => {
+                    if (status === 'OK') {
+                        map.fitBounds(result[0].geometry.viewport);
+                    }
+                });
+            } else {
+                bounds.extend(marker.center);
+                map.fitBounds(bounds);
+            }
+        } else if (markers.length) {
```
mukiienko, line 49: it is hard to grasp what is going on in this method `fitBounds`. We have here 12 different conditions, it is hard to read. I think it should be refactored.

igorarkhipenko: makes sense, I'll refactor it

src/components/Autosuggest/Autosuggest.js
```diff
@@ -74,22 +74,25 @@ export class Autosuggest extends React.Component {
...
-        } else {
-            openMenu();
-        }
+            if (!isMultiselect) {
+                this.inputRef.current.blur();
+                this.handleBlur();
+            } else {
+                openMenu();
+            }
+        });
```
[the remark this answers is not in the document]
igorarkhipenko: It looks weird indeed but it was the only way I managed to get a correct behavior. I'll revisit this part during the refactoring session.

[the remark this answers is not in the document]
mukiienko: Can we write?:
```
this.setState((state) => {

});

```

carlobernardini:
@igorarkhipenko @mukiienko Please only fix what was introduced with this PR, let's create a separate bugfix ticket for all the rest (if applicable)

src/components/LocationSelector/LocationSelector.tsx
igorarkhipenko: Apparently `CenterClass` is a value and cannot be used as a type here. Maybe we should check for an additional package which would add google maps types to the project but I think it would be better to do that in the next story.

image

src/components/Autosuggest/Autosuggest.js
```diff
@@ -34,7 +34,7 @@ export class Autosuggest extends React.Component {
         this.handleWrapperKeyDown = memoize(this.handleWrapperKeyDown);
```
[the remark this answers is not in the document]
igorarkhipenko: I think this component returns all of the information it fetches from google api. I'm not sure how exactly it can define a postal code for a city in this case so maybe this is a reason it's missing.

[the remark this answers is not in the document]
carlobernardini: I think that using the `(regions)` place type you can search on a postal code but you don't get postal codes returned in the predictions. In other words, a limitation on Google's side. If you do need such additional metadata, perhaps see if you can make use of the [Place Details API](https://developers.google.com/places/web-service/details).

src/components/LocationCard/LocationCard.js
carlobernardini: It's fine to keep it as is for now, let's look into this later.

src/components/Autosuggest/Autosuggest.js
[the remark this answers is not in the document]
mukiienko: I don't know how we can integrate `LocationSearch` component to Monster search if postcode is required in terms of story REACH-82. I think that point should be clarified with Ruud or Chris.
https://gis.stackexchange.com/questions/33966/google-geocoder-lookup-get-postal-code-by-country-and-city

src/components/LocationSelector/LocationSelector.tsx
igorarkhipenko: I've just converted `js` file to `tsx`. I guess we can do that but I'm not sure if it's really necessary. Current version of OneUI includes breaking changes what makes future updates of JF safe enough. Also api has not been changed and LocationSelector already has all of the required props included.

src/components/Autosuggest/Autosuggest.js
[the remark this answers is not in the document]
carlobernardini: Postal code is indeed required for Monster. But I can also understand that it doesn't make much sense to return postal codes for (large) cities as they can span multiple postal codes. Although probably not very pretty or efficient, you could look into a solution like making two requests in sequence: first get the autocomplete predictions, then get additional details for the selected place id using the API I mentioned before (assuming that will include a postal code, which may depend on the place type you request autocomplete predictions for). This could perhaps also be done outside of the LocationSelector scope / on the application side.

This needs further investigation, maybe involve Chris. It would be good if we can determine asap. if this can be solved outside of OneUI or that it is indeed related to this component / PR.

src/components/LocationSelector/LocationSelector.tsx
carlobernardini: Only mark breaking change if it was introduced with this PR and when it involves a particular change that hasn't occurred / was already marked as a breaking change before.

src/components/LocationCard/LocationCard.js
mukiienko: @carlobernardini should we add that comment to ONEUI-184?

PR #256, opened 2020-05-13
feat(component): polishing autocompletemulti

mukiienko:
BREAKING CHANGE: logic with deleting selection was moved to onSelectionRemove callback in AutosuggestMulti

### DIFF
--- a/src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
+++ b/src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
@@ -81,34 +81,46 @@ describe('ComboboxMulti', () => {
             expect(inputNode).toBe(document.activeElement);
             expect(wrapper.find('li')).not.toHaveLength(0);
         });
-        it('should blur on pressing Escape button', () => {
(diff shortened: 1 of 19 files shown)


src/components/Select/ComboboxMulti/__tests__/ComboboxMulti.spec.js
```diff
@@ -81,34 +81,46 @@ describe('ComboboxMulti', () => {
...
-            expect(inputNode).not.toBe(document.activeElement);
-            expect(blurSpy).toHaveBeenCalled();
-            expect(mockOnBlur).toHaveBeenCalled();
+            setTimeout(() => {
+                wrapper.update();
+                expect(inputNode).not.toBe(document.activeElement);
+                expect(blurSpy).toHaveBeenCalled();
+                expect(mockOnBlur).toHaveBeenCalled();
+                done();
+            }, 0);
```
igorarkhipenko, line 99: 0 as a parameter can be omitted since it's a default value for the second argument

mukiienko: done!

PR #288, opened 2020-05-28
chore(deps): reconfigure renovate

carlobernardini:
* Reduce PR noise by grouping and auto-merging certain dependencies
* Schedule to run weekly

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
+    "schedule:earlyMondays"
   ],
   "packageRules": [{
     "packagePatterns": ["^eslint"],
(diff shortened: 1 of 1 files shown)


carlobernardini:
For reference:

1. [Configuration options](https://docs.renovatebot.com/configuration-options/)
2. [packageRules options](https://docs.renovatebot.com/configuration-options/#packagerules)

mukiienko:
Let's try that config. What do you think about schedule updates in some working hours?
https://docs.renovatebot.com/presets-schedule/

carlobernardini:
> Let's try that config. What do you think about schedule updates in some working hours?
> https://docs.renovatebot.com/presets-schedule/

@mukiienko Yes I was thinking of that too, we could also schedule it to run weekly / monthly..

PR #296, opened 2020-06-02
feat(AutosuggestMulti): improve component to autosuggest requirements

mukiienko:
Improvements:
- 'isLoading' added support loading state;
- 'suggestionItemRenderer' prop defines render function for suggestion item;
- the first item doesn't have default highlighting while input empty. The highlighting appears only when a user typing something in the input;

### DIFF
--- a/src/components/Select/SelectBase/interfaces.ts
+++ b/src/components/Select/SelectBase/interfaces.ts
@@ -15,6 +15,8 @@ export interface CommonProps<S> extends React.HTMLAttributes<HTMLDivElement> {
     rootRef?: React.RefObject<HTMLDivElement>;
     /** suggestions list ref */
     listRef?: React.RefObject<HTMLUListElement>;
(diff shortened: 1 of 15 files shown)


carlobernardini:
Can you please add a short description / summary of what exactly the improvement entails, for reference?

src/components/Select/SelectBase/interfaces.ts
```diff
@@ -15,6 +15,8 @@ export interface CommonProps<S> extends React.HTMLAttributes<HTMLDivElement> {
     rootRef?: React.RefObject<HTMLDivElement>;
     /** suggestions list ref */
     listRef?: React.RefObject<HTMLUListElement>;
+    /** onFocus() is called when the component is focussed */
```
igorarkhipenko, line 18: there's a typo in 'focused'

mukiienko: Thanks

PR #299, opened 2020-06-10
fix(LocationSelector): fix modal opening on form submitting

igorarkhipenko:
* Fix modal opening on form submitting
* Fix background color of the main input button

### DIFF
--- a/src/components/LocationSelector/LocationSelector.tsx
+++ b/src/components/LocationSelector/LocationSelector.tsx
@@ -69,11 +69,11 @@ interface Props {
     /** label to be used on the clear all button */
     clearLabel: string;
     /** function called with location object as an argument when it is selected from the suggestions */
-    onAddLocation: (Location) => void;
+    onAddLocation: (location: Location) => void;
     /** function called with a location details as an argument to be changed */
-    onUpdateLocation: (Location) => void;
+    onUpdateLocation: (location: Location) => void;
     /** function with a locationId as an argument to be removed */
-    onRemoveLocation: (Location) => void;
+    onRemoveLocation: (location: Location) => void;
     /** callback function for the Clear button click */
     onRemoveAllLocations: () => void;
     /** callback function for closed modal */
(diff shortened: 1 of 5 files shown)


stories/LocationSelector.tsx
```diff
@@ -196,41 +196,52 @@ storiesOf('Organisms|LocationSelector', module)
...
+                    e.preventDefault();
                 }}
-                onLocationAutocompleteError={() =>
-                    console.log('onLocationAutocompleteError was called')
-                }
-            />
+            >
+                <input value="dsadsa" />
+                <input value="dsadsadsadsadsa" />
+                <input value="dsadsadsadsadsa421321" />
```
antipin, line 208: Let's make it bit nicer, I'd suggest to use OneUI's `<Input />` and set empty values (or something meaningful)

src/components/LocationSelector/LocationSelector.tsx
```diff
@@ -213,7 +213,7 @@ export const LocationSelector: React.FC<Props> = (props) => {
                 isFocused={isWrapperFocused}
                 showClearButton={hasLocationsSelected}
                 clearLabel={clearLabel}
-                onClick={handleOpenModal}
+                onMouseUp={handleOpenModal}
```
antipin, line 216: I was investigating why you had to use `any` for `LocationSelector`...  Although, the reason is different, could you pls fix these types:

* [onAddLocation](https://github.com/textkernel/oneui/pull/299/files#diff-8ecbfeea0d57301b84c682e7f6873a11R72)
* [onUpdateLocation](https://github.com/textkernel/oneui/pull/299/files#diff-8ecbfeea0d57301b84c682e7f6873a11R74)
* [onRemoveLocation](https://github.com/textkernel/oneui/pull/299/files#diff-8ecbfeea0d57301b84c682e7f6873a11R74)

Should be like this:
```typescript
onAddLocation: (location: Location) => void;
onUpdateLocation: (location: Location) => void;
onRemoveLocation: (location: Location) => void;
```

stories/LocationSelector.tsx
igorarkhipenko: Oops, I didn't mean to add it 😬

PR #305, opened 2020-06-17
fix(LocationSelector): fixed a bug where a modal got opened during tab switching

igorarkhipenko:
Story [ONEUI-193](https://jira.textkernel.nl/browse/ONEUI-193)

### DIFF
--- a/src/components/LocationSelector/LocationSelector.tsx
+++ b/src/components/LocationSelector/LocationSelector.tsx
@@ -129,12 +129,29 @@ export const LocationSelector: React.FC<Props> = (props) => {
 
     const [isOpen, setIsOpen] = React.useState(false);
     const [isWrapperFocused, setIsWrapperFocused] = React.useState(false);
+    const [isBrowserTabVisible, setIsBrowserTabVisible] = React.useState(true);
     const buttonRef = React.useRef<HTMLButtonElement>();
 
     const hasLocationsSelected = selectedLocations && selectedLocations.length > 0;
 
+    React.useEffect(() => {
+        const handleFocushandleVisibilityChange = () => {
+            if (document.hidden) {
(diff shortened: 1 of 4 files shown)


src/components/LocationSelector/LocationSelector.tsx
```diff
@@ -129,20 +129,37 @@ export const LocationSelector: React.FC<Props> = (props) => {
...
+            }
+        };
+
+        document.addEventListener('visibilitychange', handleFocushandleVisibilityChange);
+
+        return () => {
+            document.removeEventListener('visibilitychange', handleFocushandleVisibilityChange);
+        };
+    });
+
```
mukiienko, line 152: I noticed that code is duplicated in `SelectBase` component. Could we create some general wrapper for managing the tab's switching behavior? or move it to the `hooks` folder and reuse it after?

igorarkhipenko: making a hook out of that is a nice idea, thanks

src/hooks/useBrowserTabVisibilityChange.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    React.useEffect(() => {
+        document.addEventListener('visibilitychange', handleFocusHandleVisibilityChange);
+
+        return () => {
+            document.removeEventListener('visibilitychange', handleFocusHandleVisibilityChange);
+        };
+    });
+
+    return isBrowserTabVisible;
```
mukiienko, line 21: It makes sense to wrap all code of the hook to `useEffect` with the second argument `[]`.
Because now every rerender of the component add `addEventListener` to the document. It should be added just once on the init state.

mukiienko:
The fix works! it can be merged

PR #347, opened 2020-07-27
feat(LocationAutocomplete): add an optional Clear button, update snapshots

igorarkhipenko:
Story: [REACH-278](https://jira.textkernel.nl/browse/REACH-278)

### DIFF
--- a/src/components/Autosuggest/Autosuggest.js
+++ b/src/components/Autosuggest/Autosuggest.js
@@ -147,6 +147,7 @@ export class Autosuggest extends React.Component {
         const { onClearAllSelected } = this.props;
 
         e.stopPropagation();
+        this.setState({ inputValue: '' });
         if (onClearAllSelected) {
             onClearAllSelected();
         }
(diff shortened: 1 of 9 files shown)


src/components/Autosuggest/Autosuggest.js
```diff
@@ -328,8 +329,9 @@ export class Autosuggest extends React.Component {
 
         const stateAndProps = { ...this.props, ...this.state };
         const hideInputPlaceholder = !focused && !!selectedPlaceholder;
-        const doShowClearButton =
-            showClearButton && !!selectedSuggestions && !!selectedSuggestions.length && !focused;
+        const doShowClearButton = saveSelectedValueToInput
+            ? showClearButton && inputValue.length > 0 && !focused
```
antipin, line 333: When the field is focused, Clear button is not visible.
Maybe I'm missing something, and there was a discussion behind it, but for me it feels like a bug.
Why a user can't clear the field when it is focused?

igorarkhipenko: Unfortunately it is not a straight-forward solution and it requires some breaking changes or updates in structuring of Autosuggest component. I noted that down to [REACH-278](https://jira.textkernel.nl/browse/REACH-278) so we can continue this discussion from there.

PR #360, opened 2020-08-04
feat(RangeSlider): new component

eszthoff:
Story: [ONEUI-196](https://jira.textkernel.nl/browse/ONEUI-196)

# Checklist


src/components/Sliders/RangeSlider/RangeSlider.js
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Range from 'rc-slider/lib/Range';
+import '../Slider.scss';
+
+export const RangeSlider = React.forwardRef((props, ref) => <Range ref={ref} {...props} />);
+
+RangeSlider.propTypes = {
```
carlobernardini, line 8: Should callback props like `onChange` / `onAfterChange` be listed here as well?

eszthoff: I was wandering about it too. I'm not really sure what was the original motivation of listing these in Slider, since they are all just duplicates of props from `rc-slider`. The only thing I see as an advantage is the better description in storybook.

carlobernardini: Yes, I thought the same. To me it's a bit odd that we include some arbitrary props and do not include others (that are equally important). I'd be a bit puzzled as to what props the component supports when only looking at Storybook.
Perhaps it would be more evident if we don't duplicate any prop definitions at all (if not 'ours') and just refer to the [rc-slider docs](https://github.com/schrodinger/rc-slider#props)? Then we also don't have to keep the definition in sync (in case of major changes to that package). Alternatively I think we should include all props, at least those that are relevant to our use cases.

eszthoff: ok, I removed them all.

carlobernardini:
@eszthoff there are some failing checks that need to be resolved before code review (and the PR is still marked as WIP)

PR #370, opened 2020-08-17
WIP: feat(Select): new component

eszthoff:
Story: [ONEUI-210](https://jira.textkernel.nl/browse/ONEUI-210)

# Checklist


src/components/Select/Select/Select.tsx
```diff
@@ -0,0 +1,98 @@
...
+        selectedSuggestion,
+        suggestionToString,
+        suggestionItemRenderer,
+        onSelectionAdd,
+        onBlur,
+        onFocus,
+        rootRef,
+        listRef,
+        ...rest
+    } = props;
```
igorarkhipenko, line 30: There's a bug of Select component being closed once it's opened and the browser tab is switched. It can be fixed with `useBrowserTabVisibilityChange`. You can take ComboboxMulti component as an example.

src/components/Select/Select/Select.tsx
```diff
@@ -0,0 +1,98 @@
...
+            onBlur?.();
+        } else {
+            onFocus?.();
+        }
+    };
+
+    return (
+        <div ref={rootRef} {...rest} {...block({})}>
+            <div {...elem('main', { isOpen })}>
+                <FieldWrapper isFocused={isOpen} {...elem('field', {})}>
```
igorarkhipenko, line 61: It's not really necessary to pass an empty object to `elem` and `block` functions as a second parameter so it can be omitted.

src/components/Select/Select/Select.tsx
```diff
@@ -0,0 +1,98 @@
...
+                        ) : (
+                            <IoMdArrowDropdown {...elem('dropdownIcon', {})} />
+                        )}
+                    </div>
+                    <List
+                        {...getMenuProps({ ref: listRef })}
+                        {...elem('list', { isOpen })}
+                        isControlledNavigation
+                    >
+                        {isOpen ? (
```
igorarkhipenko, line 82: Ternary operator can be replaced with `&&` in this case

src/components/Select/Select/Select.tsx
```diff
@@ -0,0 +1,98 @@
...
+import styles from './Select.scss';
+
+const { block, elem } = bem('Select', styles);
+
+interface Props<S> extends Omit<BasicSelectProps<S>, 'isProminent'> {
+    /** The suggestions that is currently selected */
+    selectedSuggestion: S;
+}
+
+export function Select<S>(props: Props<S>) {
```
igorarkhipenko, line 18: Can we rename the folder containing all of the select components to avoid repetitions? E.g. SelectComponents

src/components/Select/Select/Select.tsx
eszthoff: Our custom hook doesn't capture the event. Also found that downshift is inconsistent in its state update in this case, meaning sometimes it collapses, and sometimes not. After offline discussions with both Ihor and Carlo, created an issue in their repo: https://github.com/downshift-js/downshift/issues/1157

src/components/Select/Select/Select.tsx
eszthoff: good catch. removed

src/components/Select/Select/Select.tsx
eszthoff: changed

src/components/Select/Select/Select.tsx
eszthoff: done

PR #377, opened 2020-08-27
feat(Modal): add prop to toggle modal position to scrollable/fixed

eszthoff:
Story: [JF-2367](https://jira.textkernel.nl/browse/JF-2367)

# Checklist


src/components/Modal/Modal.scss
```diff
@@ -1,10 +1,19 @@
 .Modal {
+    &__body--open {
```
antipin, line 2: Having element `body` with modifier `open` doesn't make much sense because if it is in a DOM tree, it is already open (in other words such thing as `body--closed` doesn't exists).

So, can be just `body`.

src/components/Modal/Modal.tsx
```diff
@@ -19,6 +45,7 @@ export const Modal = (props) => {
     } = props;
 
     const { className: portalClass } = block({ className: portalClassName, ...rest });
+    const { className: bodyOpenClass } = elem('body--open', props);
```
antipin, line 48: Here, if we'd need to have a class name `body--open`, that proper way to achieve it would be:
`elem('body', { ...props, open: true });`

But since [we don't need `open` modifier](https://github.com/textkernel/oneui/pull/377/files#diff-df00b1220859d4028b7de06ff6d065b2R2) at all, it can be just `elem('body', props);`

src/components/Modal/Modal.tsx
```diff
@@ -29,12 +56,24 @@ export const Modal = (props) => {
     const { className: contentEnteredClass } = elem('content--entered', props);
     const { className: contentExitedClass } = elem('content--exited', props);
```
antipin, line 57: Ah, I believe you copypasted `elem('body--open', props);` from here, bu these are examples of not correct usage of `elem` function.

Should be:
```ts
const { className: contentEnteredClass } = elem('content', { ...props, entered: true });
const { className: contentExitedClass } = elem('content', { ...props, exited: true });
```

antipin (changes requested):
Nice improvement! Left few comments

src/components/Modal/Modal.tsx
eszthoff: Because the way this is used in React Modal, setting it in this way makes final DOM having repeated classes, that looks messy. So I kept it as it is now. Alternatively we could rename the enter/exit classes in the scss files so that they don't use the usual modifier patterns to avoid internal confusion, e.g. `&__enteredContent` instead of `&__content--entered`.

antipin: Ah, I see. Now I remember there was a task to cover this case in bem. Ok, let's keep it this way then so far.

antipin (approved):
👍

PR #388, opened 2020-09-07
fix(Drawer): eliminated two callbacks on expand button click

mukiienko:


src/components/Drawer/Drawer.tsx
```diff
@@ -103,26 +117,9 @@ export const Drawer = (props) => {
...
-    /** Set open/close status for state */
-    isExpanded: PropTypes.bool,
-    /** Fired after click open/close button */
-    onClick: PropTypes.func,
-    /** Node(s) to be rendered inside the container */
-    children: PropTypes.node.isRequired,
-};
-
 Drawer.defaultProps = {
     initialIsExpanded: false,
```
carlobernardini, line 121: Should these two `defaultProps` still stay if you removed all of the `propTypes`?

mukiienko: yes, we still need to use `propTypes` ](https://github.com/textkernel/oneui/blob/master/CONTRIBUTING.md#type-definitions-for-props-and-states), otherwise StoryBook doesn't pick up it.

carlobernardini (approved):

PR #389, opened 2020-09-09
fix(AutosuggestMulti): only show noSuggestions if user already started to search

eszthoff:
Story: [JF-2387](https://jira.textkernel.nl/browse/JF-2387)

# Checklist

### DIFF
--- a/src/components/SelectComponents/AutosuggestMulti/AutosuggestMulti.tsx
+++ b/src/components/SelectComponents/AutosuggestMulti/AutosuggestMulti.tsx
@@ -9,11 +9,15 @@ import {
     SelectInputFieldProps,
     FocusedRendererHelpers,
     BlurredRendererHelpers,
+    ListRendererHelper,
 } from '../SelectBase';
 import styles from './AutosuggestMulti.scss';
 import { BACKSPACE_KEY, ESCAPE_KEY, ENTER_KEY } from '../../../constants';
 
-interface Props<S> extends BasicSelectProps<S>, SelectClearButtonProps, SelectInputFieldProps {
+interface Props<S>
+    extends BasicSelectProps<S>,
+        SelectClearButtonProps,
+        Omit<SelectInputFieldProps, 'clearInputAfterSelection'> {
     /** HTML id for the input element */
     id?: string;
     /** Creates a unique (React) key for a suggestion item. If undefined suggestionToString will be used */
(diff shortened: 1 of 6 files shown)


stories/Select.tsx
```diff
@@ -139,8 +139,122 @@ storiesOf('Organisms|Select Components', module)
...
         'AutosuggestMulti',
-        ({ parameters }: any) => {
-            const store = parameters.getStore();
+        (storyContext) => {
+            const store = storyContext?.parameters.getStore();
+            const getSuggestions = (): TSuggestion[] => {
+                const suggestions = SUGGESTIONS.filter(
+                    (item: TSuggestion) =>
+                        item.name.toLocaleLowerCase() !==
+                            store.get('inputValue').toLocaleLowerCase() &&
```
igorarkhipenko, line 148: this condition should be removed because it doesn't allow to select an item once there's an exact match with the input value

src/components/SelectComponents/AutosuggestMulti/AutosuggestMulti.tsx
```diff
@@ -164,6 +168,22 @@ export function AutosuggestMulti<S>(props: Props<S>) {
         </div>
     );
 
+    const renderList: ListRendererHelper<S> = (listProps) => {
+        const { inputValue: inputToList } = listProps;
```
igorarkhipenko, line 172: Unexpected 'No suggestion' placeholder still gets displayed after selecting a suggestion.

eszthoff: was a bug in the story. Should be fine now

stories/Select.tsx
eszthoff: good catch

stories/Select.tsx
```diff
@@ -144,23 +144,17 @@ storiesOf('Organisms|Select Components', module)
             const getSuggestions = (): TSuggestion[] => {
                 const suggestions = SUGGESTIONS.filter(
                     (item: TSuggestion) =>
-                        item.name.toLocaleLowerCase() !==
-                            store.get('inputValue').toLocaleLowerCase() &&
-                        item.name
-                            .toLocaleLowerCase()
-                            .includes(store.get('inputValue').toLocaleLowerCase()) &&
```
igorarkhipenko: `item.name.toLocaleLowerCase().includes(store.get('inputValue').toLocaleLowerCase())` should have not been removed as well as `store.set` below. Filtering by input functionality doesn't work without them.

eszthoff: It was buggy and it is just an example implementation, so I removed it. It is not the main point here, and you can think of the suggestions like in profession suggestions which are smart and not only letter based.

PR #390, opened 2020-09-09
feat: implemented BooleanQueryTokenizer package

antipin:
Move BooleanQueryTokenizer from reach-ui project to OneUI library.

BooleanQueryTokenizer allows to tokenize boolean query language.
Also it provides a method `extractWordsAndPhrases` that returns bag of phrases and words.

# Checklis

### DIFF
--- a/src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
+++ b/src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
@@ -1,0 +1,175 @@
+/**
+ * @module BooleanQueryTokenizer is responsible for tokenizing boolean query language. Also
+ * it provides a method `extractWordsAndPhrases` that returns bag of phrases and words.
+ *
+ * Glossary:
+ * A _token_ is a pair consisting of a token name and matched _lexeme_.
+ * The token name represents a kind of lexical unit, e.g.,
+ * particular keyword, phrase, operator, separator, etc.
+ *
+ * A _lexeme_ is a sequence of characters in the input string that matches the pattern for
+ * a token and is identified by the lexical analyzer as an instance of that token.
+ *
+ * A _pattern_ is a description of the form that the lexemes of a token may take.
+ * In the case of a keyword as a token, the pattern is just the sequence of characters that
+ * form the keyword. For identifiers and some other tokens, the pattern is more complex
+ * structure that is matched by many strings.
+ */
+
+export enum TokenType {
+    unknown,
+    separator,
+    word,
+    wildcardWord,
+    wildcard,
+    phrase,
+    booleanOperator,
+    aroundOperator,
+    modifier,
+    parentheses,
+}
+type TokenPattern = [
+    TokenType, // Token type
+    RegExp // Pattern of a token in a RegExp form
+];
+type Token = {
+    type: TokenType;
+    lexeme: string;
+};
+
+// Character classes
+const spaceSymbols = '\\s\\t\\n';
+/**
+ * We need a character class that matches a letter from any alphabet.
+ * The easiest way to achieve it is to use the regexp /\p{Letter}/,
+ * but only if RegExp engine supports Unicode Property Escapes.
+ * Unfortunately, RegExp engine in Internet Explorer does not. So we have to use a slightly
+ * modified version of transpiled /\p{Letter}/ expression.
+ * @see https://github.com/tc39/proposal-regexp-unicode-property-escapes
+ * @see https://mothereff.in/regexpu#input=var+regex+%3D+/%5Cp%7BL%7D/u%3B&unicodePropertyEscape=1
+ * If you need to decode it to readable characters, @see https://r12a.github.io/app-conversion/
+ */
+const letterSymbols =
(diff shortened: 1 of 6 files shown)


stories/packages/BooleanQueryTokenizer.tsx
```diff
@@ -0,0 +1,55 @@
...
+    'java* NOT *script\n' +
+    'NEAR (rust OR tex*nel) OR\n' +
+    '-.NET "TO BE OR NOT TO BE"';
+
+const columnStyle = {
+    width: '30%',
+    padding: '1em',
+};
+
+storiesOf('packages|BooleanQueryTokenizer', module).add('BooleanQueryTokenizer', () => {
```
eszthoff, line 17: Good call creating a section for packages. Maybe can move the highlighter here as well?

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -0,0 +1,178 @@
...
+ * particular keyword, phrase, operator, separator, etc.
+ *
+ * A _lexeme_ is a sequence of characters in the input string that matches the pattern for
+ * a token and is identified by the lexical analyzer as an instance of that token.
+ *
+ * A _pattern_ is a description of the form that the lexemes of a token may take.
+ * In the case of a keyword as a token, the pattern is just the sequence of characters that
+ * form the keyword. For identifiers and some other tokens, the pattern is more complex
+ * structure that is matched by many strings.
+ */
```
eszthoff, line 17: Great to have all this documentation!
Small question, what is `lexeme` stands for?

stories/packages/BooleanQueryTokenizer.tsx
```diff
@@ -0,0 +1,55 @@
...
+            </div>
+            <div style={columnStyle}>
+                <h3>Words and phrases</h3>
+                <pre>{JSON.stringify(wordsAndPhrases, null, 4)}</pre>
+            </div>
+            <div style={columnStyle}>
+                <h3>Tokens</h3>
+                <pre>{JSON.stringify(tokens, null, 4)}</pre>
+            </div>
+        </div>
```
eszthoff, line 53: Maybe out of scope here, but would be interesting to see how to use this together with the highlighter, especially for none trivial cases. E.g. excluding `NOT xxx` or when using `NEAR` or `java * script` type of things.

eszthoff (changes requested):

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: It's a commonly used term in CS/parsing field. I believe it came from linguistics https://en.wikipedia.org/wiki/Lexeme

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -0,0 +1,178 @@
...
+                // so we should remove them.
+                .filter((str) => str)
+        );
+    }
+
+    /**
+     * `tokenizeCurrentChunk` tries to find the best possible match for the current
+     * part of the input text
+     */
+    // eslint-disable-next-line class-methods-use-this
```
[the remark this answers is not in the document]
antipin: The goal is to have all public methods of BooleanQueryTokenizer as non-static methods for simplicity.

Otherwise, the interface of the module would look like this:
```ts
BooleanQueryTokenizer.tokenize() // <- static
BooleanQueryTokenizer#extractWordsAndPhrases() // <- not static
```

But indeed, it makes more sense to move this `// eslint-disable-next-line class-methods-use-this` to `BooleanQueryTokenizer.tokenize()` directly.

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
eszthoff: we learn something new everyday. Or is it relearning it 🤔

antipin: > we learn something new everyday.
true :D

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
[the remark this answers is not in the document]
antipin: addressed

src/packages/BooleanQueryTokenizer/__tests__/BooleanQueryTokenizer.spec.js
```diff
@@ -0,0 +1,266 @@
...
+            ).toEqual(['Fasten', 'your', 'belts.', "It's", 'going', 'to', 'be', 'a', 'night.']);
+            expect(tokenizer.extractWordsAndPhrases('people like to * each other')).toEqual([
+                'people',
+                'like',
+                'to',
+                'each',
+                'other',
+            ]);
+
+            expect(tokenizer.extractWordsAndPhrases('people like to * each other')).toEqual([
```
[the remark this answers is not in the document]
antipin: good catch

stories/packages/BooleanQueryTokenizer.tsx
antipin: Since Highlighter is a mix of components and package, I guess it shouldn't be used as a package only.
And current Highlighter story is made for component, so it would be weird to move it under package directory.

stories/packages/BooleanQueryTokenizer.tsx
antipin: I think is is out of scope indeed. `BooleanQueryTokenizer` doesn't depend from Higlighter (the other way around maybe, but still there is no hard dependency).

src/packages/BooleanQueryTokenizer/__tests__/BooleanQueryTokenizer.spec.js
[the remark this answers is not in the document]
antipin: fixed

stories/packages/BooleanQueryTokenizer.tsx
eszthoff: you are right

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -0,0 +1,177 @@
...
+            if (lexemeMatch !== null) {
+                const [lexeme] = lexemeMatch;
+                if (!tokenCandidate) {
+                    tokenCandidate = { type: name, lexeme };
+                }
+                // Longer lexeme always wins. If tokens' lexemes are equal in length, then wins a
+                // token who's lexeme has been matched by a pattern with higher priority.
+                // Which means the current one (because we iterate through patterns from left to
+                // right and right patterns have higher priority).
+                if (lexeme.length >= tokenCandidate.lexeme.length) {
```
eszthoff, line 170: This condition will still be true also when the candidate was just set for the first time, I would still use `else if` here. It will also be clearer to read the logic (even though it is quite clear given the comment 👍 )

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -0,0 +1,177 @@
...
+            ),
+        ],
+        [TokenType.parentheses, new RegExp(`^${parentheses}`)],
+    ];
+
+    /**
+     * Tokenizer splits an input text into meaningful chunks and labels each chunk according to its
+     * lexical meaning, suchwise producing a list of tokens.
+     */
+    // eslint-disable-next-line class-methods-use-this
```
eszthoff, line 102: Why not make all functions static. That makes sense, there is no actual need to initialise the class or have different instances of it running somewhere. It can be used simply like a little library.

eszthoff (changes requested):

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: That's true, addressed

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: Makes sense, updated

eszthoff (approved):

PR #398, opened 2020-09-15
feat(Select): Implement 'disabled' behaviour for ComboboxMulti

igorarkhipenko:
[JF-1997](https://jira.textkernel.nl/browse/JF-1997): Implement 'disabled' behaviour for AutosuggestMulti and ComboboxMulti, add keepExpandedAfterSelection prop to ComboboxMulti

# Checklist


stories/Select.tsx
```diff
@@ -116,6 +120,7 @@ storiesOf('Organisms|Select Components', module)
                         noSuggestionsPlaceholder={text('No suggestions', 'No suggestions found...')}
                         suggestions={getSuggestions()}
                         suggestionToString={SUGGESTION_TO_STRING}
+                        disabled={store.get('disabled')}
```
eszthoff, line 123: I think we should also add an option to toggle it from the knobs

src/components/FieldWrapper/FieldWrapper.js
```diff
@@ -49,11 +49,14 @@ FieldWrapper.propTypes = {
     onClear: PropTypes.func,
     /** to style the field as it is focused */
     isFocused: PropTypes.bool,
+    /** defines if the component is disabled */
```
eszthoff, line 52: Perhaps it is a good opportunity to move this component to TS if you are already changing it.

src/components/FieldWrapper/FieldWrapper.js
```diff
@@ -49,11 +49,14 @@ FieldWrapper.propTypes = {
...
+    disabled: PropTypes.bool,
 };
 
 FieldWrapper.defaultProps = {
     showClearButton: false,
     clearLabel: '',
     onClear: null,
     isFocused: false,
+    disabled: false,
```
eszthoff, line 61: Should add this prop to the story as well

eszthoff (changes requested):

src/components/FieldWrapper/FieldWrapper.js
igorarkhipenko: I didn't add that on purpose since it doesn't change a lot. It just makes FieldWrapper borders disabled and the component it wraps remains active. Maybe it would be better to rework the representative component so we can have disabling option which works for both the wrapper and the child component?

eszthoff: What do you mean by "representative component"?

igorarkhipenko: I mean FieldWrapper component we have on the storybook page

eszthoff: I think storybook should be kept as simple as possible. If we write complex implementations even for basic components, it won't be representative. I think it is ok to just add it as a knob, so one can see what the styling of the disabled component looks like

eszthoff (approved):

PR #414, opened 2020-10-09
feat(SelectButtonGroup): restyled component

carlobernardini:
Restyles SelectButtonGroup components

# Checklist

### DIFF
--- a/src/components/SelectButtonGroup/SelectButton/SelectButton.scss
+++ b/src/components/SelectButtonGroup/SelectButton/SelectButton.scss
@@ -1,50 +1,84 @@
 .SelectButton {
-    box-shadow: inset 0 -3px 0 var(--color-neutral-25);
-    height: 38px;
+    // height: 38px;
     box-sizing: border-box;
     padding: var(--spacing-3x) var(--spacing-4x);
-    border: 1px solid var(--color-neutral-40);
-    border-right: none;
     white-space: nowrap;
     text-overflow: ellipsis;
     text-align: center;
     overflow: hidden;
     cursor: pointer;
+    outline: none;
+    color: var(--color-neutral-75);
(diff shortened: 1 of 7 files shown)


src/components/SelectButtonGroup/SelectButton/SelectButton.scss
```diff
@@ -1,49 +1,96 @@
...
+
+    + .SelectButton {
+        box-shadow: -1px 0 0 0 var(--color-background);
+    }
+
+    &[disabled] {
+        cursor: not-allowed;
     }
 
     &--isSelected.SelectButton--selectedContext_neutral {
```
eszthoff, line 36: would it make it easier to read this if this and the next 2 rules would be combined?
```
&--isSelected.SelectButton--selectedContext {
  &_neutral {...}
  &_good {...}
  &_brand {...}
}
```

eszthoff (changes requested):

src/components/SelectButtonGroup/SelectButton/SelectButton.scss
carlobernardini: We could indeed do that for `good` and `brand` but not including `neutral` because its styles (`border-color` and `background-color`) are different. Maybe at this point we should just make it support any context out of the box instead of a particular subset.

eszthoff: sure, if you think that's easy. we haven't implemented the other contexts before, because we had no design for them

eszthoff: you can still scope the rules though as above, they can all have their own rules

eszthoff (approved):

PR #427, opened 2020-10-26
feat: add statuses to Teaser

agorodetskaya:
- add statuses array
- add fixed height
- implement 3 rows architecture where:
 1.  `title` + `location`
 2.  `subTitle` (or `details` if no `subTitle`)
3.  `details` (if there is `subTitle`) + `statuses` (always on the last row in the right split by ` & `)

image
- updated storybook story of Teaser component. I created 3 stories:
1. all fields
2. with one status
3. without statuses
I had to do this, because in storybook we can't change count of array elements or remove array at all. Also I don't want to create some `withStatuses` boolean knob, because it doesn't represent any real component prop.

### DIFF
--- a/src/components/Teaser/Teaser.scss
+++ b/src/components/Teaser/Teaser.scss
@@ -2,10 +2,22 @@
     min-width: 0;
     margin: 0;
     display: flex;
+    flex: 1;
     flex-direction: column;
+    height: 63px;
 
-    &__titleLine {
+    &__line {
         display: flex;
+        flex: 1;
+        align-items: center;
+        padding-bottom: var(--spacing-2x);
+    }
+
+    &__lastLine {
+        display: flex;
+        flex: 1;
+        align-items: center;
+        justify-content: flex-end;
     }
 
     &__title {
@@ -28,8 +40,14 @@
         text-overflow: ellipsis;
     }
 
-    &__titleLine:not(:last-child),
-    &__subTitle:not(:last-child) {
-        padding-bottom: var(--spacing-2x);
+    &__details {
+        flex: 1;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+    &__statuses {
+        padding-left: var(--spacing-2x);
     }
 }
--- a/src/components/Teaser/Teaser.scss
+++ b/src/components/Teaser/Teaser.scss
@@ -11,13 +11,11 @@
         flex: 1;
         align-items: center;
         padding-bottom: var(--spacing-2x);
-    }
 
-    &__lastLine {
-        display: flex;
-        flex: 1;
-        align-items: center;
-        justify-content: flex-end;
+        &:last-child {
+            padding-bottom: 0;
+            justify-content: flex-end;
+        }
     }
 
     &__title {
@@ -48,6 +46,6 @@
     }
 
     &__statuses {
-        padding-left: var(--spacing-2x);
+        padding-left: var(--spacing-5x);
     }
 }
--- a/src/components/Teaser/Teaser.tsx
+++ b/src/components/Teaser/Teaser.tsx
@@ -27,6 +27,16 @@ const { block, elem } = bem('Teaser', styles);
 export const Teaser: React.FC<Props> = (props) => {
     const { title, subTitle, location, details, statuses, ...rest } = props;
 
+    const subTitleElement = (
+        <Text inline context="accent" {...elem('subTitle', props)}>
+            {subTitle}
+        </Text>
+    );
+    const detailsElement = (
+        <Text inline context="muted" {...elem('details', props)}>
+            {details}
+        </Text>
+    );
     return (
         <div {...rest} {...block(props)}>
             <div {...elem('line', props)}>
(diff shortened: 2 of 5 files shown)


src/components/Teaser/Teaser.scss
```diff
@@ -2,10 +2,22 @@
...
+        flex: 1;
+        align-items: center;
+        padding-bottom: var(--spacing-2x);
+    }
+
+    &__lastLine {
+        display: flex;
+        flex: 1;
+        align-items: center;
+        justify-content: flex-end;
```
antipin, line 20: I'd suggest to use `:last-child` pseudo-selector here instead of introducing new element `lastLine` since it exists especially for this kind of cases
```css
    &__line:last-child {
        padding-bottom: 0;
        justify-content: flex-end;
    }
```

stories/Teaser.tsx
```diff
@@ -1,18 +1,56 @@
 import * as React from 'react';
 import { storiesOf } from '@storybook/react';
-import { withKnobs, text } from '@storybook/addon-knobs';
+import { object, text, withKnobs } from '@storybook/addon-knobs';
 import { Teaser } from '@textkernel/oneui';
 
 storiesOf('Molecules|Teaser', module)
     .addDecorator(withKnobs)
-    .add('Teaser', () => {
+    .add('all fields', () => {
```
antipin, line 8: For consistency with stories below, let's also add "with" `all fields` -> `With all fields` (also mind capital W)

stories/Teaser.tsx
```diff
@@ -1,18 +1,56 @@
...
+                    }),
+                    object('Imported status', {
+                        label: 'Imported',
+                        tooltip: 'Imported two days ago',
+                    }),
+                ]}
+            />
+        );
+    })
+    .add('with one status', () => {
```
antipin, line 29: All the other stories' names start with capital letter, let's do the same here

stories/Teaser.tsx
```diff
@@ -1,18 +1,56 @@
...
+                statuses={[
+                    object('Viewed status', {
+                        label: 'Viewed',
+                        tooltip: 'Viewed one day ago',
+                    }),
+                ]}
+            />
+        );
+    })
+    .add('without statuses', () => {
```
antipin, line 46: All the other stories' names start with capital letter, let's do the same here

src/components/Teaser/Teaser.tsx
```diff
@@ -31,16 +39,43 @@ export const Teaser: React.FC<Props> = (props) => {
...
+                    </Text>
+                )}
+            </div>
+            <div {...elem('lastLine', props)}>
+                {subTitle && details && (
+                    <Text inline context="muted" {...elem('details', props)}>
+                        {details}
+                    </Text>
+                )}
+                {statuses && statuses.length > 0 && (
```
antipin, line 60: How about extract wrapped `subTitle` and `details` into variables in order not to repeat same code?

```tsx
...

export const Teaser: React.FC<Props> = (props) => {
    const { title, subTitle, location, details, statuses, ...rest } = props;
// 👀 <---
    const subTitleElem = (
        <Text inline context="accent" {...elem('subTitle', props)}>
                {subTitle}
        </Text>
    );
// 👀 <---
    const detailsElem = (
        <Text inline context="muted" {...elem('details', props)}>
                {details}
        </Text>
    );

    return (
            <div {...elem('line', props)}>
                <Text inline context="brand" {...elem('title', props)}>
                    {title}
                </Text>
                {location && (
                    <Text inline context="muted" {...elem('location', props)}>
                        {location}
                    </Text>
                )}
            </div>
            <div {...elem('line', props)}>
// 👀 <---
                {subTitle ? subTitleElem : }
                {!subTitle && details && detailsElem}
            </div>
            <div {...elem('line', props)}>
// 👀 <---
                {subTitle && details && detailsElem}
    ...
```

antipin (changes requested):
Works perfect! 👍 
Few comments on the code

src/components/Teaser/Teaser.scss
```diff
@@ -28,8 +40,14 @@
...
-        padding-bottom: var(--spacing-2x);
+    &__details {
+        flex: 1;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+    &__statuses {
+        padding-left: var(--spacing-2x);
```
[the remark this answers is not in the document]
agorodetskaya: agree!

src/components/Teaser/Teaser.scss
agorodetskaya: agree, looks better!

src/components/Teaser/Teaser.tsx
agorodetskaya: we don't need subTitleElem because it is used once and I usually don't like to move render parts to variable or separate methods, moreover such small. But I don't mind to change :)

src/components/Teaser/Teaser.scss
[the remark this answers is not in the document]
agorodetskaya: with 5x spacing
image

src/components/Teaser/Teaser.tsx
antipin: No need to extract `subTitleElem indeed

antipin (approved):
👍

PR #432, opened 2020-11-27
feat: WeightedResultBar component

eszthoff:

# Checklist


src/components/WeightedResultBar/__tests__/WeightedResultBar.spec.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import toJson from 'enzyme-to-json';
+import { WeightedResultBar } from '../WeightedResultBar';
+
+describe('WeightedResultBar', () => {
+    it('should render correctly in loading state', () => {
+        const wrapper = mount(<WeightedResultBar>Result</WeightedResultBar>);
```
igorarkhipenko, line 7: The unit title is misleading here. Also there are no values passed to the component which breaks calculations in the snapshot below.

src/components/WeightedResultBar/__tests__/__snapshots__/WeightedResultBar.spec.js.snap
```diff
@@ -0,0 +1,56 @@
...
+      animated={false}
+      context="brand"
+      hidden={false}
+      small={true}
+    >
+      <div
+        aria-hidden={false}
+        aria-valuemax="100"
+        aria-valuemin="0"
+        aria-valuenow={NaN}
```
igorarkhipenko, line 40: ^ Right here

src/components/WeightedResultBar/__tests__/WeightedResultBar.spec.js
eszthoff: uupps 👍

carlobernardini:
I suppose transitions between loading to some value will not work this way. Can you confirm? How much would you need to change to make that work?

eszthoff:
@carlobernardini this is a new requirement, can you please elaborate what you need. So far we never had transition between these states. If it is needed it can be done with some animation effect between mounting and unmounting as usual, like using react-transition-group

carlobernardini:
See the demo in PR #428, (un)check `isLoading` in Storybook to see how it looks like.

eszthoff:
@Carlo I added some code, so that the progress bar is now animated when either component is first rendered. So it will be there also if there is no loading state before the actual data is shown and vise versa. I hope that works for you.

PR #434, opened 2020-12-03
feat(LocationSelector): expose Location type

mukiienko:

# Checklist


src/components/LocationSelector/LocationSelector.tsx
```diff
@@ -182,23 +187,23 @@ export const LocationSelector: React.FC<Props> = (props) => {
...
+        const { Geocoder } = window.google.maps;
         const geocoder = new Geocoder();
 
         return findCenter(geocoder, location.place_id)
-            .then((center: any) => {
-                const lng = typeof center.lng === 'function' ? center.lng() : center.lng;
-                const lat = typeof center.lat === 'function' ? center.lat() : center.lat;
+            .then((center) => {
+                const lng = center.lng();
+                const lat = center.lat();
```
antipin, line 198: This looks like implementation change, are you sure it will not break anything?

antipin (approved):

src/components/LocationSelector/LocationSelector.tsx
mukiienko: In order to google API `lng()`, `lat()` should be always function https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/googlemaps/reference/coordinates.d.ts . 
I'm assuming initially the type number was added for testing purposes. @igorarkhipenko  could you please confirm that?

igorarkhipenko: I can't recall it from top of my head for what reason we had that conditional checking here. But indeed the last time I was working with that component I checked both real responses and the API and it's always been a function. I'm not sure if there's a real reason to maintain that behaviour anymore. Just to be sure I'd like to check how this build is doing with JF while you could check Reach projects and then push the pull request forward if it's all good.

mukiienko: The implementation is the same, but the Location object type is a little bit changed. Probably it can cause some typescript errors. I will mark PR as a braking change.

PR #436, opened 2020-12-24
feat(AutosuggestMulti -> Autosuggest): refactor component so it can behave as single select as well

eszthoff:
Story: [ONEUI-228](https://jira.textkernel.nl/browse/ONEUI-228)

# Checklist

BREAKING CHANGE: older `Autosuggest` component has been renamed to `AutosuggestDeprecated`;
`AutosuggestMulti` has been renamed to `Autosuggest`;
`isFirstItemAlwaysVisible` prop has been renamed to `allowMixingSuggestionsAndLoading` and will now display all suggestion before rendering loading placeholders;
`isProminent` prop has been removed


carlobernardini:
PO reviewed, I have several comments:

1. In some Storybook stories, the example doesn't filter like it would in reality. While typing, it only highlights matches but it doesn't filter out nodes that don't match. This seems broken.
2. In the story "Autosuggest with selecting input as free text", when typing something and then selecting the first suggestion, the input is cleared but matching suggestions are still highlighted.
3. The styling is off in some cases. E.g. with "Autosuggest as single-select component", when I type something and hit enter the alignment of my input jumps by a lot. While at it, it would also be nice if we can align the component height with that of Select / ComboboxMulti.
4. Why should we still keep "AutosuggestDeprecated - old"?
5. The commit message is not really in line with conventional commits spec, I think. Suggest not to use scope in this case (just `feat:`) and use `BREAKING CHANGE` only once in the body.
> A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., fix(parser):

I think the implementation should also be reviewed by either @mukiienko or @antipin to check compatibility with Reach.

eszthoff:
After f2f with Carlo, his points 1 and 2 got outscoped, since they were not changed in this story, and are only related to story implementation.
3. addressed in latest changes
4. we need old AutoSuggest since it is used within OneUI in LocationSelector. @carlobernardini will create ticket to refactor that
5. Addressed - finger crossed that change log will be generated correctly

mukiienko:
I have only one comment about the height of the block Autosuggest. It looks like 3 point is not completely addressed:
Screenshot 2021-01-13 at 13 00 00
Screenshot 2021-01-13 at 12 58 57

src/components/SelectComponents/Autosuggest/Autosuggest.tsx
```diff
@@ -30,21 +30,32 @@ interface Props<S>
...
+    /** If suggestions are still loading, i.e. display placeholders */
     isLoading?: boolean;
+    /** Means to overwrite the look and feel of the UI in its blurred state */
+    customBlurParams?: {
+        /** React node that will be rendered to indicate the current state of selections */
+        selectionIndicator: ReactNode;
+        /** Should the input field be shown. Normally should set this to true (hide) */
+        isInputHidden: boolean;
+        /** Should a clear button be shown on hover */
+        showClearButton: boolean;
```
mukiienko, line 50: What do you think about making `showClearButton`, `isInputHidden `,  `selectionIndicator ` optional as well?

mukiienko, line 50: Not really understand why we have `showClearButton` two times in props.

eszthoff:
@Carlo please advise. I rather not start changing LocationSelector as well in this story, but since we made Autosuggest 2px higher to match Select, now it is misaligned with LocationSelector. Is that acceptable, or shall I revert that 2px growth?

src/components/SelectComponents/Autosuggest/Autosuggest.tsx
```diff
@@ -30,21 +30,32 @@ interface Props<S>
...
     useOptimizeListRender?: boolean;
     /** Function to be called when a suggestion is removed  */
     onSelectionRemove?: (item: S) => void;
     /** Function to be called on submitting form */
     onSubmit?: () => void;
-    /** if suggestions are still loading, i.e. display placeholders */
+    /** If suggestions are still loading, i.e. display placeholders */
     isLoading?: boolean;
+    /** Means to overwrite the look and feel of the UI in its blurred state */
+    customBlurParams?: {
```
mukiienko, line 44: It would be nice to have an example of using `customBlurParams` in storybook.

src/components/SelectComponents/Autosuggest/Autosuggest.tsx
eszthoff: I don't think we can make intelligent assumptions of what these would be. That's the whole point of it making a custom param.

eszthoff: These are inside a customBlurParam, so they can use a different use case.

src/components/SelectComponents/Autosuggest/Autosuggest.tsx
eszthoff: There are 2 stories: `single select` and `dance as multi and single select`. You need to select something for it to be passed

mukiienko:
> @carlo please advise. I rather not start changing LocationSelector as well in this story, but since we made Autosuggest 2px higher to match Select, now it is misaligned with LocationSelector. Is that acceptable, or shall I revert that 2px growth?

What is the reason was to increase the height? then we need to increase the height for `Input` component as well.

eszthoff:
> > @carlo please advise. I rather not start changing LocationSelector as well in this story, but since we made Autosuggest 2px higher to match Select, now it is misaligned with LocationSelector. Is that acceptable, or shall I revert that 2px growth?
> 
> What was the reason to increase height? then we need to increase the height for `Input` component as well.

See PO review point 3 above.

mukiienko:
> > > @carlo please advise. I rather not start changing LocationSelector as well in this story, but since we made Autosuggest 2px higher to match Select, now it is misaligned with LocationSelector. Is that acceptable, or shall I revert that 2px growth?
> > 
> > 
> > What was the reason to increase height? then we need to increase the height for `Input` component as well.
> 
> See PO review point 3 above.

It is strange. The height of AutocopleteMulty before changes was aligned with all components (except deprecated Autoseggest). See https://textkernel.github.io/oneui/?path=/story/organisms-select-components--autosuggestmulti

src/components/SelectComponents/Autosuggest/Autosuggest.tsx
eszthoff: It is needed for the logic here: https://github.com/textkernel/oneui/pull/436/files#diff-41965cb515e81f9279b4c06f7984d72b53f05bdbe8bb6c7d4d097cf921465d0fR214

carlobernardini:
@eszthoff I mistakenly took the height of Select as the standard, please revert the 2px increase. Apologies.

mukiienko (approved):

PR #437, opened 2021-01-18
feat(Dropdown): new component

agorodetskaya:
New component which provides navigatable list of actions.
Relates to: https://jira.textkernel.nl/browse/REACH-411

# Checklist

---
Open question: _do we need to blur button force on closing popup?_


mukiienko:
I think we need to convert the task REACH-411 into OneUI task.

src/components/List/ListItem/ListItem.tsx
```diff
@@ -19,6 +19,8 @@ interface Props extends Omit<React.HTMLAttributes<HTMLLIElement>, 'onClick'> {
     highlightContext?: Context | 'default';
     /** Ref to access the li element */
     ref?: React.RefObject<HTMLLIElement>;
+    /** Item identifier */
+    value?: string | number | boolean | object;
```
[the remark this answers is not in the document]
eszthoff: This should also be destructured from props, so that it is not assigned to the `li` element. Otherwise we might make incorrect HTML, since `value` on `li` should be an integer (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li)

src/components/MultiActionButton/__tests__/MultiActionButton.spec.js
```diff
@@ -0,0 +1,130 @@
...
+                placement="top-start"
+            >
+                <ListItem key="disabled-key" disabled>
+                    Disabled
+                </ListItem>
+                <ListItem key="first-key" value="first-value">
+                    With value
+                </ListItem>
+            </MultiActionButton>
+        );
```
eszthoff, line 24: Wouldn't it be easier to mount the component in a `beforeEach` block, and only mount it inside `it` when it is different from the ususal?

src/utils/mergeRefs.ts
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+/**
+ * Merge all refs into one ref function
+ * @param refs - list of refs to be merged (skip undefined)
+ */
+export function mergeRefs<T>(
```
eszthoff, line 7: maybe should add some simple test to this function

src/utils/mergeRefs.ts
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
eszthoff, line 1: Nice!

src/components/MultiActionButton/MultiActionButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const { elem } = bem('MultiActionButton', styles);
+
+interface Props<V> extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange'> {
+    /**
+     *  Popup content to be placed inside List.
+     *  Navigation available only if element has not empty `value` attribute
+     *  and empty/false `disabled` attribute.
+     *  Use {@link ListItem} component.
+     */
```
eszthoff, line 20: Nice! 
It is good to add this to the story too, so we can see it in storybook. For examples see stories with `info` section.

eszthoff (changes requested):

src/components/List/ListItem/ListItem.tsx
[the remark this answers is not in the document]
agorodetskaya: It should be integer only for ordered lists `<ol>` which we don't use, but yes we don't know for sure how this component can be used =)
If one day we need to read it from element we can use data attribute.

---

> I think unknown type is better fits for the value that can have different types

agree, just `unknown` looks weird for me

src/components/MultiActionButton/MultiActionButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+    children: React.ReactElement<ListItemProps> | React.ReactElement<ListItemProps>[];
+    /** {@link Button} element, controlled by current component */
+    button: React.FunctionComponentElement<ButtonProps>;
```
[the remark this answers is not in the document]
agorodetskaya: We need it to describe what type of element we expect. We have the same here [ButtonGroup props](https://github.com/textkernel/oneui/blob/master/src/components/ButtonGroup/ButtonGroup.tsx#L9)

First I was thinking about creating some custom type to show that we use such child props:  `value`, 'isHighlighted' and `disabled`
But then I decided just make reference to ListItem.

---
I open for suggestions if you have some

src/components/MultiActionButton/__tests__/MultiActionButton.spec.js
agorodetskaya: Agree! It is more obvious and you don't need to check if something was changed in component for current test

src/components/MultiActionButton/MultiActionButton.tsx
agorodetskaya: What exactly you want to add to stories?
Now we already have a story which shows available/not available for navigation items

src/utils/mergeRefs.ts
agorodetskaya: agree!

src/components/MultiActionButton/MultiActionButton.scss
```diff
@@ -0,0 +1,9 @@
+.MultiActionButton {
+    &__list {
+        background-color: var(--color-background);
+        border-radius: var(--border-radius);
+        border: 1px solid var(--color-neutral);
+        box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.1);
+        z-index: 999;
```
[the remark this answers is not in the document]
agorodetskaya: make sense!

src/components/MultiActionButton/MultiActionButton.tsx
eszthoff: I would add the whole thing, as a reminder when we want to use this component 3 month down the line, and we forgot. It is true that it is demonstrated in the story code, but it is nice to spell things out sometime, to understand what makes it tick. And you summed it up so nicely already.

carlobernardini:
I have a few questions
1. Have we considered the simple name "Dropdown"?
2. Does this work as expected (mainly in terms of styling) when wrapped inside a ButtonGroup? I guess that would make sense

RE point 2, it doesn't seem to get styled exactly like it should. That's because the rounded corners are applied to :last-child, which in this case it isn't (the hidden dropdown is the last child). We might be able to fix that by using :last-of-type.
Screenshot 2021-01-21 at 10 45 57

src/components/MultiActionButton/MultiActionButton.tsx
agorodetskaya: aaah, oki! I got it

mukiienko (approved):

eszthoff (approved):

PR #443, opened 2021-01-26
feat(BooleanQueryTokenizer): implement bag of words with excluded negations

antipin:

# Checklist

### DIFF
--- a/src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
+++ b/src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
@@ -15,6 +15,7 @@
  * form the keyword. For identifiers and some other tokens, the pattern is more complex
  * structure that is matched by many strings.
  */
+/* eslint-disable class-methods-use-this */
 
 export enum TokenType {
     unknown,
@@ -35,6 +36,9 @@ type TokenPattern = [
 type Token = {
     type: TokenType;
     lexeme: string;
+};
+type Extract = {
+    wordsAndPhrases: string[];
 };
 
 // Character classes
@@ -62,7 +66,7 @@ const separatorOrEnd = `(?:${separator}|$)`;
 const parentheses = '(?:\\(|\\)|\\[|\\]|\\{|\\}|\\<|\\>)';
 const wildcard = '\\*';
 const phrase = '(?:"[^"]*"|\'[^\']*\')';
-const simpleOperator = '(?:OR|AND|NEAR|NOT|\\&|\\!|\\|)';
+const simpleOperator = '(?:OR|AND|NEAR|NOT|\\&|\\|)';
 const andNotOperator = `AND${separator}NOT`;
 const aroundOperator = `AROUND${separator}\\d+`;
 
@@ -79,7 +83,7 @@ export class BooleanQueryTokenizer {
     private static readonly tokenPatterns: TokenPattern[] = [
         [TokenType.separator, new RegExp(`^(?:${separator}|$)`)],
         [TokenType.word, new RegExp(`^${word}`)],
-        [TokenType.modifier, new RegExp(`^${wordModifier}(?=${word})`)],
+        [TokenType.modifier, new RegExp(`^(?:${wordModifier})(?=${phrase}|\\(|${word})`)],
         [
             TokenType.wildcard,
             new RegExp(`^(?:${wildcard}(?=${word})|${wildcard}(?=${separatorOrEnd}))`),
@@ -88,8 +92,10 @@ export class BooleanQueryTokenizer {
         [TokenType.phrase, new RegExp(`^${phrase}`)],
         [
             TokenType.booleanOperator,
+            // Cases like `NOT word`, `NOT(group...`, `AND"a phrase"`
             new RegExp(
-                `^(?:${andNotOperator}|${aroundOperator}|${simpleOperator})(?=${separatorOrEnd})`
+                `^(?:${andNotOperator}|${aroundOperator}|${simpleOperator})` +
+                    `(?=${separatorOrEnd}|\\(|${phrase})`
             ),
         ],
         [TokenType.parentheses, new RegExp(`^${parentheses}`)],
@@ -99,11 +105,11 @@ export class BooleanQueryTokenizer {
      * Tokenizer splits an input text into meaningful chunks and labels each chunk according to its
      * lexical meaning, suchwise producing a list of tokens.
      */
-    public static tokenize(inputString: string): Token[] {
+    public tokenize(inputString: string): Token[] {
         const tokens: Token[] = [];
         let remainingInputString = inputString;
         while (remainingInputString.length > 0) {
-            const token = BooleanQueryTokenizer.tokenizeCurrentChunk(remainingInputString);
+            const token = this.tokenizeCurrentChunk(remainingInputString);
             if (token) {
                 // If token has been found, reduce the `remainingInputString `
                 // on the length of the matched lexeme and add the token to the result.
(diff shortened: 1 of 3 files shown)


src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
...
+                    lexeme: '',
+                };
+            }
+        });
+        return tokens;
+    }
+
+    private static isNegation(token: Token): boolean {
+        return (
+            (token.type === TokenType.booleanOperator &&
```
agorodetskaya, line 228: I understand that prettier configured to have such code style, but for me looks hard to read.
Something like this is much more obvious:
```suggestion
            (
                token.type === TokenType.booleanOperator
                && (token.lexeme === 'NOT' || token.lexeme === 'AND NOT')
            )
            || (
                token.type === TokenType.modifier
                && (token.lexeme === '-' || token.lexeme === '!' )
            )
```

or

```suggestion
            (token.type === TokenType.booleanOperator
                && (token.lexeme === 'NOT' || token.lexeme === 'AND NOT'))
            || (token.type === TokenType.modifier
                && (token.lexeme === '-' || token.lexeme === '!' ))
```
---
This is not remark to current PR but just point to think about code style rules

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
...
+                (token.lexeme === 'NOT' || token.lexeme === 'AND NOT')) ||
+            (token.type === TokenType.modifier && (token.lexeme === '-' || token.lexeme === '!'))
```
agorodetskaya, line 230: This check doesn't cover case:
`java AND js  -"Spring Web Flux"`
because in this case `-` token is 
```
{
   "type": 0,
   "lexeme": "-"
}
```
---
This minus phrase is used for search in google, so i guess it could be used for google based searchers in Reach

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
...
+            })
+            // After `phrase` tokens preprocessing, some strings can become empty,
+            // so we should remove them.
+            .filter((str) => str);
+        return {
+            wordsAndPhrases,
+        };
+    }
+
+    private static replaceTokens(tokens: Token[], from: number, to: number): Token[] {
```
agorodetskaya, line 213: I think `replaceTokens ` name is not obvious, because we do not understand to what we are going to replace. You have to look at implementation to see that we replace _from-to_ tokens to `unknown`.

My first thoughts were - we swap to tokens.

Maybe something like `resetSubsequenceToUnknown`

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -129,6 +133,9 @@ export class BooleanQueryTokenizer {
         return tokens;
     }
 
+    /**
+     * @deprecated use BooleanQueryTokenizer#extractWords instead
```
agorodetskaya, line 137: Storybook also should be updated

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
...
+            }
+        }
+        // Filter only words and phrases
+        const wordsAndPhrases = tokens
+            .filter((token) => token.type === TokenType.word || token.type === TokenType.phrase)
+            .map((wordOrPhrase) => {
+                // `phrase` tokens require some dirty preprocessing
+                // to remove quotes and possible inner wildcards.
+                if (wordOrPhrase.type === TokenType.phrase) {
+                    return wordOrPhrase.lexeme.replace(/('|"|\*)/g, '').trim();
```
agorodetskaya, line 201: I think we need a simple test for this

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
...
+                            const parenthesisLoopToken = tokens[k];
+                            if (parenthesisLoopToken.type === TokenType.parentheses) {
+                                if (parenthesisLoopToken.lexeme === '(') {
+                                    parenthesisTracker += 1;
+                                } else if (parenthesisLoopToken.lexeme === ')') {
+                                    parenthesisTracker -= 1;
+                                }
+                            }
+                            if (parenthesisTracker === 0) {
+                                BooleanQueryTokenizer.replaceTokens(tokens, i, k);
```
agorodetskaya, line 183: I think at this moment we also can do `i = k`, because we do need to go through reseted tokens again in main loop.

The same in line 169 we can de `i = j`

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
...
+
+    private static isNegationable(token: Token): boolean {
+        return (
+            token.type === TokenType.word ||
+            token.type === TokenType.wildcardWord ||
+            token.type === TokenType.phrase
+        );
+    }
+
+    private static isParenthesisGroup(token: Token): boolean {
```
agorodetskaya, line 242: I think `isParenthesisGroupStart` name is more suitable here, because we are checking exactly start of group

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -148,6 +155,94 @@ export class BooleanQueryTokenizer {
         );
     }
 
+    public static extractWords(inputString: string): Extract {
```
agorodetskaya, line 158: I think we need to add doc here or on result type to describe that we are going to exclude negationed words and phrases

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
```diff
@@ -36,6 +36,9 @@ type Token = {
     type: TokenType;
     lexeme: string;
 };
+type Extract = {
+    wordsAndPhrases: string[];
```
agorodetskaya, line 40: Why do you think we need result as object and not just array of strings?

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
agorodetskaya, line 158: I don't know if it was discussed, but for now we just ignore invalid negations like
```
java AND NOT (js OR python
```
now it will return: `["java", "js", "python"]`

And this is fine because we don't know what exactly user wanted to do and where he made a mistake.
Just I think we need to mention in doc that we ignore invalid negations

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: That's because we are planning to extend the result of this function in the future.
Besides `wordsAndPhrases` we will need to return wildcard words and phrases separately.

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: > because we do need to go through

you mean "we do NOT need to go... ?"

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: good point

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: I also struggled with prettifier, doesn't make much sense to me, but let's change it in a separate MR

Created a task https://jira.textkernel.nl/browse/ONEUI-235

src/packages/BooleanQueryTokenizer/BooleanQueryTokenizer.ts
antipin: nice optimisation anyways :)

agorodetskaya: > you mean "we do NOT need to go... ?"

yes =)

agorodetskaya (approved):

PR #449, opened 2021-02-18
feat(ProductTour): new component

eszthoff:

# Checklist

### DIFF
--- a/src/components/ProductTour/ProductTour.scss
+++ b/src/components/ProductTour/ProductTour.scss
@@ -37,6 +37,7 @@
         &--isActive {
             background-color: var(--color-neutral-60);
             cursor: default;
+            outline: none;
         }
 
         & + .ProductTour__dot {
--- a/src/components/ProductTour/ProductTour.scss
+++ b/src/components/ProductTour/ProductTour.scss
@@ -1,0 +1,3 @@
+.ProductTour {
+    display: flex;
+}
--- a/src/components/ProductTour/ProductTour.scss
+++ b/src/components/ProductTour/ProductTour.scss
@@ -1,3 +1,51 @@
 .ProductTour {
-    display: flex;
+    height: 600px;
+    width: 800px;
+
+    &__slide {
+        padding: 25px;
+        height: 540px;
+        overflow: auto;
+        box-sizing: border-box;
+    }
+
+    &__footer {
+        position: absolute;
+        bottom: 0;
+        box-sizing: border-box;
+        width: 100%;
+        height: 60px;
+        padding: 0 25px;
+        border-top: 1px solid var(--border-color-light);
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+    }
+
+    &__checkbox {
+        flex-basis: 40%;
+    }
+
+    &__dot {
+        display: inline-block;
+        width: 10px;
+        height: 10px;
+        border-radius: 50%;
+        background-color: var(--color-neutral-25);
+        cursor: pointer;
+
+        &--isActive {
+            background-color: var(--color-neutral-60);
+            cursor: default;
+        }
+
+        & + .ProductTour__dot {
+            margin-left: 10px;
+        }
+    }
+
+    &__buttons {
+        flex-basis: 40%;
+        text-align: end;
+    }
 }
--- a/src/components/ProductTour/ProductTour.tsx
+++ b/src/components/ProductTour/ProductTour.tsx
@@ -32,8 +32,8 @@ export const ProductTour: React.FC<Props> = (props) => {
         isOpen,
         children,
         checkboxLabel,
-        cancelLabel,
-        continueLabel,
+        cancelLabel = '',
+        continueLabel = '',
         finishLabel,
         onCancel,
         onFinished,
@@ -135,10 +135,10 @@ export const ProductTour: React.FC<Props> = (props) => {
                     ) : (
                         <>
                             <Button context="link" onClick={handleCancel}>
-                                {cancelLabel || ''}
+                                {cancelLabel}
                             </Button>
                             <Button context="brand" onClick={handleNext}>
-                                {continueLabel || ''}
+                                {continueLabel}
                             </Button>
                         </>
                     )}
--- a/src/components/ProductTour/ProductTour.tsx
+++ b/src/components/ProductTour/ProductTour.tsx
@@ -9,6 +9,12 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
     children: React.ReactElement[];
     /** The label for the checkbox. If undefined or empty, checkbox will not be shown */
     checkboxLabel?: string;
+    /** Cancel button label - required if there are is then 1 slide */
+    cancelLabel?: string;
+    /** Continue button label - required if there are is then 1 slide */
+    continueLabel?: string;
+    /** Finish button label */
+    finishLabel: string;
     /** A callback if the user clicked the Cancel button without finishing the tour */
     onCancel?: (isChecked?: boolean) => void;
     /** A callback if the user clicked the Finish button */
--- a/src/components/ProductTour/ProductTour.tsx
+++ b/src/components/ProductTour/ProductTour.tsx
@@ -1,0 +1,28 @@
+import * as React from 'react';
+import { bem } from '../../utils';
+import styles from './ProductTour.scss';
+
+interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** Should the tour be shown */
+    isOpen: boolean;
+    /** An array of ReactElement where each element is one slide */
+    children: React.ReactElement[];
+    /** The label for the checkbox. If undefined or empty, checkbox will not be shown */
+    checkboxLabel?: string;
+    /** A callback if the user clicked the Cancel button without finishing the tour */
+    onCancel?: (isChecked?: boolean) => void;
+    /** A callback if the user clicked the Finish button */
+    onFinished: (isChecked?: boolean) => void;
+}
+
+const { block, elem } = bem('ProductTour', styles);
+
+export const ProductTour: React.FC<Props> = (props) => {
+    return null;
+};
+
+ProductTour.displayName = 'ProductTour';
+
+ProductTour.defaultProps = {
+    isOpen: false,
+};
--- a/src/components/ProductTour/ProductTour.tsx
+++ b/src/components/ProductTour/ProductTour.tsx
@@ -1,17 +1,21 @@
 import * as React from 'react';
 import { bem } from '../../utils';
+import { ENTER_KEY } from '../../constants';
+import { Button } from '../Buttons';
+import { Checkbox } from '../Checkbox';
+import { Modal, ModalProps } from '../Modal';
 import styles from './ProductTour.scss';
 
-interface Props extends React.HTMLAttributes<HTMLDivElement> {
+interface Props extends ModalProps {
     /** Should the tour be shown */
     isOpen: boolean;
     /** An array of ReactElement where each element is one slide */
-    children: React.ReactElement[];
+    children: React.ReactElement | React.ReactElement[];
     /** The label for the checkbox. If undefined or empty, checkbox will not be shown */
     checkboxLabel?: string;
-    /** Cancel button label - required if there are is then 1 slide */
+    /** Cancel button label - required if there is more then 1 slide */
     cancelLabel?: string;
-    /** Continue button label - required if there are is then 1 slide */
+    /** Continue button label - required if there is more then 1 slide */
     continueLabel?: string;
     /** Finish button label */
     finishLabel: string;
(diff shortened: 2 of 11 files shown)


src/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import { bem } from '../../utils';
+import styles from './ProductTour.scss';
+
+interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** Should the tour be shown */
+    isOpen: boolean;
+    /** An array of ReactElement where each element is one slide */
+    children: React.ReactElement[];
```
carlobernardini, line 9: Would it be an idea to constrain it to be of HTML type `section`, so that it will throw a warning when using something that isn't suitable (like `span` or whatever - now you could in theory use `button` as a slide container)

carlobernardini:
How does your API deal with the JF use case for remembering not to show this modal again? I see how to render that checkbox but how do you intend to deal with the business logic.

eszthoff:
> How does your API deal with the JF use case for remembering not to show this modal again? I see how to render that checkbox but how do you intend to deal with the business logic.

It is up to the application. The state of the checkbox is passed to the application during the `onCancel` and `onFinish` callback. The application then can decide what it wants to do, e.g. save in local storage, put in the backend, etc. Does that answer your question?

src/components/ProductTour/ProductTour.tsx
eszthoff: I'm not sure how useful that is, and how easy is it to implement. E.g. how to validate when the slide is an other React Element, e.g. `MySlide` that in turn can return different top elements, or even a fragment?

src/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,34 @@
...
+    cancelLabel?: string;
+    /** Continue button label - required if there are is then 1 slide */
+    continueLabel?: string;
+    /** Finish button label */
+    finishLabel: string;
+    /** A callback if the user clicked the Cancel button without finishing the tour */
+    onCancel?: (isChecked?: boolean) => void;
+    /** A callback if the user clicked the Finish button */
+    onFinished: (isChecked?: boolean) => void;
+}
```
mukiienko, line 22: I'm wondering how I can go one step back if I by mistake to push the "continue" button? @carlobernardini

src/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,34 @@
...
+    /** The label for the checkbox. If undefined or empty, checkbox will not be shown */
+    checkboxLabel?: string;
+    /** Cancel button label - required if there are is then 1 slide */
+    cancelLabel?: string;
+    /** Continue button label - required if there are is then 1 slide */
+    continueLabel?: string;
+    /** Finish button label */
+    finishLabel: string;
+    /** A callback if the user clicked the Cancel button without finishing the tour */
+    onCancel?: (isChecked?: boolean) => void;
```
mukiienko, line 19: What do you think about passing the last viewed step in the second argument? Actually, I don't know why we need this information :) Maybe we can use that information for analytics or we can start  `ProductTour` from that step when the user will back the next time.

src/components/ProductTour/ProductTour.tsx
carlobernardini: I see. Then perhaps it might be a good idea to include some `<Slide />` component ourselves, because then we're sure a correct element will be used as wrapper and we retain full control over how that content wrapper is to be styled within the context of OneUI rather than the application.
For example, say we decide to restyle this component some time in the future, and the content needs more padding. Then we simply change that in OneUI without having to adapt applications that use this component.

src/components/ProductTour/ProductTour.tsx
carlobernardini: There are dots in the footer that each represent a slide. You can click on any of these dots to move back and forth to any slide you wish.

Screenshot 2021-02-23 at 08 11 01

src/components/ProductTour/ProductTour.tsx
eszthoff: I was thinking to just wrap the children, whomever they are in `<section>` and add styling there. Do you think that would work?

src/components/ProductTour/ProductTour.tsx
eszthoff: I think it is a good idea, but I would add it when we have a usecase for it. Then we are sure we are doing what we need. E.g. if you want to follow on analytics, you might want to log all events of clicking on any of the buttons. In that case we should add an `onContinue`, and for the dots we could have an `onDotClick` params. But in the usecase you have, we would need to change the args on the current callback. I think they are all valid use cases, let's see which one we will need once there are requirements for it.

src/components/ProductTour/ProductTour.tsx
mukiienko: So these dots should be clickable. It makes sense. According to the story, they only indicate a step, and navigation should be done on the right side of the bottom.

carlobernardini: Yes, I forgot to describe this functionality. My mistake.

src/components/ProductTour/ProductTour.tsx
carlobernardini: Yes, that should indeed work too. I was just thinking if this API would be a bit too flexible but maybe it's fine to keep it simple.

src/components/ProductTour/ProductTour.tsx
eszthoff: Speaking of which... do we need additional styling on hover (apart from `cursor: pointer`)?

src/components/ProductTour/ProductTour.tsx
eszthoff: Could you add some spacing sizes to the ticket description? Both for the content, and maybe for the dots as well. Thanks

src/components/ProductTour/ProductTour.tsx
carlobernardini: No, it's fine as is (except indeed for the pointer)

src/components/ProductTour/ProductTour.tsx
carlobernardini: Slides have a padding of 25px all around (you don't necessarily need to use OneUI spacing variables in this case)
Dots are positioned in the middle / center of the footer, sizes are already described in the ticket: 10x10px
The footer in the design is 60px high (padding around the button is 15px)

stories/ProductTour.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
+                    incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
+                    nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+                    Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
+                    fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
+                    culpa qui officia deserunt mollit anim id est laborum.
+                </p>
+            </div>,
+        ];
+        const numberOfSlides = select('Number of slides', [1, 2, 3, 4], 4);
```
carlobernardini, line 62: https://github.com/storybookjs/storybook/tree/master/addons/knobs#number-bound-by-range

stories/ProductTour.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <ProductTour
+                isOpen={boolean('Show tour', true)}
+                checkboxLabel={text('Checkbox label', "Don't show again")}
+                cancelLabel={text('Cancel button label', 'Skip tour')}
+                continueLabel={text('Continue buttom label', 'Next')}
+                finishLabel={text('Finishe buttom label', 'Finish')}
+                onCancel={handleCancel}
+                onFinished={handleFinished}
+            >
+                {slides.slice(0, numberOfSlides)}
```
carlobernardini, line 74: I think this can be simplified:

`{ Array(numberOfSlides).fill(0).map((_, i) => <div key={`slide-${ i }`}>Lipsum...</div>) }`

eszthoff: Not sure why is it simple. I also changed it now, so that different slides are a bit different to see what happens if the text is too long for the fixed size window. In addition the mock is reused in tests.

src/components/ProductTour/ProductTour.scss
```diff
@@ -0,0 +1,51 @@
...
+    &__slide {
+        padding: 25px;
+        height: 540px;
+        overflow: auto;
+        box-sizing: border-box;
+    }
+
+    &__footer {
+        position: absolute;
+        bottom: 0;
```
carlobernardini, line 14: Not sure that `position: absolute`, `bottom: 0` and `width: 100%` are necessary.
The previous sibling will already push the footer down since it has a fixed height.
And the footer seems to be a block level element, so it should take up full width without explicit definition.

stories/ProductTour.tsx
```diff
@@ -0,0 +1,40 @@
...
+        const numberOfSlides = number('Number of slides', 4, {
+            range: true,
+            min: 1,
+            max: 4,
+            step: 1,
+        });
+        const isMultiSlide = numberOfSlides > 1;
+
+        return (
+            <ProductTour
```
carlobernardini, line 26: Maybe consider making it possible to dismiss the modal in the story, otherwise I can't read the documentation behind the backdrop ;)

src/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const handleCancel = () => {
+        onCancel?.(isChecked);
+    };
+
+    const handleNext = () => {
+        setCurrentSlide(currentSlide + 1);
+    };
+
+    const handleFinish = () => {
+        onFinished(isChecked);
```
carlobernardini, line 81: Very much of a corner case but still FYI:
Consider you have checked the checkbox, then prop `checkboxLabel` is changed to `undefined`, `isChecked` will still be `true` while there is no checkbox.

src/components/ProductTour/ProductTour.scss
eszthoff: I had it without it, and this solution seemed to me a pix better, but indeed it would work without it as well

stories/ProductTour.tsx
eszthoff: You can :-) just click the `show tour` checkbox in the knobs. I think adding an actual implementation story for it will be misleading a bit - it is clearer this way that you have to control it from the outside.

src/components/ProductTour/ProductTour.tsx
eszthoff: Indeed an edge case. I can add code to not return anything if there is no checkbox. For now I assumed the application knows if they want to read this variable or not. It might make it easier for TS if the def always have the (last) value of the checkbox sent over. But if you think there is real danger here, I can remove it.

carlobernardini: No danger :) just so you know

src/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,155 @@
...
+            handleCancel();
+        }
+    };
+
+    return (
+        <Modal
+            isOpen={isOpen}
+            isPositionFixed
+            {...rest}
+            {...block()}
```
carlobernardini, line 97: Should this not be `{...block(props)}`?

igorarkhipenko:
There's a blue outline around the dot once it's clicked. I think it should be removed.
image

src/components/ProductTour/ProductTour.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        <Button context="primary" onClick={handleFinish}>
+                            {finishLabel}
+                        </Button>
+                    ) : (
+                        <>
+                            <Button context="link" onClick={handleCancel}>
+                                {cancelLabel || ''}
+                            </Button>
+                            <Button context="brand" onClick={handleNext}>
+                                {continueLabel || ''}
```
igorarkhipenko, line 141: I think these two label values can be set as default parameters during `props` object deconstruction

mukiienko (approved):

src/components/ProductTour/ProductTour.tsx
eszthoff: Normally yes. But with modal classname is used a bit differently, so no need. They are passed with `{...rest}`.

eszthoff:
> There's a blue outline around the dot once it's clicked. I think it should be removed.
> image

It does help with keyboard navigation, otherwise you don't know you are on the button. As far as I can tell, it is only visible during the click, so I think it is fine. @carlobernardini ?

igorarkhipenko:
> > There's a blue outline around the dot once it's clicked. I think it should be removed.
> > image
> 
> It does help with keyboard navigation, otherwise you don't know you are on the button. As far as I can tell, it is only visible during the click, so I think it is fine. @carlobernardini ?

For me it remains there even after the click is done. I agree it might get in handy in terms of accessibility but I'm just checking since it's something which was not in the original AC.
BTW I also noticed that switching the focus between dots by pressing Tab behaves a bit oddly. It resets the position of the next dot-to-be-focused each time the selection is made. This is something out of the scope for sure but maybe to consider together with outlining.

carlobernardini:
The outline can stay, it's actually an accessibility best practice. It remains after the click because the element is still focused.
I can't reproduce the positioning issue you describe.

igorarkhipenko:
Nevermind, it just disappeared for me as well while I was trying to record a demo.

igorarkhipenko:
Ooh, now I see — I was testing it in Safari and there you can (hopefully) reproduce it. But I'm not sure how crucial it is for now.

PR #450, opened 2021-02-18
feat(Toggle): new component

agorodetskaya:
Add a new Toggle component.
Related story: https://jira.textkernel.nl/browse/ONEUI-241

# Checklist

### DIFF
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -53,7 +53,7 @@
     }
 
     &__text {
-        padding-left: var(--spacing-2x);
+        padding: var(--spacing-normal) 0 var(--spacing-normal) var(--spacing-2x);
     }
 
     &--disabled {
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -1,0 +1,64 @@
+@import '../../themes/oneui/oneui.scss';
+
+.Toggle {
+    display: flex;
+    align-items: center;
+    cursor: pointer;
+
+    &__toggleContainer {
+        width: 36px;
+        height: 20px;
+        border-radius: 20px;
+        position: relative;
+        background-color: var(--color-neutral);
+        transition: border-color 300ms;
+
+        .Toggle__realCheckbox {
+            position: absolute;
+            clip: rect(0,0,0,0);
+
+            & + .Toggle__toggleButton {
+                position: absolute;
+                top: 0;
+                left: 0;
+                right: 0;
+                bottom: 0;
+                border-radius: 20px;
+                transition: all 300ms;
+
+                &::before {
+                    content: '';
+                    display: inline-block;
+                    width: 18px;
+                    height: 18px;
+                    margin: 1px;
+                    background-color: white;
+                    border-radius: 50%;
+                    box-shadow: 1px 1px 3px rgba(0,0,0,0.5) ;
+                    transition: all 300ms ease-in-out;
+                }
+
+            }
+
+            &:checked + .Toggle__toggleButton {
+                background-color: var(--color-good-40);
+                &::before{
+                    margin-left: 17px;
+                }
+            }
+
+            &:disabled + .Toggle__toggleButton {
+                background-color: var(--color-neutral);
+            }
+        }
+    }
+
+    &__text {
+        padding-left: var(--spacing-normal);
+    }
+
+    &--disabled {
+        cursor: not-allowed;
+    }
+}
+
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -34,7 +34,6 @@
                     margin: 1px;
                     background-color: white;
                     border-radius: 50%;
-                    box-shadow: 1px 1px 3px rgba(0,0,0,0.5) ;
                     transition: all 300ms ease-in-out;
                 }
 
@@ -54,7 +53,7 @@
     }
 
     &__text {
-        padding-left: var(--spacing-normal);
+        padding-left: var(--spacing-2x);
     }
 
     &--disabled {
--- a/src/components/Toggle/Toggle.tsx
+++ b/src/components/Toggle/Toggle.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+import { bem } from '../../utils';
+import { Text } from '../Text';
+import styles from './Toggle.scss';
+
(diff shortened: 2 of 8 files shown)


src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const { block, elem } = bem('Toggle', styles);
+
+export const Toggle: React.FC<Props> = React.forwardRef((props, ref) => {
+    const { children, disabled, id, ...rest } = props;
+    return (
+        <label {...block(props)} htmlFor={id}>
+            <div {...elem('toggleContainer')}>
+                <input
```
carlobernardini, line 24: Consider adding  `role="switch"` and `aria-checked="true/false"` for accessibility

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            &:disabled + .Toggle__toggleButton {
+                background-color: var(--color-neutral);
+            }
+        }
+    }
+
+    &__text {
+        padding-left: var(--spacing-normal);
```
carlobernardini, line 57: The spacing on the left of a Checkbox label is `--spacing-2x`

carlobernardini:
When the toggle is switched on _and_ disabled, it looks like it's switched off. Can you apply `--color-green-25` when it's on + disabled?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,64 @@
...
+
+                &::before {
+                    content: '';
+                    display: inline-block;
+                    width: 18px;
+                    height: 18px;
+                    margin: 1px;
+                    background-color: white;
+                    border-radius: 50%;
+                    box-shadow: 1px 1px 3px rgba(0,0,0,0.5) ;
```
carlobernardini, line 37: I think it looks better without box shadow :)

agorodetskaya: as you wish =)

src/components/Toggle/Toggle.tsx
agorodetskaya: hm, why do we need this if we already have input with type checkbox?
we can switch on/off by Space.
Another question is that we do not see that toggle is focused. But I found the same with Checkbox component, so postponed it for a while to ask you.

agorodetskaya:
> When the toggle is switched on and disabled, it looks like it's switched off. Can you apply --color-green-25 when it's on + disabled?

sure, I can!
but again I did the same as we have in Checkbox (color natural if it checked), but yeah in Checkbox it is more obvious that it is checked =)

src/components/Toggle/Toggle.tsx
carlobernardini: As far as I can tell it's not exactly the same as checkbox:
>The ARIA switch role is functionally identical to the checkbox role, except that instead of representing "checked" and "unchecked" states, which are fairly generic in meaning, the switch role represents the states "on" and "off."

But it's not super relevant to us.

With regard to focus state: good catch, I'll have a look.

carlobernardini:
>but again I did the same as we have in Checkbox
While indeed it's more obvious for Checkbox, I think we can do the same there. I'll look into it + focus state.

src/components/Toggle/Toggle.tsx
agorodetskaya: Hi @carlobernardini !
> With regard to focus state: good catch, I'll have a look.

Maybe we can do this focusing functionality (if needed) in dedicated story for both: Toggle and Checkbox. What do you think?

carlobernardini: Sounds good to me 👍

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,47 @@
...
+    /** Ref to access the toggle input element */
+    ref?: React.RefObject<HTMLInputElement>;
+}
+
+const { block, elem } = bem('Toggle', styles);
+
+export const Toggle: React.FC<Props> = React.forwardRef((props, ref) => {
+    const { children, disabled, id, ...rest } = props;
+    return (
+        <label {...block(props)} htmlFor={id}>
```
mukiienko, line 22: It is minor, but we can rid of `htmlFor`, because the input is explicitly associated with the label when it inside the label.

stories/Toggle.js
```diff
@@ -0,0 +1,27 @@
...
+    .addDecorator(withKnobs)
+    .add('Toggle with label', () => (
+        <Toggle
+            disabled={boolean('disabled', false)}
+            id={text('id', 'my-toggle-1')}
+            onChange={(event) => {
+                console.log('Toggle state changed. event.target.checked: ', event.target.checked);
+            }}
+        >
+            This is title!
```
mukiienko, line 16: It would be nice to have the label in knobs as well

mukiienko:
For me, Toogle label looks more aligned without `align-items: center;`

mukiienko:
Animation doesn't work smoothly the first time:

https://user-images.githubusercontent.com/19436304/109417430-f327fd80-79c3-11eb-850b-bd34c54467ee.mov

agorodetskaya:
> For me, Toogle label looks more aligned without `align-items: center;`

-----
@mukiienko 
Well, for me it is more aligned with `align-items: center;`

https://user-images.githubusercontent.com/17009609/109478549-ae5e9e00-7a79-11eb-8f1a-331ff9e0d0bd.mov


@carlobernardini what do you think?

src/components/Toggle/Toggle.tsx
agorodetskaya: without this we have an eslint error

agorodetskaya:
> Animation doesn't work smoothly the first time:

@mukiienko interesting. For me it works smoothly in all browsers.

https://user-images.githubusercontent.com/17009609/109479610-d5699f80-7a7a-11eb-9eeb-5bed728ebd85.mov

What browser and what version do you use?

carlobernardini:
@agorodetskaya For me it also seems to be better aligned without `align-items: center` in Chrome. In Safari I'm not sure, but the difference is more subtle.

What seems to work for me is when I apply the same styling as for Checkbox:
1. Apply `var(--spacing-normal) 0 var(--spacing-normal) var(--spacing-2x)` to the label text.
2. Keep `align-items: center`.

PR #493, opened 2021-04-29
feat(Dropdown): add Dropdown improvements

mukiienko:
BREAKING CHANGE: Dropdown children is function instead of component. Introduced new props `items`, `onOpen`, `onClose`, `isOpen`, `itemToString` in favor to make Dropdown as controlled component.

Story: ONEUI-265

# Checklist

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,45 +1,67 @@
 import React, { useState } from 'react';
-import { useSelect } from 'downshift';
+import { useSelect, PropGetters } from 'downshift';
 import { usePopper } from 'react-popper';
 import { bem } from '../../utils/bem/bem';
 import { mergeRefs } from '../../utils/mergeRefs';
 import { PopupPlacement } from '../../constants';
 import { ButtonProps } from '../Buttons';
-import { ListItemProps } from '../List/ListItem';
 import { List } from '../List';
 import styles from './Dropdown.scss';
 
 const { elem } = bem('Dropdown', styles);
 
+export type DropdownRenderArgs<V> = {
+    getItemPropsByIndex: (index: number) => PropGetters<V>;
+};
+
 interface Props<V> extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange'> {
-    /**
-     *  Popup content to be placed inside List.
-     *  Navigation available only if element has not empty `value` attribute
-     *  and empty/false `disabled` attribute.
-     *  Use {@link ListItem} component.
-     */
-    children:
-        | React.ReactElement<ListItemProps>
-        | React.ReactElement<ListItemProps>[]
-        | (React.ReactElement<ListItemProps> | React.ReactElement<ListItemProps>[])[];
+    /** List of items */
+    items: V[];
+    /** Popup content */
+    children: (args: DropdownRenderArgs<V>) => React.ReactElement;
     /** {@link Button} element, controlled by current component */
     button: React.FunctionComponentElement<ButtonProps>;
-    /**
-     * Callback called on selecting one of the passed as children items.
-     * Value parameter it is a `value` attribute of children item ({@link ListItemProps.value}).
-     */
+    /** onChange called when an item is selected */
     onChange: (value: V) => void;
+    /** onOpen called on request to be open */
+    onOpen?: () => void;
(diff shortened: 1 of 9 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -91,6 +111,16 @@ export function Dropdown<V>(props: Props<V>) {
         ...state.attributes.popper,
         ...elem('list'),
     };
+    const getItemPropsByIndex = (index: number) => {
+        return {
+            ...getItemProps({
+                index,
+                item: items[index],
+            }),
+            isHighlighted: highlightedIndex === index,
```
eszthoff, line 120: Is this a ListItem specific prop? What happens when it is used on other components, as per your AC?

eszthoff (changes requested):

src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,45 +1,67 @@
...
-     *  Popup content to be placed inside List.
-     *  Navigation available only if element has not empty `value` attribute
-     *  and empty/false `disabled` attribute.
-     *  Use {@link ListItem} component.
-     */
-    children:
-        | React.ReactElement<ListItemProps>
-        | React.ReactElement<ListItemProps>[]
-        | (React.ReactElement<ListItemProps> | React.ReactElement<ListItemProps>[])[];
+    /** List of items */
```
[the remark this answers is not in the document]
agorodetskaya: Agree that it is not clear that it should be list of navigatable values

src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,45 +1,67 @@
...
+    isOpen?: boolean;
     /**
-     * Callback called on selecting one of the passed as children items.
-     * Value parameter it is a `value` attribute of children item ({@link ListItemProps.value}).
+     * Convert item to string.
+     * Prop is required if passed items are not list of strings.
+     * This is need for accessibility aria-live messages for downshift.
      */
-    onChange: (value: V) => void;
+    itemToString?: (item: V | null) => string;
```
agorodetskaya, line 37: This prop confuses me.
We have to implement list rendering on our own, but also send items list and this prop... Why do we need it if do everything on our own?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,45 +1,67 @@
 import React, { useState } from 'react';
```
agorodetskaya, line 1: I think current implementation can cause extra work if we have grouped list like:
image
because `getItemPropsByIndex` now requires index shifted on prev group size.
Example from stories:
```
const customValues1 = [
    'ListItem with value 1',
    'ListItem with value 2',
    'ListItem with value 3',
];
const customValues2 = [
    'ListItem with value 1-2',
    'ListItem with value 2-2',
    'ListItem with value 3-2',
];
return (
    <div style={styles.content}>
        <Dropdown<string>
            items={customValues1.concat(customValues2)}
            ...
        >
            {({ getItemPropsByIndex }) => {
                return (
                    <>
                        <ListItem disabled>Disabled 1</ListItem>
                        {customValues1.map((value, index) => (
                            <ListItem key={value} {...getItemPropsByIndex(index)}>
                                <div style={styles.listItem}>
                                    <IconTextkernel context="brand" style={styles.icon} />
                                    {value}
                                </div>
                            </ListItem>
                        ))}
                        <ListItem disabled>Disabled 2</ListItem>
                        {customValues2.map((value, index) => (
                            <ListItem key={value} {...getItemPropsByIndex(index + customValues1.length)}>
                                <div style={styles.listItem}>
                                    <IconTextkernel context="brand" style={styles.icon} />
                                    {value}
                                </div>
                            </ListItem>
                        ))}
                        <div style={styles.customDiv}>Just custom div element</div>
                    </>
                );
            }}
        </Dropdown>
    </div>
```

agorodetskaya, line 1: If we need possibility to manually close Dropdown - maybe it will be better to just add public `close` method which you can call using ref?

mukiienko:
We decided to postpone with improvements in favor of coming up with a better idea of how it can be implemented without breaking changes.

mukiienko:
We decided to postpone with improvements in favor of coming up with a better idea of how it can be implemented without breaking changes.

PR #495, opened 2021-04-30
feat(Autosuggest, Combobox): add prop to set HTML attributes on input element

eszthoff:

# Checklist

### DIFF
--- a/src/components/SelectComponents/Autosuggest/Autosuggest.tsx
+++ b/src/components/SelectComponents/Autosuggest/Autosuggest.tsx
@@ -43,6 +43,8 @@ interface Props<S>
     /** Means to overwrite the look and feel of the UI in its blurred state.
      * It should be null or undefined if there is no selection yet */
     customSelectionIndicator?: ReactNode;
+    /** Additional HTML attributes to be applied to the input element */
+    inputAttr?: DictionaryOf<string | boolean>;
 }
 
 const { elem } = bem('Autosuggest', styles);
@@ -71,6 +73,7 @@ export function Autosuggest<S>(props: Props<S>) {
         inputRef: inputRefFromProps,
         customSelectionIndicator,
         initInputValue,
+        inputAttr,
         ...rest
     } = props;
     const inputRef = inputRefFromProps || React.createRef<HTMLInputElement>();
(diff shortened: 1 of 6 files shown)


src/components/SelectComponents/Autosuggest/Autosuggest.tsx
```diff
@@ -151,6 +154,7 @@ export function Autosuggest<S>(props: Props<S>) {
                     onFocus: onFocusInput,
                     onKeyDown: (e) => handleInputKeyDown(e, highlightedIndex),
                     'data-lpignore': true,
+                    ...inputAttr,
```
carlobernardini, line 157: I wonder, does this sequence of attributes make sense? You could now theoretically overwrite previous attributes like `id` etc. Could that break the component?

src/components/SelectComponents/Autosuggest/Autosuggest.tsx
```diff
@@ -43,6 +43,8 @@ interface Props<S>
     /** Means to overwrite the look and feel of the UI in its blurred state.
      * It should be null or undefined if there is no selection yet */
     customSelectionIndicator?: ReactNode;
+    /** Additional HTML attributes to be applied to the input element */
+    inputAttr?: DictionaryOf<string | boolean>;
```
mukiienko, line 47: Let's use `React.InputHTMLAttributes<HTMLInputElement>`

mukiienko, line 47: What do you think about renaming `inputAttr`  -> `inputProps`. It will be a more React-friendly way I think.

eszthoff: I tried but that won't allow for arbitrary `data-something` attributes which is what we need here

mukiienko: I thought `React.InputHTMLAttributes<HTMLInputElement>` supports data attribute :( 
I found some solution, but it seems only proposal https://github.com/microsoft/TypeScript/pull/21044.
Why do we need to support boolean?

mukiienko: I got your point. What about adding `s` `inputAttrs` ?

eszthoff: boolean for data attributes... I don't think there is a simple way to pass them without adding the value

mukiienko (approved):

PR #500, opened 2021-05-05
feat(Map, LocationSelector): add option to have a default highlighted area on the map

eszthoff:
Story: [JF-2786](https://jira.textkernel.nl/browse/JF-2786)

# Checklist


src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.js
```diff
@@ -217,6 +222,7 @@ LocationSelectorDialog.propTypes = {
     getMarkers: PropTypes.func.isRequired,
     /** function to be called when teh Done button is clicked */
     onCloseModal: PropTypes.func,
+    defaultHighlight: PropTypes.object,
```
carlobernardini, line 225: What's the difference with `defaultArea`?

eszthoff: I added the description in Map (will add it everywhere if the direction is approved)
```
    /** The default parameters to determine the viewport when no markers are present. */
    defaultArea: PropTypes.oneOfType([
        PropTypes.shape({
            address: PropTypes.string,
        }),
        PropTypes.shape({
            center: PropTypes.oneOfType([
                PropTypes.arrayOf(PropTypes.number),
                PropTypes.shape({
                    lng: PropTypes.number.isRequired,
                    lat: PropTypes.number.isRequired,
                }),
            ]),
            zoom: PropTypes.number,
        }),
    ]),
    /** A geoJson description of the area that should be highlighted when there are no other markers present */
    defaultHighlight: PropTypes.array,
```

src/components/Map/Map/Map.js
```diff
@@ -70,12 +71,35 @@ export const Map = React.forwardRef((props, ref) => {
         }
     }, [defaultArea, mapRef, markers]);
 
-    useEffect(fitBounds);
+    const manageDefaultHighlight = React.useCallback(() => {
+        if (mapRef.current && mapRef.current.state.map) {
+            const { map } = mapRef.current.state;
+            if (defaultHighlight && !markers.length && !highlightFeatures) {
```
antipin, line 77: This condition looks like we only can render highlighted area if there are no markers passed.

It might be an JobFeed requirement, but then I think it should be handled on the application level.

And if it is technically possible to render both markers and areas, I don't see why we should not allow to do in on the component level.

antipin (changes requested):

src/components/Map/Map/Map.js
eszthoff: These are the requirements in the ticket. @carlobernardini ?

antipin: > These are the requirements in the ticket

Sure, and this is product level requirements: it can be implemented either on OneUI level or on the application level. My point is just to remind that AFAIR once we decided to keep OneUI components as simple as possible without "business" logic in it.

Imagine.. Years from now... There is a developer trying to use this `<Map />` component. They want to render a geojson data, they copy an example from storybook (it has few markers in place), they add `defaultHighlighting`,  check the results and see only few markers instead of highlighted area. That's confusing... :) Expected behaviour would be to see both things and instantly realise that "ah.. I also passed markers, I should remove them".

But yeah, I wasn't involved into interface design phase.. Since 2 pair of eyes checked this idea, I guess we can live with it then :)

carlobernardini: I do agree with your reasoning @antipin. But since @eszthoff already put quite some effort in this PR let's keep the API as it is. And later, when the scenario of highlighting multiple areas or mixing areas and markers presents itself, add an additional prop e.g. `polygons` and either keep `defaultHighlighting` for the current use case - or we 
deprecate it if you eventually decide to move the business logic into the application (which I'd then prefer to keep the API concise).

antipin (approved):

PR #532, opened 2021-06-02
feat(Highlighter): add highlightStyles to find method

mukiienko:

# Checklist


src/packages/Highlighter/Highlighter.ts
```diff
@@ -196,14 +196,18 @@ export class Highlighter {
...
+        highlightStyles?: CSSStyleDeclaration
+    ) {
         const matchedTerms: HighlighterResultItem[] = [];
         const notMatchedTerms: HighlighterResultItem[] = [];
         const sortedTerms = HighlighterCore.sortTerms(terms);
 
         sortedTerms.forEach((term, index) => {
             let amount = 0;
-            const styles = Highlighter.generateHighlightStyles(index);
+            const styles = highlightStyles || Highlighter.generateHighlightStyles(index);
```
agorodetskaya, line 210: what do you think about overriding styles, not replacing:
```
const styles = {
  ...Highlighter.generateHighlightStyles(index),
  ...highlightStyles,
};
```

agorodetskaya:
@mukiienko as I understand Reach needs to have single highlight color for all terms. Maybe it is better to implements such flag or something directly related to colors and do not allow to fully override styles?

src/packages/Highlighter/Highlighter.ts
mukiienko: `generateHighlightStyles` returns the bunch of styles that actually redundant for the current purpose - having the only background color. 
```
    border-radius: 1px;
    display: inline-block;
    text-indent: initial;
    background-color: rgb(214, 244, 129);
```

mukiienko:
I would like to keep the ability to control all styles. It is also consistent with `StringHightligher` that accepts any styles. So, if we want to make highlighted terms bold it will be easy to tweak on Reach side without amendment OneUI

src/packages/Highlighter/Highlighter.ts
agorodetskaya: ` border-radius: 1px;` is'n it useful one?
So if we just need to change color - we also have to send other styles to have consistent style everywhere in TK

agorodetskaya:
I just had doubts about having different highlighters across TK. For example some teams add border-radius and other not.
But oki, I don't mind to have styles here, moreover if we have it for StringHightligher.

src/packages/Highlighter/Highlighter.ts
mukiienko: The `border-radius` looks fancy for the overview profile where we have bright colors in the background and use the border for selected terms and elements should be `inline-block`. For yellow highlighting (string), it doesn't make sense.

mukiienko:
@agorodetskaya I see your point, but if we can still change the color it will be breaking consistency between teams anyway. Initially, when we started to develop OneUI library the idea was to have a public library that is not tightly coupled with our products, and other users can reuse our components.

agorodetskaya (approved):

carlobernardini (approved):

PR #589, opened 2021-08-31
feat(NumericStepper): add NumericStepper component

konstantin-lebejko:
Add new NumericStepper component
Related story: [https://jira.textkernel.nl/browse/ONEUI-258](https://jira.textkernel.nl/browse/ONEUI-258)

# Checklist

### DIFF
--- a/src/components/NumericStepper/NumericStepper.tsx
+++ b/src/components/NumericStepper/NumericStepper.tsx
@@ -4,7 +4,7 @@ import styles from './NumericStepper.scss';
 
 import { StepperButton } from '../Buttons';
 
-interface Props {
+interface Props extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
     /**
      * Callback is called when user changes component value by using built-in controls
      */
@@ -37,20 +37,23 @@ export const NumericStepper: React.FC<Props> = (props) => {
     const { onChange, step, minValue, maxValue, defaultValue, customWidth } = props;
 
     const [currentValue, setCurrentValue] = React.useState<number>(defaultValue || minValue || 0);
+    const [inputValue, setInputValue] = React.useState<string>(currentValue.toString());
 
+    // TODO: Remove this redundant "not-undefined" check one day
+    // Also without this check Jest also fails to render content sometimes
     if (
         step === undefined ||
         defaultValue === undefined ||
         minValue === undefined ||
         maxValue === undefined
     )
-        return null; // TODO: Remove this redundant "not-undefined" check
+        return null;
 
     const onValueUpdate = (value: number) => {
+        setInputValue(value.toString());
         setCurrentValue(value);
         onChange(value);
     };
-
     const handleStepUp = () => {
         const newValue = currentValue + step;
 
@@ -71,18 +74,27 @@ export const NumericStepper: React.FC<Props> = (props) => {
     };
     const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         const { value } = event.target;
-        setCurrentValue(parseInt(value, 10));
+        setInputValue(value);
     };
     const handleInputBlur = () => {
-        if (currentValue >= minValue && currentValue <= maxValue) {
-            // CASE: Entered value is in between minValue and maxValue
-            onValueUpdate(currentValue);
-        } else if (currentValue < minValue) {
-            // CASE: Entered value is less then minValue
-            onValueUpdate(minValue);
-        } else if (currentValue > maxValue) {
-            // CASE: Entered value is greater then maxValue
-            onValueUpdate(maxValue);
+        const parsedInputValue = parseInt(inputValue, 10);
+
+        if (parsedInputValue !== currentValue) {
+            if (Number.isNaN(parsedInputValue)) {
+                // CASE: Entered value cannot be converted to integer (like empty string)
+                onValueUpdate(currentValue);
+            }
+
+            if (parsedInputValue >= minValue && parsedInputValue <= maxValue) {
+                // CASE: Entered value is in between minValue and maxValue
+                onValueUpdate(parsedInputValue);
+            } else if (parsedInputValue < minValue) {
+                // CASE: Entered value is less then minValue
+                onValueUpdate(minValue);
+            } else if (parsedInputValue > maxValue) {
+                // CASE: Entered value is greater then maxValue
+                onValueUpdate(maxValue);
+            }
         }
     };
 
@@ -98,7 +110,7 @@ export const NumericStepper: React.FC<Props> = (props) => {
                 style={{ width: customWidth }}
                 onChange={handleInputChange}
                 onBlur={handleInputBlur}
-                value={currentValue.toString()}
+                value={inputValue}
                 type="number"
                 step={step}
                 min={minValue}
--- a/src/components/NumericStepper/NumericStepper.tsx
+++ b/src/components/NumericStepper/NumericStepper.tsx
@@ -34,20 +34,17 @@ interface Props extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'onCha
 const { block, elem } = bem('NumericStepper', styles);
 
 export const NumericStepper: React.FC<Props> = (props) => {
-    const { onChange, step, minValue, maxValue, defaultValue, customWidth } = props;
+    const {
+        onChange,
+        step = 1,
+        minValue = 0,
+        maxValue = Number.MAX_SAFE_INTEGER,
+        defaultValue = 0,
+        customWidth = '3ch',
+    } = props;
 
     const [currentValue, setCurrentValue] = React.useState<number>(defaultValue || minValue || 0);
     const [inputValue, setInputValue] = React.useState<string>(currentValue.toString());
-
-    // TODO: Remove this redundant "not-undefined" check one day
-    // Also without this check Jest also fails to render content sometimes
-    if (
-        step === undefined ||
-        defaultValue === undefined ||
-        minValue === undefined ||
-        maxValue === undefined
-    )
-        return null;
 
     const onValueUpdate = (value: number) => {
         setInputValue(value.toString());
--- a/src/components/NumericStepper/NumericStepper.tsx
+++ b/src/components/NumericStepper/NumericStepper.tsx
@@ -1,0 +1,124 @@
+import * as React from 'react';
+import { bem } from '../../utils';
+import styles from './NumericStepper.scss';
+
+import { StepperButton } from '../Buttons';
+
+interface Props {
+    /**
+     * Callback is called when user changes component value by using built-in controls
+     */
+    onChange: (value: number) => void;
+    /**
+     * Increase/decrease step value
+     */
+    step?: number;
+    /**
+     * Minimum possible value of NumericStepper
+     */
+    minValue?: number;
+    /**
+     * Maximum possible value of NumericStepper
+     */
+    maxValue?: number;
+    /**
+     * The value stepping starts from
+     */
+    defaultValue?: number;
+    /**
+     * De-facto CSS property that sets custom width for input
+     */
+    customWidth?: string;
+}
+
+const { block, elem } = bem('NumericStepper', styles);
+
+export const NumericStepper: React.FC<Props> = (props) => {
+    const { onChange, step, minValue, maxValue, defaultValue, customWidth } = props;
+
+    const [currentValue, setCurrentValue] = React.useState<number>(defaultValue || minValue || 0);
+
+    if (
+        step === undefined ||
+        defaultValue === undefined ||
+        minValue === undefined ||
+        maxValue === undefined
+    )
+        return null; // TODO: Remove this redundant "not-undefined" check
+
+    const onValueUpdate = (value: number) => {
+        setCurrentValue(value);
+        onChange(value);
+    };
+
+    const handleStepUp = () => {
+        const newValue = currentValue + step;
+
+        if (newValue > maxValue) {
+            onValueUpdate(maxValue);
+        } else {
+            onValueUpdate(newValue);
+        }
+    };
+    const handleStepDown = () => {
+        const newValue = currentValue - step;
+
+        if (newValue < minValue) {
+            onValueUpdate(minValue);
+        } else {
+            onValueUpdate(newValue);
+        }
+    };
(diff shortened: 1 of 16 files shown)


src/components/NumericStepper/NumericStepper.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const [currentValue, setCurrentValue] = React.useState<number>(defaultValue || minValue || 0);
+
+    if (
+        step === undefined ||
+        defaultValue === undefined ||
+        minValue === undefined ||
+        maxValue === undefined
+    )
+        return null; // TODO: Remove this redundant "not-undefined" check
```
konstantin-lebejko, line 47: Guys could you please assist me in this case?
I'm not that familiar with TS unit testing and don't get why the linter wants me to do this kind of check when I already have defaultProps.

src/components/NumericStepper/__tests__/NumericStepper.spec.js
```diff
@@ -0,0 +1,153 @@
...
+    it('should set top edge value if user enters value above allowed limit', () => {
+        wrapper = mount(<NumericStepper onChange={onChangeMock} maxValue={10} />);
+        const data = { target: { value: '11' } };
+
+        wrapper.find('input').simulate('change', data);
+        wrapper.find('input').simulate('blur');
+
+        expect(wrapper.find('input').prop('value')).toEqual('10');
+    });
+    it('should react on keyup/keydown press', () => {
```
konstantin-lebejko, line 146: I tried many ways to do this, but does anyone have the idea that might work?

konstantin-lebejko:
Alright, there are several moments I want to discuss. Please read attentively. Sorry for the long post by the way. 
While designing this component I kept in mind simplicity and scalability, but when I faced the implementation, I noticed that some design choices still should be discussed. 

**1. Storing a component value as a single variable**
I currently use only one variable to store data both for filling the input and calling onChange function. As I validate data when user blurs the input it cases some difficulties. 

I noticed that if a user clears the input (leaves it blank), the blur function will set the local state, but won't invoke onChange because the empty string is not a valid value. In my opinion, the best way to solve this issue is just to reset the value to the previous one. But we don't already remember it because we already overwrote it in handleInputChange.

**This issue could be solved in several ways**:
1. Adding second state variable that would contain the input value (as string)
2. Manipulating input directly thru refs (aka vanilla js)
3. Reset input value not to the previous one, but to default value

Also, current implementation slightly complicates unit testing because I could write extra stability checks. 

**2. Moved NumericStepper buttons to a separate component**
In my opinion, components called SearchButton and StepperButton are quite similar, and we could merge these buttons into a one component called - RoundButton. They have minor styling differences, and these differences could be easily managed by props. 

**3. Component dimensions**
Jira task contains the list of styles and some comments, but I was not able to find the actual dimensions of component. Do we have something like Figma mockup or any other source file? Also, tried to match all other requirements.

konstantin-lebejko:
Alright, there are several moments I want to discuss. Please read attentively. Sorry for the long post by the way. 
While designing this component I kept in mind simplicity and scalability, but when I faced the implementation, I noticed that some design choices still should be discussed. 

**1. Storing a component value as a single variable**
I currently use only one variable to store data both for filling the input and calling onChange function. As I validate data when user blurs the input it cases some difficulties. 

I noticed that if a user clears the input (leaves it blank), the blur function will set the local state, but won't invoke onChange because the empty string is not a valid value. In my opinion, the best way to solve this issue is just to reset the value to the previous one. But we don't already remember it because we already overwrote it in handleInputChange.

**This issue could be solved in several ways**:
1. Adding second state variable that would contain the input value
2. Manipulating input directly thru refs (aka vanilla js)
3. Reset input value not to the previous one, but to default one

Also, current implementation slightly complicates unit testing because I could write extra additional stability checks. 

**2. Moved NumericStepper buttons to a separate component**
In my opinion, button components called SearchButton and StepperButton are quite similar, and we could merge these buttons into a one component called - RoundButton. They have minor styling differences, and these differences could be easily managed by props. 

**3. Component dimensions**
Jira task contains the list of styles and some comments, but I was not able to find the actual dimensions of component. Do we have something like Figma mockup or any other source file? Also, tried to match all other requirements.

src/components/NumericStepper/NumericStepper.tsx
mukiienko: yep, we need the additional checks if even we have `defaultProps`. TypeScript can't infer the type from `defaultProps`.
https://github.com/textkernel/oneui/blob/master/CONTRIBUTING.md#type-definitions-for-props-and-states

src/components/NumericStepper/__tests__/NumericStepper.spec.js
mukiienko: I also tried several approaches, but it doesn't realy help. I think we need to create a follow-up story for investigating the issue.

mukiienko:
> Alright, there are several moments I want to discuss. Please read attentively. Sorry for the long post.
> While designing this component I kept in mind simplicity and scalability, but when I faced the implementation, I noticed that some design choices still should be discussed.
> 
> **1. Storing a component value as a single variable**
> I currently use only one variable to store data both for filling the input and calling onChange function. As I validate data when user blurs the input it cases some difficulties.
> 
> I noticed that if a user clears the input (leaves it blank), the handleInputChange function will set the local state, but onBlur won't invoke onChange because the empty string is not a valid value. In my opinion, the best way to solve this issue is just to reset the value to the previous one. But we don't store the previous value somewhere.
> 
> **Here is demo:**
> https://numeric-stepper-demo-hltbnnz64-liabeika-textkerneln.vercel.app/?path=/story/molecules-numericstepper--numericstepper
> 
> **This issue could be solved in several ways**:
> 
> 1. Adding second state variable that would contain the input value (as string)
> 2. Manipulating input directly thru refs (aka vanilla js)
> 3. Reset input value not to the previous one, but to default value
> 
> Also, current implementation slightly complicates unit testing because I could write extra stability checks. Also, having two values will allow is to optimize onChange calls.
> 
> **2. Moved NumericStepper buttons to a separate component**
> In my opinion, components called SearchButton and StepperButton are quite similar, and we could merge these buttons into a one component called - RoundButton. They have minor styling differences, and these differences could be easily managed by props. What do you think about it?
> 
> **3. Component dimensions**
> Jira task contains the list of styles and some comments, but I was not able to find the actual dimensions of component. Do we have something like Figma mockup or any other source file? Also, tried to match all other requirements.

Very nice explanation. Thanks!
1) I think we need to allow users to reset input if `defaultValue` is not set. It is not clear why we can't run `onChange` event for the empty string. In case when we have `defaultValue` we need probably to set the value to default @carlobernardini.
2) I like this idea. We can create `RoundButton` with a possibility to pass any icon instead of props.
3) I'd say we don't have strict rules for the dimensions of components. It would be nice to have something @carlobernardini

src/components/NumericStepper/__tests__/NumericStepper.spec.js
```diff
@@ -0,0 +1,152 @@
...
+        /**
+         * TODO: Test with E2E tests.
+         * As we actually validate data on blur, its quite challenging to chain together mouseEnter, keyDown and blur.
+         * I've tried multiple ways to do it, and don't want to spend way more time on that.
+         */
```
mukiienko, line 150: I think it is important to test the result of the callback  - `onChangeMock.toHaveBeenCalledWith`

src/components/NumericStepper/NumericStepper.tsx
```diff
@@ -0,0 +1,124 @@
...
+     */
+    maxValue?: number;
+    /**
+     * The value stepping starts from
+     */
+    defaultValue?: number;
+    /**
+     * De-facto CSS property that sets custom width for input
+     */
+    customWidth?: string;
```
mukiienko, line 31: In string value, we can pass whatever we want and format `3ch` looks complicated. Let's use a number instead of a string.

mukiienko (commented):
Screenshot 2021-09-03 at 16 01 30

I noticed some warnings related to NumericStepper while test running.

carlobernardini:
>This issue could be solved in several ways:
>1. Adding second state variable that would contain the input value (as string)
>2. Manipulating input directly thru refs (aka vanilla js)
>3. Reset input value not to the previous one, but to default value
...
> I think we need to allow users to reset input if defaultValue is not set. It is not clear why we can't run onChange event for the empty string. In case when we have defaultValue we need probably to set the value to default @carlobernardini.

To me both 1 and 3 sound fine, as long as you can't have an empty input nor an invalid value. But `defaultValue` doesn't seem to be required, so what do you do if it's undefined? Maybe option 1 is the most secure.

>3. Component dimensions
> Jira task contains the list of styles and some comments, but I was not able to find the actual dimensions of component. Do we have something like Figma mockup or any other source file? Also, tried to match all other requirements.
It's like @mukiienko says - for now there are no strict specs other than what I provide in the ticket. If something needs adjustments I'll tell you.

src/components/NumericStepper/NumericStepper.tsx
konstantin-lebejko: Not really, it's done intentionally, because this value sets input width directly in style attribute. 

It means that instead of 3ch we can use:
1. 100px -> width: 100px
2. 100% -> width: 100%
3.  10vw -> width: 10vw

That's why it still needs to remain a string.

src/components/NumericStepper/NumericStepper.tsx
eszthoff: you can also add the default values when destructure props in the beginning. It helps with the TS check, but has a potential pitfall of later updating the default in one place, but not the other, and so creating some bugs. Not the most convenient...

konstantin-lebejko: Got it.

src/components/NumericStepper/__tests__/NumericStepper.spec.js
konstantin-lebejko: Created ticket for that: https://jira.textkernel.nl/browse/ONEUI-280

konstantin-lebejko:
Implemented one of the ways we discussed - the one where we create second state varable.

src/components/NumericStepper/NumericStepper.tsx
konstantin-lebejko: Updated the code to eliminate the error message

src/components/NumericStepper/__tests__/NumericStepper.spec.js
konstantin-lebejko: Added this check where we test callback behavior.

konstantin-lebejko:
Updated tests, and added check for invalid value input.

mukiienko (approved):

eszthoff:
Before merging this PR, please make sure to name it as per convention. See CONTRIBUTE.md. We need that for the automatic change-log to work correctly.

eszthoff (changes requested):

eszthoff:
I'm not sure what the requirements from keyboard navigation are. The input doesn't get validated when clicking ENTER or ESC, but maybe that's fine. The rest LGTM

eszthoff:
> Screenshot 2021-09-03 at 16 01 30
> 
> I noticed some warnings related to NumericStepper while test running.

this seems not to have been solved yet.

src/components/NumericStepper/NumericStepper.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const {
+        onChange,
+        step = 1,
+        minValue = 0,
+        maxValue = Number.MAX_SAFE_INTEGER,
+        defaultValue = 0,
+        customWidth = '3ch',
+    } = props;
+
+    const [currentValue, setCurrentValue] = React.useState<number>(defaultValue || minValue || 0);
```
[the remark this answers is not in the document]
konstantin-lebejko: We dont.

src/components/NumericStepper/NumericStepper.tsx
```diff
@@ -0,0 +1,133 @@
...
+        onChange(value);
+    };
+    const handleStepUp = () => {
+        const newValue = currentValue + step;
+
+        if (newValue > maxValue) {
+            onValueUpdate(maxValue);
+        } else {
+            onValueUpdate(newValue);
+        }
```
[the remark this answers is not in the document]
konstantin-lebejko: Didn't know that it is even possible. Implemented.

src/components/NumericStepper/__tests__/NumericStepper.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    it('should set top edge value if user enters value above allowed limit', () => {
+        wrapper = mount(<NumericStepper onChange={onChangeMock} maxValue={10} />);
+        const data = { target: { value: '11' } };
+
+        wrapper.find('input').simulate('change', data);
+        wrapper.find('input').simulate('blur');
+
+        expect(wrapper.find('input').prop('value')).toEqual('10');
+    });
+    it('should set top edge value if user enters value above allowed limit', () => {
```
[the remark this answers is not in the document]
konstantin-lebejko: Yeah, its duplicate. Removed the duplicate. Also,  added two additional tests (1 for increase, 1 for decrease) if I understood you correctly.

konstantin-lebejko:
@eszthoff Updated the PR according to your comments. Please check if everything is fine and if there are any errors (red lines) in console.

src/components/NumericStepper/__tests__/NumericStepper.spec.js
```diff
@@ -11,6 +11,14 @@ describe('<NumericStepper> component', () => {
         jest.resetAllMocks();
     });
 
+    const decreaseClick = () => {
+        wrapper.find('StepperButton').at(0).prop('onClick')();
+    };
+
+    const increaseClick = () => {
+        wrapper.find('StepperButton').at(1).prop('onClick')();
+    };
```
eszthoff, line 20: nice! It is easier to read the tests when they call functions like this with meaningful name. Nice improvement.

I also didn't know you can call onClick like this. We normally use `simulate('click')`. Good to learn something new each time :-)

eszthoff (approved):

src/components/NumericStepper/__tests__/NumericStepper.spec.js
konstantin-lebejko: Yeah, I've read somewhere that `simulate` might be deprecated soon.

PR #595, opened 2021-09-13
feat(Alert): reimplement Alert and add Confirm dialog

eszthoff:
Story: [ONEUI-273](https://jira.textkernel.nl/browse/ONEUI-273)

BREAKING CHANGE: Alert dialog API has been completely changed.

# Checklist


src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
...
+        ...rest
+    } = props;
+
+    if ((!cancelLabel && !!onCancel) || (!!cancelLabel && !onCancel)) {
+        console.warn(
+            'It seems you are trying to render a confirm dialog. In that case make sure to pass both "onCancel" and "cancelLabel" props. Otherwise please remove both'
+        );
+    }
+
+    const isConfirm = !!cancelLabel;
```
carlobernardini, line 44: I'm a bit confused with this design. Here it seems like you infer from the presence (or not) of a cancel label / button, whether it's supposed to be an alert or confirm dialog. However, in the Storybook story I also see `type="[confirm |alert]"`. Which do you intend to use to make this distinction?
I think having a distinct `type` prop - even though more explicit / descriptive - may be superfluous because when `onCancel` is defined, it implicitly _has_ to be a confirm dialog.
Alternatively, if you want it to be more explicit, you could think of a separate Confirm component (which accepts onCancel / cancelLabel) that extends the Alert component (does not accept onCancel).

eszthoff: I removed the example component from the story. Hopefully it is clearer now. (`type` was a prop on the example only, which implements the state for the alert to be shown or not)

mukiienko: I like the idea of separating the Alert and Confirm dialog when Confirm extends the Alert component.

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
...
+    /** Dialog title */
+    title?: string;
+    /** Callback to be called on Accept button click (shown in both alert and confirm) */
+    onAccept: () => void;
+    /** Callback to be called on Cancel button click (only relevant for confirm dialog) */
+    onCancel: () => void;
+    /** Accept button's text */
+    acceptLabel: string;
+    /** Cancel button's text */
+    cancelLabel: string;
```
mukiienko, line 21: I think it makes sense to group button's props in one `Button` type. In this case, we always have a label and callback for the button. We can make  `cancelButton` optional.
```
type Button = {
    label: string;
    onClick: () => void;
}
.....
acceptButton: Button;
cancelButton?: Button;
```

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { bem } from '../../utils';
+import { Button } from '../Buttons';
+import { Modal, ModalProps } from '../Modal';
+import styles from './Alert.scss';
+
+interface Props extends ModalProps {
+    /** Should the alert appear */
+    showAlert?: boolean;
```
mukiienko, line 9: What do you think about renaming prop to `isOpen`?

mukiienko (dismissed):

konstantin-lebejko:
Looks good for me after all these minor changes.

src/components/AlertConfirm/AlertConfirm.tsx
```diff
@@ -0,0 +1,75 @@
...
+                {!!title && (
+                    <Heading level="h2" {...elem('title', props)}>
+                        {title}
+                    </Heading>
+                )}
+                {children}
+            </div>
+            <div {...elem('actions', props)}>
+                {isConfirm && (
+                    <Button onClick={cancelButton!.onClick} context="link">
```
antipin, line 55: Time to upgrade typescript (not as part of this PR of course) :) 
TS 4.4 is smart enough to understand that `cancelButton` here is of `ConfirmProps` type, so no need for `!`.

But having TS 3.8 I'd rather use `cancelButton` instead of `isConfirm` here:
```
{cancelButton && (
                    <Button onClick={cancelButton.onClick} context="link">
```

src/components/AlertConfirm/AlertConfirm.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+Dialog.displayName = 'Dialog';
+
+export const Alert: React.FC<AlertProps> = (props) => <Dialog {...props} />;
+
+Alert.displayName = 'Alert';
+
+export const Confirm: React.FC<ConfirmProps> = (props) => <Dialog {...props} />;
+
+Confirm.displayName = 'Confirm';
```
antipin, line 75: Hmm.. It's an elegant solution to create 2 components from one.

But it's not a good idea to do mix it up in the same file.

Currently, we always have a correspondence between component folder/file name and component name itself.

How about we stick to this rule and use the following file structure:
```
├── Dialogs
│   ├── Dialog
│   │   ├── Dialog.scss
│   │   ├── Dialog.tsx
│   │   ├── __tests__
│   │   └── index.ts
│   ├── Alert
│   │   ├── Alert.jsx
│   │   ├── __tests__
│   │   └── index.js
│   ├── Confirm
│   │   ├── Confirm.jsx
│   │   ├── __tests__
│   │   └── index.js
│   └── index.ts
```

src/components/AlertConfirm/AlertConfirm.tsx
```diff
@@ -0,0 +1,75 @@
...
+    /** Alerts do not have cancel button */
+    cancelButton?: never;
+}
+
+export interface ConfirmProps extends DialogProps {
+    /** Properties of the cancel button */
+    cancelButton: ButtonProps;
+}
+
+const { block, elem } = bem('Alert', styles);
```
antipin, line 36: Should be: `bem('Alert', styles);` -> bem('Dialog', styles);

antipin (changes requested):

src/components/AlertConfirm/AlertConfirm.tsx
eszthoff: I was thinking to do that too, but I thought it is more readable this way. Even if the TS isn't that nice, the dev intention is easier to understand.

eszthoff: I can add a TODO point to remove `!` when upgrading TS

antipin (approved):
💯

carlobernardini (approved):

PR #638, opened 2021-11-09
fix(Gauge): Safari bug where gradient doesn't render

eszthoff:
Story: [JF-2780](https://jira.textkernel.nl/browse/JF-2780)

See details in comments there

### DIFF
--- a/src/components/Gauge/Gauge.tsx
+++ b/src/components/Gauge/Gauge.tsx
@@ -59,12 +59,6 @@ export const Gauge: React.FC<Props> = (props) => {
     return (
         <div {...rest} {...block(props)}>
             <svg {...elem('svg', props)} width="100%">
-                <defs>
-                    <linearGradient id={`Gauge__gradient--${context}`}>
-                        <stop {...elem('gradientStart', props)} offset="50%" />
-                        <stop {...elem('gradientEnd', props)} offset="100%" />
-                    </linearGradient>
-                </defs>
                 <circle
                     {...elem('circleBackground', props)}
                     r={GAUGE_RADIUS}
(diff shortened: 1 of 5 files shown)


carlobernardini:
@eszthoff I might have figured out what the issue is - please see my last comment on the story.

eszthoff:
> @eszthoff I might have figured out what the issue is - please see my last comment on the story.

Great. When I tried for some reason it didn't work. I guess I did something wrong. It does fix it now. Updated the PR

src/components/Gauge/Gauge.tsx
```diff
@@ -60,7 +63,7 @@ export const Gauge: React.FC<Props> = (props) => {
         <div {...rest} {...block(props)}>
             <svg {...elem('svg', props)} width="100%">
                 <defs>
-                    <linearGradient id={`Gauge__gradient--${context}`}>
+                    <linearGradient id={`Gauge__gradient--${context}-${id.current}`}>
```
carlobernardini, line 66: I don't think `${context}` is relevant any longer since every definition is now unique

carlobernardini: ...apart from that, LGTM.

eszthoff: it doesn't harm

carlobernardini: I don't understand. Why keep code that doesn't serve a purpose?
Besides, it _does_ "harm": it's misleading.

eszthoff: I'm not following what do you mean. It is still the colors of that context, so the name is accurate. It is only the name, where the ID is added, is unique. 
To say the same thing in an other way, `${context}` is as redundant as `Gauge` and `gradient`.

carlobernardini: There is only a single gradient definition per instance of `Gauge`. This could make it seem as if there would be one per context. It serves no purpose. "Gauge" and "gradient" are just names - not variables.
It's not really productive to continue this debate at length so if you can't follow then go ahead.

PR #661, opened 2021-12-17
refactor(location): type location components; expose additional location utilities

eszthoff:
Story: [ONEUI-162](https://jira.textkernel.nl/browse/ONEUI-162); [JF-3000](https://jira.textkernel.nl/browse/JF-3000)

BREAKING CHANGE: LocationSelector's hasRadius props now defaults to false (was true before)


src/components/LocationAutocomplete/LocationAutocomplete.tsx
```diff
@@ -11,12 +10,47 @@ import POWERED_BY_GOOGLE_ON_WHITE_2X from './images/powered_by_google_on_white@2
...
+    /** to be shown when no suggestions are available */
+    noSuggestionsPlaceholder: string;
+    /** trigger of the initial focus of the input field */
+    isFocused?: boolean;
+    /** label for the Clear button */
+    clearLabel?: string;
+    /** defines if there's a single location to select in component */
+    singleLocation?: boolean;
+    /** callback to be called with selected value. */
+    onSelectionChange: (value: google.maps.places.AutocompletePrediction) => void;
```
mukiienko, line 29: It would be nice to have a rule of thumb to expose Types of arguments of the callbacks. For instance, if I want to define `onSelectionChange` handler outside of the render method:
```
const handlerSelectionChange = (args: onSelectionChangeLocationAutocomplete) => {
....
}
render (
  <LocationAutocomplete onSelectionChange={handlerSelectionChange} ...>
);
```
What do you think?

eszthoff: yes, I agree! Only, I'm not sure we ever set it up properly (exposing TS types - https://jira.textkernel.nl/browse/ONEUI-213). If there is a pattern I should apply to make types exposed, that is already working, please help me out with it.

mukiienko: For `LocationSelector` we already export component Props Types and helper utils Types, but probably it isn't a good example. We can wait for ONEUI-213 and then discuss it in detail about expose Types rules.

mukiienko (approved):

src/components/LocationAutocomplete/LocationAutocomplete.tsx
```diff
@@ -173,45 +209,8 @@ export const LocationAutocomplete = (props) => {
...
-    showCountryInSuggestions: PropTypes.bool,
-    /** function to remove all locations */
-    onRemoveAllLocations: PropTypes.func,
-    /** function to be executed if error occurs while fetching suggestions */
-    onError: PropTypes.func,
-    /** To hide powered by google logo. For legal reasons only set it to true if Google map is displayed on the same screen as this component! */
-    hidePoweredByGoogleLogo: PropTypes.bool,
-};
-
 LocationAutocomplete.defaultProps = {
```
carlobernardini, line 212: Should we not drop defaultProps in favour of object defaults? I noticed we already did so with some other components, wondering if this should then be the norm.

src/components/LocationAutocomplete/LocationAutocomplete.tsx
```diff
@@ -11,12 +10,47 @@ import POWERED_BY_GOOGLE_ON_WHITE_2X from './images/powered_by_google_on_white@2
...
+    hidePoweredByGoogleLogo?: boolean;
+}
+
 const { elem } = bem('LocationAutocomplete', styles);
 
 const DEBOUNCE_DELAY = 350;
 const ACCEPTABLE_API_STATUSES = ['OK', 'NOT_FOUND', 'ZERO_RESULTS'];
 
-export const LocationAutocomplete = (props) => {
+const LocationAutocomplete = (props) => {
```
carlobernardini, line 53: This does not seem to use the previously declared interface?

src/components/LocationCard/LocationCard.tsx
```diff
@@ -58,42 +85,15 @@ export const LocationCard = (props) => {
...
-    distanceRadius: PropTypes.number.isRequired,
-    /** Slider value change callback */
-    onRadiusChange: PropTypes.func,
-    /** Location card delete callback */
-    onDelete: PropTypes.func,
-    /** HTML tag to be used as a wrapping component */
-    As: PropTypes.string,
-};
-
 LocationCard.defaultProps = {
```
carlobernardini, line 88: It seems this component is using both object default values as well as defaultProps and some props (like `As` and `onRadiusChange`) are specified redundantly.

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.tsx
```diff
@@ -1,19 +1,62 @@
...
-import { LocationAutocomplete } from '../../LocationAutocomplete';
-import { Map } from '../../Map';
+import { LocationAutocomplete, LocationAutocompleteProps } from '../../LocationAutocomplete';
+import { Map, CircularMarker } from '../../Map';
 import { SIZES, ENTER_KEY } from '../../../constants';
 import styles from './LocationSelectorDialog.scss';
+import { LocationSelectorLocation } from '../utils';
+
+// Props for child components that we just push through with prop-drilling
+type CardPropsForSelectorDialog = Omit<
```
carlobernardini, line 14: Should these perhaps better be moved to `src/types/`?

src/components/Map/Map/Map.tsx
```diff
@@ -12,17 +11,57 @@ const circleOptions = (radius) => ({
...
     visible: true,
-    radius,
     zIndex: 1,
 });
 
-export const Map = React.forwardRef((props, ref) => {
-    const { defaultArea, markers, mapContainerStyle, defaultHighlight, ...rest } = props;
-    const mapRef = ref || React.createRef();
-    const [highlightFeatures, setHighlightFeatures] = React.useState(null);
+type CircularMarker = {
```
carlobernardini, line 17: Should this rather go inside `src/types/`?

src/components/LocationAutocomplete/LocationAutocomplete.tsx
```diff
@@ -11,12 +10,47 @@ import POWERED_BY_GOOGLE_ON_WHITE_2X from './images/powered_by_google_on_white@2
...
+    /** function to be executed if error occurs while fetching suggestions */
+    onError?: (error: google.maps.places.PlacesServiceStatus) => void;
+    /** To hide powered by google logo. For legal reasons only set it to true if Google map is displayed on the same screen as this component! */
+    hidePoweredByGoogleLogo?: boolean;
+}
+
 const { elem } = bem('LocationAutocomplete', styles);
 
 const DEBOUNCE_DELAY = 350;
 const ACCEPTABLE_API_STATUSES = ['OK', 'NOT_FOUND', 'ZERO_RESULTS'];
```
carlobernardini, line 51: Are these candidates for `src/constants/`?

carlobernardini:
When checking out the LocationSelector component in Storybook, it seems that the radius slider for the selected locations is gone although it's not entirely clear to me why (how is it determined whether or not radius can be set in the UI)? Can you comment on that in the PR description (so it ends up in the changelog) and is that perhaps a breaking change?

src/components/LocationAutocomplete/LocationAutocomplete.tsx
eszthoff: I think the norm is that we use default props, but that is not picked up by TS properly. So we also add it as default at object destructuring sometimes, just for that reason. We could look at the overall standards, but I would not do it in this PR. Here I tried to keep the changes to minimum, so that it is still reviewable.

src/components/LocationCard/LocationCard.tsx
eszthoff: yes, to help out TS. There is a section about it in CONTRIBUTE.md#L94

src/components/LocationAutocomplete/LocationAutocomplete.tsx
eszthoff: hhmmm. Good catch. I wonder why the static checks didn't complain.

src/components/LocationAutocomplete/LocationAutocomplete.tsx
eszthoff: perhaps. but I didn't want to make any other changes then what is needed for TS migration. There are a few things "funny" with these components, like it uses `function` definitions instead of `const`. But If I change all that, the PR will be very difficult to review, and more changes for bugs.

src/components/LocationSelector/LocationSelectorDialog/LocationSelectorDialog.tsx
eszthoff: I think it is pretty specific here and won't make sense in the global space.

src/components/Map/Map/Map.tsx
eszthoff: ok, I can move it.

src/components/LocationAutocomplete/LocationAutocomplete.tsx
carlobernardini: We should then perhaps consider making object default values the norm (to reduce redundancy and promote consistency) but indeed, if that's not yet the case lets then deal with that separately.

src/components/Map/Map/Map.tsx
eszthoff: looking at it closer... we don't have many global type declarations in OneUI. They more local and exported only when needed. I would keep it as it is for now.
If we move it to a separate file, I would argue to move all types from LocationSelector/utils.ts as well. And then look around if there are some others. Then the question is how do we expose these outside of OneUI, which we already have a ticket for. So I would wait for that, and do it all at once.

carlobernardini (approved):

PR #664, opened 2021-12-21
fix(Location): extend prop typing with HTML element props

eszthoff:


stories/LocationSelector.tsx
```diff
@@ -41,6 +41,7 @@ storiesOf('Organisms|LocationSelector', module)
 
         return (
             <LocationSelector
+                className="test-class"
```
carlobernardini, line 44: Maybe add a test?

eszthoff: what sort of test do you have in mind?

src/components/LocationSelector/LocationSelector.tsx
```diff
@@ -19,7 +19,9 @@ import styles from './LocationSelector.scss';
 
 const { block, elem } = bem('LocationSelector', styles);
 
-interface Props extends LocationSelectorDialogWithGoogleLoaderProps {
+interface Props
+    extends LocationSelectorDialogWithGoogleLoaderProps,
+        Omit<React.HTMLAttributes<HTMLDivElement>, 'id'> {
```
suleymangezsat, line 24: Do we need to omit `id` here ? It's already defined as optional string in `HTMLAttributes`

stories/LocationSelector.tsx
suleymangezsat, line 44: I guess it's forgotten to delete

src/components/LocationSelector/LocationSelector.tsx
eszthoff: yes, and the definitions clash, so had to Omit it.

stories/LocationSelector.tsx
eszthoff: no, it is intentional. This is the error we had that inspired this PR. TS would throw when adding a className. So as a "test" I added it to storybook. I wonder if we shouldn't do it in every story, as this is not the first time we forgot to handle it. But that is for a separate discussion I think.

src/components/LocationSelector/LocationSelector.tsx
suleymangezsat: Interesting 🤔 I've tried to remove omit on my local and TS didn't give any error or warning

eszthoff: 🤔  I tried it again, and now TS is happier. Good catch. Thanks. I will remove it

stories/LocationSelector.tsx
carlobernardini: A test that ensures custom html attributes can be applied (without causing errors, but I guess that's already covered with linting) like you seem to have added this line for

PR #702, opened 2022-02-02
chore: update package with node/npm version

eszthoff:
Upgrade required Node version to >=15; npm version >=7. Enforce package-lock file version 2


carlobernardini:
Without defining `engine-strict=true` in [.npmrc](.npmrc), `npm install` will only issue a warning while still overwriting package-lock.json, instead of erroring. I think it's best if we strictly enforce the use of a supported node version.

https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict

package.json
```diff
@@ -24,6 +24,10 @@
         "coveralls": "cat ./coverage/lcov.info | node node_modules/.bin/coveralls",
         "release": "standard-version -t ''"
     },
+    "engines": {
+        "node": ">=12",
+        "npm": ">=7.1.0"
```
carlobernardini, line 29: Node 7 is required to generate a v2 lockfile. Node 15 is the first version that ships with npm 7. So shouldn't we require node >= 15 rather than 12?

carlobernardini:
Our Github Action (CI/CD) configured in [.github/actions/npm-cache/action.yml](https://github.com/textkernel/oneui/tree/master/.github/actions/npm-cache/action.yml) is currently using node 14. Please check that this won't create conflicts or if it should rather use 16 too.

carlobernardini:
Please include a brief description :)

eszthoff:
>

eszthoff:
> Our Github Action (CI/CD) configured in [.github/actions/npm-cache/action.yml](https://github.com/textkernel/oneui/tree/master/.github/actions/npm-cache/action.yml) is currently using node 14. Please check that this won't create conflicts or if it should rather use 16 too.

@antipin can you advise? can we change the version there to 16?

carlobernardini (approved):

PR #703, opened 2022-02-04
feat(Map): Extend Map component to support geoJson files as markers

konstantin-lebejko:
Story: [https://jira.textkernel.nl/browse/ONEUI-291](https://jira.textkernel.nl/browse/ONEUI-291)

# Checklist


src/components/Map/Map/Map.tsx
```diff
@@ -56,10 +61,13 @@ interface Props extends Omit<GoogleMap, 'onLoad' | 'mapContainerStyle' | 'option
 const Map = React.forwardRef<GoogleMap, Props>((props, ref) => {
     const { defaultArea, markers = [], mapContainerStyle, defaultHighlight, ...rest } = props;
     const mapRef = ref || React.createRef<GoogleMap>();
-    const [highlightFeatures, setHighlightFeatures] = React.useState<
+    const [defaultHighligh, setHighlightFeatures] = React.useState<
```
eszthoff, line 64: maybe can change the name of the set function as well to match

src/components/Map/Map/Map.tsx
```diff
@@ -77,64 +85,121 @@ const Map = React.forwardRef<GoogleMap, Props>((props, ref) => {
...
+        const cMarkers: CircularMarker[] = [];
+        const rAreas: RegionArea[] = [];
+
+        markers.forEach((item) => {
+            if ('center' in item) cMarkers.push(item);
+            if ('type' in item && 'features' in item) rAreas.push(item);
+        });
+
+        if (JSON.stringify(circularMarkers) !== JSON.stringify(cMarkers))
+            setCircularMarkers(cMarkers);
```
eszthoff, line 165: Let's be consistent with the {}. If it is not set in the linter, can you add https://eslint.org/docs/rules/curly with option 'all'?

src/components/Map/__tests__/Map.spec.js
```diff
@@ -82,4 +86,35 @@ describe('<Map/> that renders a Map with markers', () => {
...
+        const wrapper = mount(
+            <Map markers={[DR_DRENTHE, FL_FLEVOLAND, FR_FRIESLAND, GE_GELDERLAND]} />
+        );
+        expect(setZoomMock).toHaveBeenCalledTimes(1);
+        expect(setCenterMock).toHaveBeenCalledTimes(1);
+        wrapper.setProps({ markers: [pointMarker, regionMarker] });
+        expect(setZoomMock).toHaveBeenCalledTimes(2);
+        expect(setCenterMock).toHaveBeenCalledTimes(2);
+    });
+    it('should center map when geoJson objects are used as markers', () => {
```
eszthoff, line 105: this is (almost) same as test in line 89. This one can be removed

stories/Map.tsx
```diff
@@ -26,11 +30,27 @@ storiesOf('Atoms|Map', module)
...
+                    lat: 52.5112671,
+                    lng: 7.2535521,
+                },
+            };
             const markers = {
                 marker1: [defaultMarker],
                 marker2: [addedMarker],
                 multiple: [defaultMarker, addedMarker],
                 none: [],
+                'pointer marker': [pointMarker],
```
eszthoff, line 44: zoom for point-marker doesn't work well. But anyway, it is not our usecase. Maybe better remove it from the examples.

stories/Map.tsx
```diff
@@ -26,11 +30,27 @@ storiesOf('Atoms|Map', module)
...
+                areas: [DR_DRENTHE, FL_FLEVOLAND, FR_FRIESLAND, GE_GELDERLAND],
```
eszthoff, line 45: indeed this takes visibly long time. Can we just list here, and the next example, just 1 area. I think that demonstrates the component well enough. It is not OneUI's responsibility to optimize the performance caused by large data passed as prop. So it is ok, if we simplify it here

eszthoff (changes requested):

src/components/Map/Map/Map.tsx
konstantin-lebejko: Fixed. Also renamed variable to defaultAreaHighlight to avoid confusion with defaultHighlight prop.

src/components/Map/Map/Map.tsx
konstantin-lebejko: Added this rule. Also updated some other components.

src/components/Map/__tests__/Map.spec.js
konstantin-lebejko: Removed.

stories/Map.tsx
konstantin-lebejko: Actually, I added it for purpose. After my updates a couple of tests were failing and I was wondered why it happens when we add marker without radius. So, even considering that the zoom does not work properly I would like to keep it, because it might be helpful for debug and demonstration purposes.

stories/Map.tsx
konstantin-lebejko: Left only one geoJson object.

eszthoff (approved):

carlobernardini (approved):

PR #742, opened 2022-04-13
feat(LabelPicker): new component

eszthoff:
Story: [JF-3191](https://jira.textkernel.nl/browse/JF-3191) | [ONEUI-300](https://jira.textkernel.nl/browse/ONEUI-300)

BREAKING CHANGE: Type checking for Checkbox moved to TS and might break certain implementations that were not as intended by the original API

# Checklist
- [x] The component comes with a detailed [types](CONTRIBUTING.md#component-props) (and defaultProps) definition.


src/components/LabelPicker/LabelPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+    /** a button like element that supports onClick handler to be used as the trigger */
+    children: React.ReactElement<React.ButtonHTMLAttributes<B>>;
+    /** callback to be called when the state of a checkbox changes */
+    onChange: (Label) => void;
+    /** callback to add new label */
+    onAdd: (name: string) => void;
+    /** callback when clicking the Done button */
+    onDone?: () => void;
+    /** callback on outer click */
+    onCancel?: () => void;
```
mukiienko, line 24: What do you think about renaming prop `onCancel` -> `onClose`?

src/components/LabelPicker/LabelPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+    /** text to be shown as input placeholder for adding labels */
+    inputPlaceholder: string;
+    /** label for the done button */
+    doneLabel: string;
+    /** Ref to access the input field */
+    ref?: React.ForwardedRef<HTMLButtonElement>;
+}
```
mukiienko, line 31: I'd also add some validation message prop, e.g `validationMessage`. For instance, if a typed value already exists in labels or a value are too short. In this scenario, some warning message will be shown, and add button will be disabled. What do you think?

src/components/LabelPicker/LabelPicker.tsx
eszthoff: Not sure. clicking on `Done` should close the dialog too. We can call it `onOuterClick` maybe

src/components/LabelPicker/LabelPicker.tsx
eszthoff: do you mean you want to validate the name of the new label somehow? I think that can be done on the application side all together, since the actual adding happens on the user side (see story)

mukiienko: I agree that validation is a part of business logic, but I'm thinking about how we can notify the users that something is wrong with typed values. Probably we need to show some warning message in the label's popup. Also, in this scenario, we need to disable `Add` button.

eszthoff: after f2f discussion decided to leave it as it is

src/components/LabelPicker/LabelPicker.tsx
eszthoff: after f2f discussion decided to keep only 1 callback `onClose` that will be called regardless where the user clicked to close the component (trigger button, done button or outer click)

src/components/LabelPicker/LabelPicker.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const { labels, children, onChange, onAdd, onClose, inputPlaceholder, doneLabel } = props;
+    const [inputValue, setInputValue] = React.useState('');
+
+    const getChangeHandler = (label) => (event) => {
+        onChange(label, event);
+    };
+
+    const handleAdd = () => {
+        onAdd(inputValue);
+    };
```
mukiienko, line 50: Should we clear the input value after adding a new item?

src/components/LabelPicker/LabelPicker.tsx
```diff
@@ -0,0 +1,151 @@
...
+                                </Text>
+                            </Checkbox>
+                        ))}
+                        <div {...elem('inputLine')}>
+                            <Input
+                                placeholder={inputPlaceholder}
+                                size="small"
+                                onChange={handleInputChange}
+                                onKeyDown={handleKeyPress}
+                                {...elem('input')}
```
mukiienko, line 112: Let's add `isBlock` to the input

src/components/LabelPicker/__tests__/LabelPicker.spec.js
```diff
@@ -0,0 +1,367 @@
...
+                <LabelPicker labels={labelsMock} onClose={mockOnClose}>
+                    <Button>Click me</Button>
+                </LabelPicker>
+            );
+            wrapper.find('Button').simulate('click');
+
+            wrapper.find('Button').at(0).simulate('click');
+            expect(mockOnClose).toHaveBeenCalledTimes(1);
+        });
+    });
```
mukiienko, line 366: Nice tests coverage!

src/components/LabelPicker/LabelPicker.scss
```diff
@@ -0,0 +1,39 @@
...
+        z-index: 400;
+    }
+
+    &__container {
+        background-color: var(--color-background);
+        border: 1px solid var(--color-neutral);
+        border-radius: var(--border-radius);
+        box-shadow: 0 2px 2px 0 rgba(0, 0, 0, 0.1);
+        padding: var(--spacing-2x);
+    }
```
mukiienko, line 12: I think we need to add `max-width` for the popup, otherwise, the long label stretches the container or maybe we can allow the application itself to control block width via props/styles/classes. Also nice to have truncating for the long labels `...`
Screenshot 2022-04-26 at 12 15 20

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,15 +1,32 @@
...
 import styles from './Checkbox.scss';
 import { CHECKBOX_VIEWBOX } from '../../constants';
 
+interface Props extends React.InputHTMLAttributes<HTMLInputElement> {
+    /** A unique id to reference this checkbox */
+    id: string;
+    /** If the checkbox should be disabled */
+    disabled?: boolean;
+    /** The label for the checkbox */
+    children?: string | React.ReactElement;
```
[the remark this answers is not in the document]
eszthoff: I'm not sure how we should support this. Normally I think Checkbox is asking for a single node and this will have children as an array. The main thing is that we are adding a wrapper around the child if it is a string, otherwise not. In this case it is 2 children and the expected behaviour is not clear to me. See lines 21-28 in the same file. What would you expect to happen there?
I'm not sure how this passed validation so far, but I think we should stick with a single node as child. So you could wrap the whole thing in a `<span>` and then the current implementation works.

src/components/LabelPicker/__tests__/LabelPicker.spec.js
eszthoff: thanks :-))

mukiienko (approved):

PR #749, opened 2022-04-29
feat(Select): added option for no default selection

mukiienko:

# Checklist


src/components/SelectComponents/Select/__tests__/Select.spec.js
```diff
@@ -54,25 +51,72 @@ describe('Select', () => {
...
             expect(wrapper.find('li')).toHaveLength(0);
 
             // open items list
-            wrapper.find('.Select__dropdownIcon').at(0).simulate('click');
+            wrapper.find('svg').at(0).simulate('click');
             expect(wrapper.find('li')).toHaveLength(SUGGESTIONS.length);
 
-            // close items list
-            wrapper.find('.Select__dropdownIcon').at(0).simulate('click');
+            // select item
```
eszthoff, line 62: why this change? I think it is a good idea to see that clicking on the icon works. Also checking that selection closes the dropdown is a different test then implied in the description, and you have it (see 3 test below)

src/components/SelectComponents/Select/__tests__/Select.spec.js
```diff
@@ -54,25 +51,72 @@ describe('Select', () => {
...
+
     describe('callbacks', () => {
         describe('onFocus', () => {
-            it('should be called on clicking when opening the dropdown', () => {
-                expect(mockOnFocus).not.toHaveBeenCalled();
+            it('should be called on clicking when opening the dropdown', (done) => {
+                const focusedElement = wrapper.find('.Select__wrapper').getDOMNode();
+                const focusSpy = jest.spyOn(focusedElement, 'focus');
+
+                expect(focusSpy).not.toHaveBeenCalled();
```
eszthoff, line 112: you just defined the function, there was no interaction yet with the DOM, so this check will always work.

stories/Select.tsx
```diff
@@ -46,16 +46,24 @@ storiesOf('Organisms|Select Components', module)
...
+            };
+
             return (
                 <>
                     <Select<TSuggestion>
                         style={{ width: '650px' }}
                         items={SUGGESTIONS}
                         itemToString={SUGGESTION_TO_STRING}
                         selectedItem={store.get('selectedItem')}
+                        clearLabel={text('Clear button label', 'Clear')}
```
eszthoff, line 61: Please add two separate stories, one for clearable select and one for Select that always have a selectedItem

src/components/FieldWrapper/FieldWrapper.tsx
```diff
@@ -8,6 +9,10 @@ const { block, elem } = bem('FieldWrapper', styles);
 interface Props extends React.HTMLAttributes<HTMLDivElement> {
     /** content of the wrapper */
     children: NotEmptyReactNode;
+    /** show dropdown icon */
+    showArrow?: boolean;
+    /** set direction of dropdown icon (default drop down) */
+    isArrowDropUp?: boolean;
```
eszthoff, line 15: I find the `drop` confusing. Can it be just `isArrowUp`?

src/components/FieldWrapper/FieldWrapper.tsx
```diff
@@ -41,18 +55,27 @@ export const FieldWrapper: React.FC<Props> = React.forwardRef((props, ref) => {
...
                     context="link"
                     onClick={handleClear}
-                    {...elem('clearButton', props)}
+                    {...elem('clearButton', { ...props, rightIndent: showArrow })}
                 >
                     {clearLabel}
                 </Button>
             )}
+
+            {showArrow &&
```
eszthoff, line 64: I think this should be tested

src/components/SelectComponents/Select/Select.scss
```diff
@@ -1,26 +1,4 @@
...
-            position: absolute;
-            z-index: 400;
-            width: 100%;
-        }
-    }
-
-    &__field {
-        padding: 0;
-    }
-
```
antipin, line 23: nice clean up!

src/components/FieldWrapper/FieldWrapper.tsx
antipin: Same, I'd propose `isArrowPointingUp` or `isArrowUp`

antipin (changes requested):

src/components/SelectComponents/Select/__tests__/Select.spec.js
mukiienko: Actually, the arrow has the `pointer-events: none` style and it is not testable. For some reason, in the test environment, it opens when it was clicked, but it doesn't close.

src/components/SelectComponents/Select/__tests__/Select.spec.js
mukiienko: hmm.. you are right! There were a lot of broken tests after migration to `SelectBase` component. I spent more than half of the day fixing them and I didn't figure out how we can test "onFocus" in this case. It simply doesn't work. Do you think we need `onFocus` and `onBlur` for `Select` component?

stories/Select.tsx
mukiienko: Nice point!

src/components/FieldWrapper/FieldWrapper.tsx
mukiienko: `isArrowUp` sounds better

src/components/FieldWrapper/FieldWrapper.tsx
mukiienko: Done

src/components/SelectComponents/Select/Select.tsx
```diff
@@ -16,97 +14,103 @@ interface Props<S> extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange'
...
-                                suggestions={items}
-                                getItemProps={getItemProps}
-                                highlightedIndex={highlightedIndex}
-                                passDisabledToListItems
-                            />
-                        )}
-                    </List>
-                </FieldWrapper>
-            </div>
+        <div {...rest} {...block({ className })}>
```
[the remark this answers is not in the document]
mukiienko: I think we don't need `<div {...rest} {...block({ className })}> at all. In this case `className` is also redundant. @eszthoff what do you think about removing `div` wrapper and `className` prop? It will be breaking changes then. Pushed changes.

src/components/SelectComponents/Select/__tests__/Select.spec.js
eszthoff: In that case, perhaps better to comment out the original part, add a comment to fix it. Now it is a different test then what we want to do, and it is not clear unless someone looks into the details. So changes can be more easily missed. 

Should we consider moving interaction tests to Storybook? (maybe can think about it separately)

src/components/SelectComponents/Select/__tests__/Select.spec.js
eszthoff: I looked through our code and we have 1 place where we them.
I recall in other places that testing focus is not easy with enzyme. Maybe other testing platform. We have to move away anyway..

src/components/SelectComponents/Select/Select.tsx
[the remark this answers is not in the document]
eszthoff: Why is `className` redundant? We need that and we use that

src/components/FieldWrapper/__tests__/FieldWrapper.spec.js
```diff
@@ -19,6 +19,20 @@ describe('FieldWrapper', () => {
...
+        expect(toJson(wrapper)).toMatchSnapshot();
+    });
+    it('should render arrow icon pointing up', () => {
+        const wrapper = mount(
+            <FieldWrapper showArrow isArrowUp>
+                some children
+            </FieldWrapper>
+        );
+
+        expect(toJson(wrapper)).toMatchSnapshot();
```
eszthoff, line 34: I find  snapshot tests are good for general DOM layout but they can be too easily updated and hence can easily going wrong in the next iteration. Can you add checks to what in snapshot are we looking for?

stories/Select.tsx
```diff
@@ -71,6 +72,66 @@ storiesOf('Organisms|Select Components', module)
             },
         }
     )
+    .addParameters(
+        StoreInjector.withStore({
```
eszthoff, line 76: did you know that you can just use `React.useState` inside the story? I find this injector pattern very hard to read. I'm ok to leave it like this as you already wrote it, but maybe something to consider

src/components/SelectComponents/Select/__tests__/Select.spec.js
mukiienko: I commented out the original test and also added `TODO` description. I think it is a nice idea to have integration tests.

src/components/SelectComponents/Select/__tests__/Select.spec.js
mukiienko: I returned back the original test and skipped it. I remember I faced the same problem while I was working on `ComboboxMulti` which is also based on `SelectBase` and `Downshift`. The interesting part is that `onFocus` is not tested in `SelectBase` as well, so I think that problem comes from `SelectBase`.

src/components/SelectComponents/Select/Select.tsx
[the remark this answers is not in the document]
mukiienko: We don't use `className` approach for other Selects components (Autosuggest, ComboboxMulti), so I thought it is not used in our applications. I added  `className` to `SelectBase` and then `Select`.

stories/Select.tsx
```diff
@@ -53,6 +53,7 @@ storiesOf('Organisms|Select Components', module)
                         items={SUGGESTIONS}
                         itemToString={SUGGESTION_TO_STRING}
                         selectedItem={store.get('selectedItem')}
+                        placeholder={text('Placeholder', 'Choose item...')}
```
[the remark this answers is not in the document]
mukiienko: Done

stories/Select.tsx
mukiienko: Cool, I'll reimplement with `useState `.

eszthoff (changes requested):

eszthoff:
Maybe it is too late for this point, but I wonder why you decided to implement the arrow in FieldWrapper / SelectBase that is used in different places and not in the Select itself, as it was before, and as it is done in Combobox for instance.

src/components/FieldWrapper/__tests__/FieldWrapper.spec.js
mukiienko: It works! I thought we can find only HTML tags.

stories/Select.tsx
mukiienko: Yes, with `useState` the code is cleaner. I think we can use `useState` for other stories. I returned back `addParameters` with `StoreInjector` and created a follow-up task for replacing it https://jira.textkernel.nl/browse/ONEUI-302

src/components/SelectComponents/Select/__tests__/Select.spec.js
```diff
@@ -54,22 +54,72 @@ describe('Select', () => {
             clickWrapper();
             expect(wrapper.find('li')).toHaveLength(SUGGESTIONS.length);
         });
-        it('should toggle list when arrow element is clicked', () => {
+        it('should open list when arrow element is clicked', () => {
```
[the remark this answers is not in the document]
mukiienko: I created a separate test.

mukiienko:
> Maybe it is too late for this point, but I wonder why you decided to implement the arrow in FieldWrapper / SelectBase that is used in different places and not in the Select itself, as it was before, and as it is done in Combobox for instance.

The "clear" button is implemented `FieldWrapper` and it is hard to combine "clear" and "arrow" buttons together if they are in different components.

mukiienko:
> When I tried tested the `onFocus` call: it works in the clearable version as expected. but in the simple implementation for some reason, clicking on the icon directly doesn't trigger `onFocus`. Can you have a look?

Fixed, now it works for the both implementations. These changes also fixed `onFocus` test.

eszthoff:
> > Maybe it is too late for this point, but I wonder why you decided to implement the arrow in FieldWrapper / SelectBase that is used in different places and not in the Select itself, as it was before, and as it is done in Combobox for instance.
> 
> The "clear" button is implemented `FieldWrapper` and it is hard to combine "clear" and "arrow" buttons together if they are in different components.

should we use this option also in Combobox? maybe it is a quick fix can be done here, or if not, create a followup story (that we will never pick up with the current pace, but at least have a record for future generations :-P )

eszthoff:
Last thing, I think the PR title / commit message after squash should be `feat(Select): added option for no default selection`. Something like that, main point being:
* Select is included right after feat
* better be specific of what 'improvements' were made

eszthoff (approved):

mukiienko:
>

mukiienko:
> > > Maybe it is too late for this point, but I wonder why you decided to implement the arrow in FieldWrapper / SelectBase that is used in different places and not in the Select itself, as it was before, and as it is done in Combobox for instance.
> > 
> > 
> > The "clear" button is implemented `FieldWrapper` and it is hard to combine "clear" and "arrow" buttons together if they are in different components.
> 
> should we use this option also in Combobox? maybe it is a quick fix can be done here, or if not, create a followup story (that we will never pick up with the current pace, but at least have a record for future generations :-P )

It makes sense, I already was trying to do that, but it requires some changes in Combobox. I created task for refactoring ONEUI-303

PR #801, opened 2022-08-09
chore(Storybook): migrate to version 6

eszthoff:
Pending issues after the upgrade:
* CSS loaded in the wrong order (e.g. see Pagination story)
* Console logs no longer redirected to Actions tab. [Addon doesn't support v6](https://github.com/storybookjs/storybook-addon-console/issues/53). No solution at this time
* migrating text only stories (ArrangingOfZIndices) is incomplete. Tried to move to `.mdx` file, but was not loaded to the UI.
* should we update AutosuggestDeprecated story?


antipin:
Previous attempt: https://github.com/textkernel/oneui/pull/679

.storybook/main.js
```diff
@@ -0,0 +1,24 @@
+const path = require('path');
+const { getRules } = require('../scripts/build/webpack.config');
+
+module.exports = {
+    stories: ['../stories/**/*.@(js|jsx|ts|tsx)'],
```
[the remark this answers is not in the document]
eszthoff: yes, I tried that. But that didn't help with loading them. I think it is better to have that sorted out in the other ticket for it.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -6,6 +6,14 @@ import './DatePicker.scss';
...
 const DEFAULT_YEAR_RANGE = 100;
 
+/**
+ * ## Usage information
+ * This component is a thin wrapper around [react-datepicker](https://github.com/Hacker0x01/react-datepicker/),
+ * only setting the header element and add css.
+ *
+ * You can pass [props](https://github.com/Hacker0x01/react-datepicker/blob/master/docs/datepicker.md) according to their definition.
+ * See also [examples](https://reactdatepicker.com/) on their website.
+ */
```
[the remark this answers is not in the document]
eszthoff: yes, me too. :)

src/components/List/List.tsx
```diff
@@ -3,7 +3,7 @@ import { bem, scrollIntoViewIfNeeded } from '../../utils';
 import { LIST_NAVIGATION_DIRECTIONS, ENTER_KEY } from '../../constants';
 import styles from './List.scss';
 
-interface Props extends React.HTMLAttributes<HTMLUListElement> {
+export interface Props extends React.HTMLAttributes<HTMLUListElement> {
```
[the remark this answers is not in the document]
eszthoff: I can go through and rename them, just let me know if you want me to. Or is it better to come up with a standard in the other ticket maybe.

.storybook/preview-head.html
```diff
@@ -1,8 +0,0 @@
-<style>
-    body {
```
eszthoff, line 2: I forgot to migrate this bit. Maybe that's why the font is the wrong one.

.eslintrc.json
```diff
@@ -84,7 +84,9 @@
                 "react/display-name": "off",
```
mukiienko, line 84: I think in the current implementation we don't need `stories/index.ts`

.storybook/main.js
mukiienko, line 5: In the old version loading files was implemented via `index.ts` and we had a right overriding because stories were ordered in the right way in the `index.ts`. In the current implementation we still can control order by items in the array `stories`, but first we need to group our stories into subgroups/subfolders `Concepts`, `Theme`, `Atoms` etc (in the same way as we have for preview.js), for example:
```
 stories: ['../stories/theme/*.@(js|jsx|ts|tsx)', '../stories/**/*.@(js|jsx|ts|tsx)'],
```
It fixed the issue in Pagination story:
Screenshot 2022-08-19 at 13 19 22

src/components/List/List.tsx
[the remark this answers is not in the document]
mukiienko: Let's do that in an other ticket.

.storybook/main.js
eszthoff: I'm sure I tried this before, and it didn't work. But you are right, and it does work. Nice one!!!
I'll push a commit with this.

.eslintrc.json
eszthoff: removed it now, that the folders import works

.storybook/main.js
eszthoff: done

stories/organisms/AutosuggestDeprecated.js
```diff
@@ -0,0 +1,184 @@
+// import * as React from 'react';
```
andrentk, line 1: Is this file still needed?

.storybook/main.js
mukiienko: We still have a problem with overriding styles for `Pagination`, because these styles come from `Themeroller` that comes after `Pagination`, we need to put it on before high-level components:
```
    stories: [
        '../stories/theme/*.@(js|jsx|ts|tsx)',
        '../stories/atoms/*.@(js|jsx|ts|tsx)',
        '../stories/molecules/*.@(js|jsx|ts|tsx)',
        '../stories/organisms/*.@(js|jsx|ts|tsx)',
        '../stories/packages/*.@(js|jsx|ts|tsx)',
        '../stories/**/*.@(js|jsx|ts|tsx)',
    ],
```

stories/organisms/AutosuggestDeprecated.js
eszthoff: I was wondering about that. I was thinking a bit to keep it for internal reference. But perhaps that is not needed

.storybook/main.js
eszthoff: I'm actually not convinced that the theme should be at the beginning. Its whole purpose is to overwrite the styles, and then storybook will showcase the components as they will look like if adding that theme to the application. What do you think?

stories/organisms/AutosuggestDeprecated.js
eszthoff: removed it

.storybook/main.js
eszthoff: done

andrentk (approved):

PR #846, opened 2022-10-27
feat(NavItem): support react-router v6

eszthoff:
Story: [ONEUI-331](https://textkernel.atlassian.net/browse/ONEUI-331)

### DIFF
--- a/src/components/Navigation/NavItem/NavItem.tsx
+++ b/src/components/Navigation/NavItem/NavItem.tsx
@@ -9,18 +9,47 @@ export interface Props extends React.HTMLAttributes<HTMLDivElement> {
     pullRight?: boolean;
     /** add activeClassName to the rendered item. You should add this to react-routers NavLink */
     useActiveClass?: boolean;
+    /** React-Router version to determine how to handle useActiveClass */
+    routerVersion?: 5 | 6;
     /** Node to be rendered as a single navigation item. There should be only 1 child */
     children: React.ReactElement;
 }
 
+interface ChildProps extends Omit<Props, 'className' | 'children' | 'aria-current'> {
+    className?: string | (({ isActive }: { isActive: boolean }) => string);
+    activeClassName?: string;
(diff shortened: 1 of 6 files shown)


stories/molecules/Navigation.tsx
```diff
@@ -17,12 +17,12 @@ export const Navigation = (args) => (
...
             </NavItem>
-            <NavItem useActiveClass pullRight>
-                <NavLink exact to="/">
+            <NavItem useActiveClass pullRight routerVersion={6}>
+                <NavLink end to="/somewhere">
                     Item on the right
                 </NavLink>
             </NavItem>
-            <NavItem useActiveClass>
+            <NavItem useActiveClass routerVersion={6}>
```
mukiienko, line 25: The version is not selected in the "Controls" Storybook by default. When I choose 6 version I see item without styles. Is it ok?
Screenshot 2022-10-27 at 13 02 13

src/components/Navigation/NavItem/NavItem.tsx
```diff
@@ -9,18 +9,45 @@ export interface Props extends React.HTMLAttributes<HTMLDivElement> {
...
         newProps.activeClassName = elem('active', props).className;
+    } else {
+        const activeClass = elem('active', props).className as string;
+        newProps.className = ({ isActive }) => {
+            const classArr: string[] = [block(props).className];
+            if (isActive) {
+                classArr.push(activeClass);
+            }
+            return classArr.join(' ');
+        };
```
mukiienko, line 50: Now I'm thinking that probably was not a good idea to combine 2 versions in one component. There are a lot of `if` conditions and it's hard to review. But, let's keep it as it is.

eszthoff: I hope we can remove support for v5 later, then it will look ok again

mukiienko (approved):

PR #866, opened 2022-12-12
refactor: add props export

PavelKirvelTextkernel:

# Checklist


src/index.ts
```diff
@@ -8,64 +8,116 @@ export * from './packages/ThemeGenerator';
...
+    ListProps,
+    ListItem,
+    ListItemProps,
+    ListActions,
+    ListActionsProps,
+    ListOptimizer,
+    ListOptimizerProps,
+} from './components/List';
+export { Modal, ModalProps } from './components/Modal';
 export { MapWithGoogleLoader, Map } from './components/Map';
```
mukiienko, line 44: What about `Map` types?

src/index.ts
```diff
@@ -8,64 +8,116 @@ export * from './packages/ThemeGenerator';
...
+export { Callout, CalloutProps } from './components/Callout';
+export { CandidateAvatar, CandidateAvatarProps } from './components/CandidateAvatar';
+export { Chip, ChipProps } from './components/Chip';
+export { ContentPlaceholder, ContentPlaceholderProps } from './components/ContentPlaceholder';
+export { Heading, HeadingProps } from './components/Heading';
+export {
+    StringHighlighter,
+    StringHighlighterProps,
+    ReactElementHighlighter,
+    ReactElementHighlighterProps,
```
mukiienko, line 28: What about other dependencies types, for example, `ReactElementHighlighterTermResult`? the same for other components.

src/components/Checkbox/index.ts
```diff
@@ -1 +1 @@
-export { Checkbox } from './Checkbox';
+export { Checkbox, Props as CheckboxProps } from './Checkbox';
```
mukiienko, line 1: What about exporting from `packages`, `utils`, `constants`?

PavelKirvelTextkernel: I've added export for constants, packages have been already exported and utils (default) about others utils i don't think we need to export them

PavelKirvelTextkernel:
Task: https://textkernel.atlassian.net/browse/ONEUI-172

mukiienko:
@PavelKirvelTextkernel could you please update `CONTRIBUTING.md` (remove the part about react propTypes, defaultValues, etc and add  information about exposing TS types)

src/components/Checkbox/index.ts
eszthoff: what about the types in `types/types.d.ts`?

src/index.ts
```diff
@@ -6,66 +6,123 @@ export * from './packages/Highlighter';
...
 
 // Atoms
-export { Button, SearchButton, StepperButton } from './components/Buttons';
-export { Callout } from './components/Callout';
-export { CandidateAvatar } from './components/CandidateAvatar';
-export { Chip } from './components/Chip';
-export { ContentPlaceholder } from './components/ContentPlaceholder';
-export { Heading } from './components/Heading';
-export { StringHighlighter, ReactElementHighlighter } from './components/Highlighter';
+export {
```
eszthoff, line 14: would it make sense to use `export * from '...'` everywhere instead of listing everything one by one?

PavelKirvelTextkernel: I think yes

src/components/Checkbox/index.ts
PavelKirvelTextkernel: I think we don't need it, also it's not a module

eszthoff: It is used in our components. In applications using OneUI, TS cannot validate these props, as the definition is not exported, and TS doesn't know what they mean.

PavelKirvelTextkernel: @mukiienko @eszthoff Is it will be okay if after add export for types/types.d.ts that we need to import  types to components in OneUI?

mukiienko: I think it should be a way to expose `d.ts files in the build via Webpack.

PavelKirvelTextkernel: @mukiienko as i understand the idea is that we can see types here
Screenshot 2022-12-19 at 13 43 30

PavelKirvelTextkernel:
For use export * from '...' for './components/Map' needs to migrate to TS. I've got some warnings/errors I would like to do it in a separate task

src/components/Checkbox/index.ts
PavelKirvelTextkernel: @mukiienko @eszthoff 
I've added solution with copy-webpack-plugin ( https://webpack.js.org/plugins/copy-webpack-plugin/).  Is it wiil be okay to use ?
I also have another solution migrate types.d.ts to  .ts file and add importing in files where we use it

mukiienko (approved):

eszthoff:
When I try `npm run build` there are some warnings regarding types. 
```
WARNING in ./index.ts 25:0-98
"export 'CircularMarker' was not found in './components/Map'

WARNING in ./index.ts 25:0-98
"export 'MapProps' was not found in './components/Map'

WARNING in ./index.ts 25:0-98
"export 'RegionArea' was not found in './components/Map'
```

eszthoff:
I wonder what will happen when we import this version to the application. Will `.d.ts` files interpreted as global? Will we have possible type crashes with things defined in the application? I think it is normally solved with namespaces, but I don't think we are using that now. Have you looked into this?

src/components/Checkbox/index.ts
eszthoff: I tried to build locally and had a quick look at the dist files. I still see that these types are not resolved there. E.g. `dist/dts/src/components/BulkActionsToolbar/BulkActionsToolbar.d.ts` have red highlights

Screenshot 2022-12-27 at 14 45 42

eszthoff (changes requested):

PR #867, opened 2022-12-14
feat(Combobox): new component

eszthoff:
Story: [ONEUI-138](https://textkernel.atlassian.net/browse/ONEUI-138)

# Checklist


andrentk (approved):

mukiienko:
The tab focus is different for Combox and for other simar input base components. The Combobox has a focused arrow when others have a blue border around the inputs. Maybe it's ok.
Screenshot 2022-12-21 at 11 44 14
Screenshot 2022-12-21 at 11 44 05

mukiienko:
The heights of components are not aligned:

https://user-images.githubusercontent.com/19436304/208888018-d6fd00ca-010c-4c73-8244-c8d8d5a1f4a3.mov

eszthoff:
> The tab focus is different for Combox and for other simar input base components. The Combobox has a focused arrow when others have a blue border around the inputs. Maybe it's ok. Screenshot 2022-12-21 at 11 44 14 Screenshot 2022-12-21 at 11 44 05

It is the same behaviour as for Select, which also has the same arrow. But I think you are right, and this change is the side-effect of making that arrow accessible via keyboard. I had a quick look and didn't see how to fix this easily. On the other hand, I think it is fine. If you are using the keyboard, it does make sense that you end up on the toggle button and you have to click it in order to actually open the dropdown. 
I can change ComboboxMuli to use the same button as well, for better alignment.

eszthoff:
> The heights of components are not aligned

Nice catch. Maybe something we can look at later. I see many "magic numbers" in our css when it comes to height of components. That might be funny also if the theme roller changes the other size variables (e.g. set the font to much bigger, or the spacers) but these will not be effected. Shall I make a ticket for the future?

For now, I made autosuggest 2px bigger, as the other ones were all the same size

mukiienko:
> > The tab focus is different for Combox and for other simar input base components. The Combobox has a focused arrow when others have a blue border around the inputs. Maybe it's ok.
> 
> It is the same behaviour as for Select, which also has the same arrow. But I think you are right, and this change is the side-effect of making that arrow accessible via keyboard. I had a quick look and didn't see how to fix this easily. On the other hand, I think it is fine. If you are using the keyboard, it does make sense that you end up on the toggle button and you have to click it in order to actually open the dropdown. I can change ComboboxMuli to use the same button as well, for better alignment.

I'd prefer to have a blue border for all input base components. But I think it's a minor thing so we can make adoption it later.

mukiienko:
> > The heights of components are not aligned
> 
> Nice catch. Maybe something we can look at later. I see many "magic numbers" in our css when it comes to height of components. That might be funny also if the theme roller changes the other size variables (e.g. set the font to much bigger, or the spacers) but these will not be effected. Shall I make a ticket for the future?
> 
> For now, I made autosuggest 2px bigger, as the other ones were all the same size

The different heights can break the layout if we have these components in one form, but we are not going to use the new Combobox at least now, so I am fine with a follow-up ticket.

eszthoff:
> The different heights can break the layout if we have these components in one form, but we are not going to use the new Combobox at least now, so I am fine with a follow-up ticket.

Changed it so now all Select-like components have 36px height

mukiienko (approved):

PR #883, opened 2023-01-09
feat: add custom types

PavelKirvelTextkernel:

# Checklist


PavelKirvelTextkernel:
After some investigation:
1. We can copy current types.d.ts to build folder but in this case we will  have red highlights(we can check type value, need to check in another app is it possible to jump in types.d.ts).
2. Rewrite typs.d.ts to types.ts/customTypes.ts and add importing where we use them.

mukiienko:
> After some investigation:
> 
> 1. We can copy current types.d.ts to build folder but in this case we will  have red highlights(we can check type value, need to check in another app is it possible to jump in types.d.ts).
> 2. Rewrite typs.d.ts to types.ts/customTypes.ts and add importing where we use them.

Hi, I didn't get your comment. How is it related with 3rd party libraries, for example, react-model?

PavelKirvelTextkernel:
About types: 
https://stackoverflow.com/questions/51983175/usage-of-types-and-interface-without-importing

PavelKirvelTextkernel:
For global type(without importing) i've tried this: export type EmptyElement = false | null | undefined;

declare global {
    // eslint-disable-next-line @typescript-eslint/naming-convention
    type emp = EmptyElement;
}
But still get red highlight

PR #907, opened 2023-03-03
feat: set up process to measure migration progress

PavelKirvelTextkernel:

# Checklist


scripts/measure-enzyme-to-rtl-tests-ratio/measure-enzyme-to-rtl-tests-ratio.js
```diff
@@ -0,0 +1,82 @@
+const { spawn } = require('child_process');
+
+const API_URL = 'https://sheetdb.io/api/v1/g3uyrf24ruo1v';
```
antipin, line 3: The `g3uyrf24ruo1v` part of the URL looks like a secret/token. If so, it should be invalidated, and new token should be passed via [GitHub encrypted secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) mechanism.

It's a public repo, so no secrets should be committed to it.

scripts/measure-enzyme-to-rtl-tests-ratio/measure-enzyme-to-rtl-tests-ratio.js
```diff
@@ -0,0 +1,82 @@
...
+    const columnData = await fetch(API_URL)
+        .then((response) => {
+            return response.json();
+        })
+        .then((data) => {
+            return data;
+        })
+        .catch((error) => {
+            console.error('SheetDB data fetch error: ', error);
+        });
```
antipin, line 19: Two different styles of hadeling async code are mixed here: `async/await` and `Promises`.

If `async/await` is chosen, then we shouldn't use `.then() / .catch()`. The code should look like this (not tested):

```ts
async function fetchColumnData(columnName) {
    try {
        const columnDataResponse = await fetch(API_URL);
        return await columnDataResponse.json();
    } catch (error) {
        console.error('SheetDB data fetch error: ', error);
        return null;
    }
}


async function getColumnData(columnName) {
    const columnData = await fetchColumnData(columnName);
    if (columnData === null) {
        // handle the error
    }
    for (let i = 0; i < columnData.length; i += 1) {
        const currentColumnData = columnData[i];
        if (columnName === ENZYME_COLUMN && currentColumnData.enzyme.includes(ID_PATTERN)) {
            return currentColumnData.enzyme;
        }
        if (columnName === RTL_COLUMN && currentColumnData.rtl.includes(ID_PATTERN)) {
            return currentColumnData.rtl;
        }
    }
    return NOT_VALID_ID_PATTERN;
}
```

antipin (changes requested):
Partially reviewed,

scripts/measure-enzyme-to-rtl-tests-ratio/measure-enzyme-to-rtl-tests-ratio.js
mukiienko: Actually, why do we need to use some external sheets if we can save the output in the current repo?

antipin:
@PavelKirvelTextkernel @mukiienko  I've tested the script once again and found a few issues.

For example, the number for Enzym was in fact total number or tests.

Also added some improvements:
- better error hadnling
- one request to spreadsheet API instead of 2
- POST instead of PATCH
- don't count same test file twice id there are multiple RTL import patterns
- mistake in `find` command (`.spec.tsx` files were missing)

Please replace the script with the code below, feel free to test it once again, and let me know if you have any questions.

```ts
const { execSync } = require('child_process');

const API_URL = 'https://sheetdb.io/api/v1';

// todo: replace with github secrets
function getToken() {
    const values = [
        'v',
        'v1',
        '1',
        '7',
        'o',
        'q',
        'u',
        'u1',
        'r',
        'r1',
        '4',
        '21',
        '2',
        '44',
        'f',
        'f1',
        'r',
        'r1',
        'y',
        'y1r',
        'u',
        'u1',
        '3',
        '313',
        'g',
    ];
    return values
        .reduce((result, item, idx) => {
            if (idx % 2 === 0) {
                return result + item;
            }
            return result;
        }, '')
        .split('')
        .reverse()
        .join('');
}

async function insertRow(url, insertData) {
    const response = await fetch(url, {
        method: 'POST',
        headers: {
            Accept: 'application/json',
            'Content-Type': 'application/json',
        },
        body: JSON.stringify({
            data: {
                insertData,
            },
        }),
    });

    if (response.status < 200 || response.status >= 300) {
        const message = await response.text();
        throw new Error(`Unable to insert a row. Request failed: ${message}`);
    }

    return response.json();
}

async function saveMetricsToSpreadsheet({ total, rtl }) {
    const url = `${API_URL}/${getToken()}`;
    const date = new Date();
    const day = date.getUTCDate();
    const month = date.getUTCMonth() + 1;
    const year = date.getUTCFullYear();
    await insertRow(url, {
        date: `${day}/${month}/${year}`,
        enzyme: total - rtl,
        rtl,
        progress: `=${rtl}/${total}`,
    });
}

async function execute() {
    const findAllTestFilesCommand =
        'set -o pipefail; find ./src -mindepth 1 -type f \\( -iname \\*.spec.js -o -iname \\*.spec.jsx -o -iname \\*.spec.ts -o -iname \\*.spec.tsx \\) | sort --unique';

    const totalTests = execSync(findAllTestFilesCommand);
    const rtlTests = execSync(
        `set -o pipefail; ${findAllTestFilesCommand} | xargs grep -E "from '@testing-library/react'|from 'react-test-renderer'" | sort --unique`
    );

    const totalMetric = totalTests
        .toString('utf-8')
        .split('\n')
        .filter((item) => item.startsWith('./src/')).length;
    const rtlMetric = rtlTests
        .toString('utf-8')
        .split('\n')
        .filter((item) => item.startsWith('./src/')).length;

    console.log(
        `Metrics were extracted successfully. enzymeMetric = ${totalMetric}; rtlMetric = ${rtlMetric}`
    );

    try {
        await saveMetricsToSpreadsheet({
            total: totalMetric,
            rtl: rtlMetric,
        });
        console.log('Metrics were saved successfully');
    } catch (error) {
        console.error(`Failed while saving the metrics: ${error.message}`);
        process.exit(1);
    }
}

(async () => execute())();
```

PR #943, opened 2023-04-19
test: migrate button tests to rtl

eszthoff:
Story: https://textkernel.atlassian.net/browse/ONEUI-355


src/components/ButtonGroup/__tests__/ButtonGroup.spec.tsx
```diff
@@ -1,33 +1,44 @@
...
+        const view = render(
             <ButtonGroup>
                 <Button key="a">A button</Button>
                 <Button key="b">Another button</Button>
             </ButtonGroup>
         );
-        expect(toJson(wrapper)).toMatchSnapshot();
+        expect(view.asFragment()).toMatchSnapshot();
+        const button = screen.getAllByRole('button');
+        expect(button[0]).toHaveClass('ButtonGroup__button--first');
```
PavelKirvelTextkernel, line 19: I think we can apply code style add one empty line.
// Act
// Empty line
// Assert

 expect(view.asFragment()).toMatchSnapshot();
const button = screen.getAllByRole('button');
// empty line
expect(button[0]).toHaveClass('ButtonGroup__button--first');
...

@eszthoff What do you think ?

PavelKirvelTextkernel (changes requested):

src/components/ButtonGroup/__tests__/ButtonGroup.spec.tsx
eszthoff: yes, that's a good point. I will change that

src/components/ButtonGroup/__tests__/ButtonGroup.spec.tsx
```diff
@@ -1,33 +1,44 @@
...
     it('should render default button correctly', () => {
-        const wrapper = mount(
+        const view = render(
             <ButtonGroup>
                 <Button key="a">A button</Button>
                 <Button key="b">Another button</Button>
             </ButtonGroup>
         );
-        expect(toJson(wrapper)).toMatchSnapshot();
+        expect(view.asFragment()).toMatchSnapshot();
```
andrentk, line 17: I think it is more clean and htmlish if we use `view.container` which returns an HTMLElement to check in the snapshot. What do you think?

eszthoff: I don't have a strong preference, and I haven't seen anything in the best practices on this. I agree with you, that it seems more natural to call `view.container` the `view.asFragment()`.  @PavelKirvelTextkernel  what do you think?

PavelKirvelTextkernel: @eszthoff @andrentk  I think yes we can use it

PavelKirvelTextkernel (approved):

src/components/ButtonGroup/__tests__/ButtonGroup.spec.tsx
eszthoff: Ok, thanks. I will make a separate PR to change this in the docs and already existing tests.

PR #946, opened 2023-04-20
fix: RadioButton Props extends

andrentk:
Story: https://textkernel.atlassian.net/browse/ONEUI-377


antipin (approved):
Looks good!

antipin:
@eszthoff yes, it solves the issue in source-ui, now it gives error if we add `rev`:

image

eszthoff (approved):

PR #955, opened 2023-04-26
test: migrate labelPicker components tests migration

PavelKirvelTextkernel:
Story: [ONEUI-363](https://textkernel.atlassian.net/browse/ONEUI-363)


src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
```diff
@@ -0,0 +1,340 @@
...
+    );
+};
+
+const rerenderLabelPicker = (props) => {
+    view = render(
+        <LabelPicker {...defaultProps} {...props}>
+            <Button>Click me</Button>
+        </LabelPicker>
+    );
+};
```
andrentk, line 57: beforeEach can handle the first render which renders the default component and using that we don't need to call in every test renderLabelPicker and looks cleaner IMO.

calling rerenderLabelPicker when it doesn't rerender can be misleading, because here it is a normal render. If it is a rerender we should use `rerender` instead of `render`. 

In the tests where we need to pass different props we can just rerender the component with the desired props using `rerender`.

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
```diff
@@ -0,0 +1,340 @@
...
+            await userEvent.click(screen.getByRole('button', { name: 'Click me' }));
+
+            expect(onClick).toHaveBeenCalledTimes(1);
+        });
+    });
+    describe('toggling dialog visibility', () => {
+        it('should toggle dialog when trigger button is clicked', async () => {
+            renderLabelPicker();
+            expect(view.container).toMatchSnapshot();
+            expect(screen.queryAllByRole('dialog').length).toBe(0);
```
andrentk, line 96:
```suggestion
            expect(screen.queryAllByRole('dialog')).toHaveLength(0);
```

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
```diff
@@ -0,0 +1,340 @@
...
+            await userEvent.click(screen.getByRole('button', { name: 'Click me' }));
+
+            expect(screen.getAllByRole('dialog').length).toBeGreaterThan(0);
+
+            // click outside of component
+            act(() => {
+                clickDocument('click');
+            });
+
+            expect(screen.queryAllByRole('dialog').length).toBe(0);
```
andrentk, line 124: same as above

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+            // open dialog
+            await userEvent.click(screen.getByRole('button', { name: 'Click me' }));
+
+            expect(screen.queryAllByRole('dialog').length).toBeGreaterThan(0);
+
+            // click outside of component
+            act(() => {
+                clickDocument('click');
```
andrentk, line 323: This was initially done to be used with Enzyme. Maybe there is a different approach to this using RTL.

andrentk (commented):
nit: add a line between the different tests

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/LabelPicker/__tests__/LabelPicker.spec.tsx
PavelKirvelTextkernel: @andrentk Do you mean to replace  clickDocument('click') or just remove act ?

andrentk: I meant for clickDocument. I think act is already part of userEvent, not 100% on the later.

PavelKirvelTextkernel: :+1:

andrentk (approved):

PR #958, opened 2023-04-27
fix: unintentional breaking changes where components don't render with children 0

eszthoff:
Story: [JF-3938](https://textkernel.atlassian.net/browse/JF-3938)


andrentk (approved):
Nice catch 😃

PR #959, opened 2023-05-01
test: migrate datepicker tests

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-357)


src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
```diff
@@ -1,16 +1,19 @@
...
     const decreaseMonth = jest.fn();
     const increaseMonth = jest.fn();
     const changeYear = jest.fn();
-    let wrapper;
+    let view;
 
     beforeEach(() => {
-        wrapper = mount(
+        view = render(
+            // @ts-ignore
```
andrentk, line 16: why // ts-ignore?

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
```diff
@@ -22,42 +25,64 @@ describe('CalendarHeader', () => {
...
+                date={new Date('2021-12-31')}
+                decreaseMonth={decreaseMonth}
+                increaseMonth={increaseMonth}
+                changeYear={changeYear}
+            />
+        );
 
-        expect(wrapper.html()).toBe(null);
         expect(consoleErrorSpy).toHaveBeenCalledTimes(1);
         expect(consoleErrorSpy.mock.calls[0][0].replace(/\s/g, '')).toMatch(
```
andrentk, line 48: Do we need to replace the expected and the actual values the same thing? If we don't replace in both it should work no?

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
```diff
@@ -22,42 +25,64 @@ describe('CalendarHeader', () => {
...
             `CalendarHeader component has received invalid props.
         Minimum selectable year (9) is larger then maximum selectable year (2)`.replace(/\s/g, '')
         );
     });
+
     it('should fill in the years correctly', () => {
-        const options = wrapper.find('option');
+        const options = screen.getAllByRole('option');
         // eslint-disable-next-line no-plusplus
```
andrentk, line 56: If we have in place the rule to not use ++ lets use the other way. If we should use the plus plus, let's change the rules in a follow up ticket to remove this comment.

src/components/DatePicker/__tests__/DatePicker.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+        // screen.getByRole('option');
+        expect(screen.getByRole('presentation')).toBeInTheDocument();
+    });
+
+    it('should set min and max date on calendar even if they are no originally passed as props', async () => {
+        const user = userEvent.setup();
+        // this is to make sure the calendar navigation and the year selection in the header are always aligned
+        await openCalendar(user);
+
+        expect(user);
```
andrentk, line 52: This will be always green or am I mistaken?

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: It asks to provide all properties:
TS2740: Type '{ yearsRange: [number, number]; date: Date; decreaseMonth: Mock ; increaseMonth: Mock ; changeYear: Mock ; }' is missing the following properties from type 'Props': monthDate, changeMonth, customHeaderCount, prevMonthButtonDisabled, and 5 more.

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: Without it returns 
image

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: I've create a ticket for that. Could you please check it https://textkernel.atlassian.net/browse/ONEUI-402

src/components/DatePicker/__tests__/DatePicker.spec.tsx
PavelKirvelTextkernel: Yes, but it's not clear what we expect here. I think we should add
expect(screen.getByText(minDate.getFullYear())).toBeInTheDocument();
expect(screen.getByText(maxDate.getFullYear())).toBeInTheDocument();

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
andrentk: If it is a ticket just to remove this comment and change the line below to not use ++ I think we don't need a ticket and just do it here. I was more referring if we should look at the eslint rules file regarding this rule (which we might not need to do).

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
andrentk: What is the reason for not providing them all?

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: Do you mean to add to ?
image

src/components/DatePicker/__tests__/DatePicker.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    it('should render correctly', () => {
+        expect(view.container).toMatchSnapshot();
+        expect(screen.getByRole('textbox')).toBeInTheDocument();
+    });
+
+    it('should render calendar header correctly when calendar is open', async () => {
+        const user = userEvent.setup();
+        await openCalendar(user);
+        // taking full snapshot of the calendar fails in CD/CI.
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: I don't get it. I've checked pipeline and it's passed

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: I've replaced it to year += 1

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/DatePicker/__tests__/DatePicker.spec.tsx
PavelKirvelTextkernel: I've update it, could you please check it

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+describe('CalendarHeader', () => {
+    const YEARS = [2016, 2025] as [number, number];
+    const decreaseMonth = jest.fn();
+    const increaseMonth = jest.fn();
+    const changeYear = jest.fn();
+    const changeMonth = jest.fn();
+    const decreaseYear = jest.fn();
+    const increaseYear = jest.fn();
+    const currentMonth = new Date();
+    let view;
```
andrentk, line 16: My TS was complaining that view has no type defined, it is any at this moment. It should have type of RenderResult I think

src/components/DatePicker/__tests__/DatePicker.spec.tsx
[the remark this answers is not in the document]
andrentk: Looks like the CD/CI failed when we added a snapshot of the calendar with enzyme. My question was if the same still happens with RTL. If not we can include the snapshot in the test, otherwise just leave the comment here as it is.

src/components/DatePicker/__tests__/DatePicker.spec.tsx
andrentk: But do we need this `expect(user)` expectation? What is the reasoning for that?

src/components/DatePicker/__tests__/DatePicker.spec.tsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+import MockDate from 'mockdate';
+import { DatePicker } from '../DatePicker';
+
+describe('DatePicker', () => {
+    let view;
```
andrentk, line 9: same as above.

src/components/DatePicker/__tests__/DatePicker.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/DatePicker/CalendarHeader/__tests__/CalendarHeader.spec.tsx
PavelKirvelTextkernel: I've added, but i don't get any warnings

src/components/DatePicker/__tests__/DatePicker.spec.tsx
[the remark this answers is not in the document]
PavelKirvelTextkernel: Added toMatchSnapshot. But what about these:
    // Some small time difference in dates - needs further investigation.
        // Since testing 3rd party library is not a good practice anyway,
        // we will just check that our custom header is correct
        // screen.getByRole('option');
I've removed them and added some cases. If i understand them correct.

src/components/DatePicker/__tests__/DatePicker.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/DatePicker/__tests__/DatePicker.spec.tsx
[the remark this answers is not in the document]
andrentk: About those I don't know.

andrentk (approved):

PR #960, opened 2023-05-02
test: migrate Highlighter tests

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-361)


src/components/Highlighter/StringHighlighter/__tests__/StringHighlighter.spec.tsx
```diff
@@ -1,21 +1,21 @@
...
     const highlightRenderer = ({ substring, ...props }) => <b {...props}>{substring}</b>;
     // eslint-disable-next-line react/display-name, react/prop-types
     const renderComponent = (props) => {
-        return mount(
+        return render(
             <StringHighlighter
                 string={string}
                 searchTerms={searchTerms}
-                highlighterCoreOptions={(ignoreDiacritics, ignoreCase)}
+                highlighterCoreOptions={ignoreDiacritics}
```
PavelKirvelTextkernel, line 18: Don't need to use const ignoreCase = true; because it's true by default

src/components/Highlighter/StringHighlighter/__tests__/StringHighlighter.spec.tsx
```diff
@@ -1,83 +1,100 @@
...
-        return mount(
-            <StringHighlighter
-                string={string}
-                searchTerms={searchTerms}
-                highlighterCoreOptions={(ignoreDiacritics, ignoreCase)}
-                highlightRenderer={highlightRenderer}
-                {...props}
-            />
-        );
+    const defaultProps = {
```
PavelKirvelTextkernel, line 11: Don't need to use const ignoreDiacritics = true; because it's true by default

PavelKirvelTextkernel: Yes, but it's not clear what we expect here. I think we should add 
 expect(screen.getByText(minDate.getFullYear())).toBeInTheDocument();
 expect(screen.getByText(maxDate.getFullYear())).toBeInTheDocument();

src/components/Highlighter/StringHighlighter/__tests__/StringHighlighter.spec.tsx
```diff
@@ -1,83 +1,100 @@
...
+
         it('should not highlight not existed terms', () => {
-            const wrapper = renderComponent({
+            rerenderView({
                 searchTerms: ['notExisting', 'dev', 'a'],
             });
-            expect(toJson(wrapper)).toMatchSnapshot();
+
+            expect(view.container).toMatchSnapshot();
+            expect(screen.getByText(string)).toBeInTheDocument();
```
eszthoff, line 38: I really like that you added specific tests like that, I think these make them a lot more resilient. Thanks!
However, it is a bit hidden logic why we are looking for the string and not for the highlights. Since the highlighter tag is only in the tests here, maybe it is ok to use getByTestId, and add the test id to the `highlightRenderer` at the top. Then here and in the tests below you can check how many of those are present. What do you think?

eszthoff (changes requested):

src/components/Highlighter/StringHighlighter/__tests__/StringHighlighter.spec.tsx
PavelKirvelTextkernel: :+1:

eszthoff: nicely done. thanks

eszthoff (approved):

PR #961, opened 2023-05-02
test: migrate Drawer tests.

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-359)


PavelKirvelTextkernel:
@eszthoff @andrentk 
About this requirement: **Might need to think of a different way to do these. Example should pass initial close status should check if the Drawer is closed instead of checking class names**
What do you think if we check how many times onClick button was called ?

src/components/Drawer/Drawer.tsx
```diff
@@ -93,6 +93,7 @@ export const Drawer: React.FC<Props> = ({
                     {...block({ [getStatus()]: state !== 'entering' })}
                     aria-hidden={!isShown}
                     {...rest}
+                    role="group"
```
PavelKirvelTextkernel, line 96: I've put "group" role here is it okay or should we use other one ?

src/components/Drawer/__tests__/__snapshots__/Drawer.spec.tsx.snap
```diff
@@ -1,141 +1,3 @@
...
-          className="Drawer__body"
-        >
-          some text
-        </section>
-      </div>
-    </mockConstructor>
-  </Portal>
-</Drawer>
-`;
+exports[`Drawer should render correctly 1`] = `<div />`;
```
eszthoff, line 3: Is this really the HTML we expect to be rendered?

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
 import { Drawer } from '../Drawer';
 
 jest.mock('react-transition-group', () => {
     const FakeTransition = jest.fn(({ children }) => children({ state: 'entered' }));
     return { Transition: FakeTransition };
 });
+const onClickMock = jest.fn();
```
eszthoff, line 11: we normally define this for each test or in the `beforeEach` block. To make it global can lead to errors, as the mock might not reset correctly between tests.

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
             </Drawer>
         );
+        const expandButton = screen.getByRole('button', { name: '' });
 
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(true);
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(false);
-        expect(
-            wrapper.find('.Drawer__expandButton').hasClass('Drawer__expandButton--isExpanded')
-        ).toBe(true);
+        expect(expandButton).toBeInTheDocument();
```
eszthoff, line 29: is this button only shown when the drawer is expanded?

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
+        expect(onClickMock).not.toHaveBeenCalled();
```
eszthoff, line 30: why do we need this check? If we didn't do any user interaction, would the code trigger itself?

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
+                some text
+            </Drawer>
+        );
 
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(true);
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(false);
-        expect(
-            wrapper.find('.Drawer__expandButton').hasClass('Drawer__expandButton--isExpanded')
-        ).toBe(false);
+        expect(onClickMock).not.toHaveBeenCalled();
```
eszthoff, line 40: Can you explain why click would have been called, and how does this check satisfy the test description `should pass initial close status`.

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
+        const expandButton = screen.getByRole('button', { name: '' });
         // Expand Drawer
-        wrapper.find('.Drawer__expandButton').simulate('click');
-
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(true);
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(false);
-        expect(
-            wrapper.find('.Drawer__expandButton').hasClass('Drawer__expandButton--isExpanded')
-        ).toBe(true);
+        await user.click(expandButton);
```
eszthoff, line 52: I think there should be some assumptions here to check that the drawer now is expanded

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
-        wrapper.find('.Drawer__expandButton').simulate('click');
+        await user.click(expandButton);
 
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(true);
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(false);
-        expect(
-            wrapper.find('.Drawer__expandButton').hasClass('Drawer__expandButton--isExpanded')
-        ).toBe(false);
+        expect(onClickMock).toHaveBeenCalledTimes(2);
+        expect(screen.getByRole('group', { hidden: true })).toHaveAttribute('aria-hidden', 'false');
```
eszthoff, line 58: looks confusing. If the group has `hidden: true` why would the aria label say the opposite?

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
-        const wrapper = mount(
+        render(
             <Drawer isExpanded title="some title">
                 some text
             </Drawer>
         );
 
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(true);
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(false);
+        expect(screen.getByRole('group', { hidden: true })).toHaveAttribute('aria-hidden', 'false');
```
eszthoff, line 88: s/a (hidden both true and false)

src/components/Drawer/__tests__/Drawer.spec.tsx
```diff
@@ -1,108 +1,104 @@
...
+        const expandButton = screen.getByRole('button', { name: '' });
+        await user.click(expandButton);
 
-        wrapper.find('.Drawer__expandButton').simulate('click');
-
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndExpanded')).toBe(false);
-        expect(wrapper.find('.Drawer').hasClass('Drawer--isShownAndClosed')).toBe(true);
-        expect(onClickMock).toHaveBeenCalled();
+        expect(onClickMock).toHaveBeenCalledTimes(1);
+        expect(screen.getByRole('group', { hidden: true })).toHaveAttribute('aria-hidden', 'false');
```
eszthoff, line 102: s/a

eszthoff (changes requested):

src/components/Drawer/Drawer.tsx
eszthoff: Good question. Looking at the docs I think it is a live-region and us such should have one of the associated roles. see https://www.w3.org/TR/wai-aria-1.2/#ref-for-dfn-live-region-1.

src/components/Drawer/__tests__/Drawer.spec.tsx
PavelKirvelTextkernel: Removed hidden: true

src/components/Drawer/__tests__/Drawer.spec.tsx
PavelKirvelTextkernel: Removed hidden: true

src/components/Drawer/__tests__/Drawer.spec.tsx
PavelKirvelTextkernel: Removed hidden: true

src/components/Drawer/__tests__/Drawer.spec.tsx
PavelKirvelTextkernel: No. I've changed it to check class name

src/components/Drawer/__tests__/Drawer.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/Drawer/__tests__/Drawer.spec.tsx
PavelKirvelTextkernel: When i do MatchSnapshot it returns the same result as for **should render correctly** case

src/components/Drawer/Drawer.tsx
mukiienko: hmm, really a good question :) It looks like for Source we can use role"status" with aria-live="polite". On another hand, the Drawer component from the OneUI perspective is a collapsible element that can wrap any data from the application, so I think the application should decide which role is most suitable. As far as I know, a reader reads all text inside the aria-live block when the element appeared. In this scenario, all elements will be read, but we need only read the last added items, thus we need to add aria-live only to the items (application level). 
My conclusion:

- use role="group" for the Drawer component. The application is responsible for the accessibility inside the Drawer. 
- use "aria-live" for the Drawer, but in this case, we need to dive into details how to use it, how to exclude some text, 
how to use us earia-atomic, aria-relevant etc.

I'd prefer the first point.

eszthoff (approved):

PR #964, opened 2023-05-04
feat(Select components): add ARIA labels and props for toggle button; migrate Select tests

eszthoff:
Story: [JF-3902](https://textkernel.atlassian.net/browse/JF-3902)
[Design document](https://docs.google.com/document/d/1cEORjxfjXCA3gREzoEGzQowBG0Y35MjCLKpvSfiqkGg/edit#) for this PR


src/components/FieldWrapper/FieldWrapper.tsx
```diff
@@ -80,7 +86,7 @@ export const FieldWrapper = React.forwardRef<HTMLDivElement, Props>(
                         title={clearTooltipLabel}
                         {...elem('clearButton', { rightIndent: showArrow })}
                     >
-                        <MdClear size={16} />
+                        <MdClear size={16} role="img" aria-label={clearTooltipLabel} />
```
mukiienko, line 89: What is the profit of using role="img" for SVG inside button? The button tag itself can be read by the reader. We only need to add a label to the button.

src/components/FieldWrapper/FieldWrapper.tsx
```diff
@@ -36,11 +40,13 @@ export const FieldWrapper = React.forwardRef<HTMLDivElement, Props>(
...
             isArrowUp = false,
             onArrowClick,
-            clearTooltipLabel = '',
+            clearTooltipLabel = CLOSE_ICON_LABEL,
             showClearButton = false,
             onClear = null,
             isFocused = false,
             disabled = false,
+            openArrowLabel = DOWN_ARROW_LABEL,
+            closeArrowLabel = UP_ARROW_LABEL,
```
mukiienko, line 49: Can we see these constants in the Storybook as a default value?

src/components/FieldWrapper/FieldWrapper.tsx
eszthoff: I think it is good to add clear roles, it allows us to grab elements based on role during testing. I'm not 100% sure of all the benefits, but at least RTL recommends to use that selection method above all others. But indeed, if we do have the label, we can also select by that.

src/components/FieldWrapper/FieldWrapper.tsx
eszthoff: yes

src/components/FieldWrapper/FieldWrapper.tsx
mukiienko: I mean why do we need to add some role to the SVG icon? The button is a clickable element, not an icon, right?

eszthoff: the role `img` reflects that it is not clickable. It might not be 100% needed, but I also don't see any harm in adding it, and being more explicit.

mukiienko (approved):

andrentk (approved):

PR #974, opened 2023-05-17
Please Ignore It

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-387)


stories/atoms/Map.tsx
```diff
@@ -28,7 +28,7 @@ const pointMarker = {
 
 export default {
     title: 'Atoms/Map',
-    component: MapWithGoogleLoader,
+    // component: MapWithGoogleLoader, // TODO
```
PavelKirvelTextkernel, line 31: Can't resolve error here

src/components/Map/__tests__/MapWithGoogleLoader.spec.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { MapWithGoogleLoader } from '..';
+import '@testing-library/jest-dom';
+
+describe('<MapWithGoogleLoader/> that loads google api and renders a Map', () => {
+    it('should render with default props', () => {
+        // @ts-ignore
+        const view = render(<MapWithGoogleLoader apiKey="someKey" />);
```
PavelKirvelTextkernel, line 9: After migrate Map to TS, it asks to provide all props 

image

eszthoff: These are not props, but react component attributes. Looks to me, as the import was somehow wrong, and this file thinks that `MapWihtGoogleLoader` is not a component

src/components/Map/__tests__/Map.spec.tsx
```diff
@@ -23,84 +26,131 @@ describe('<Map/> that renders a Map with markers', () => {
...
         expect(removeDataMock).toHaveBeenCalledTimes(1);
     });
+
     it('should render with markers', () => {
-        const wrapper = mount(<Map markers={[pointMarker, regionMarker]} />);
-        expect(wrapper.find('Marker')).toHaveLength(2);
-        expect(wrapper.find('Circle')).toHaveLength(1);
+        const view = render(<Map markers={[pointMarker, regionMarker]} />);
+
+        expect(view.container).toMatchSnapshot();
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: Could you please switch to my branch and open MapWithGoogleLoader.spec.tsx just delete // @ts-ignore

src/components/Map/__tests__/MapWithGoogleLoader.spec.tsx
PavelKirvelTextkernel: Could you please switch to my branch and open MapWithGoogleLoader.spec.tsx just delete // @ts-ignore

src/components/Map/__tests__/Map.spec.tsx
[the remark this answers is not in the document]
PavelKirvelTextkernel: I've read some topics and what's the point to test something that doesn't show

src/components/Map/__tests__/MapWithGoogleLoader.spec.tsx
PavelKirvelTextkernel: I've added for map component
image

eszthoff: I see. I think there is a problem with how you defined the props in `src/components/Map/MapWithGoogleLoader/MapWithGoogleLoader.tsx`. You extended `GoogleMaps` here as well, but you rendering `Map`, so this is not consistent. I added some comments in the code to fix it.

src/components/Map/MapWithGoogleLoader/MapWithGoogleLoader.tsx
```diff
@@ -1,6 +1,5 @@
 import * as React from 'react';
-import PropTypes from 'prop-types';
-import { LoadScriptNext } from '@react-google-maps/api';
+import { GoogleMap, LoadScriptNext } from '@react-google-maps/api';
 import { LoadingSpinner } from '../../LoadingSpinner';
 import { Map } from '../Map/Map';
```
eszthoff, line 4:
```suggestion
import { Map, MapProps } from '../Map/Map';
```

src/components/Map/MapWithGoogleLoader/MapWithGoogleLoader.tsx
```diff
@@ -12,41 +11,38 @@ import { Map } from '../Map/Map';
...
-        >
-            <Map ref={ref} {...rest} />
-        </LoadScriptNext>
-    );
-});
-
-MapWithGoogleLoader.displayName = 'MapWithGoogleLoader';
-
-MapWithGoogleLoader.propTypes = {
+interface Props extends Omit<GoogleMap, 'onLoad' | 'mapContainerStyle' | 'options'> {
```
eszthoff, line 15:
```suggestion
interface Props extends MapProps {
```

PavelKirvelTextkernel: :+1:

src/components/Map/__tests__/Map.spec.tsx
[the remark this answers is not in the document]
eszthoff: Indeed, it doesn't make sense to test things that do not affect the UI. However that is not the case here.Here it doesn't show only because we are mocking a very complex library. It would still be good in this case, to see that props are passed correctly, as there is quite a bit of logic around that in the code. Can you please research if that is possible?

src/components/Map/__tests__/Map.spec.tsx
```diff
@@ -23,84 +26,131 @@ describe('<Map/> that renders a Map with markers', () => {
...
+        const view = render(<Map markers={[pointMarker, regionMarker]} />);
+        view.rerender(<Map defaultHighlight={NL_PATHS as GeoJSON.GeoJsonObject} markers={[]} />);
+
         expect(setZoomMock).toHaveBeenCalledTimes(1);
         expect(setCenterMock).toHaveBeenCalledTimes(1);
     });
+
     it('should add highlighted areas when geoJson objects are passed as markers', () => {
-        mount(<Map markers={[FR_FRIESLAND]} />);
+        render(<Map markers={[FR_FRIESLAND as FeatureCollection]} />);
```
eszthoff, line 118: why are you using a different type then what is in the Prop definition in the component?

src/components/Map/__tests__/Map.spec.tsx
```diff
@@ -1,5 +1,8 @@
 import * as React from 'react';
-import toJson from 'enzyme-to-json';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+// eslint-disable-next-line import/no-unresolved
```
eszthoff, line 4: is there no way to remedy this?

src/components/Map/__tests__/__snapshots__/Map.spec.tsx.snap
```diff
@@ -0,0 +1,17 @@
...
+
+exports[`<Map/> that renders a Map with markers should render with default props 1`] = `
+<div>
+  <div
+    style="height: 100%; width: 100%;"
+  />
+</div>
+`;
+
+exports[`<Map/> that renders a Map with markers should render with markers 1`] = `
```
eszthoff, line 11: as discussed elsewhere, there is not much point in these snapshots for this component. I'm ok to leave the first, just the sake of completeness, but this one I'm sure we can remove.

src/components/Map/Map/Map.tsx
```diff
@@ -31,11 +31,11 @@ type RegionArea = {
     features: GeoJSON.GeoJsonObject[];
 };
 
-interface Props extends Omit<GoogleMap, 'onLoad' | 'mapContainerStyle' | 'options'> {
+interface Props {
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: When i add them we need to provide all of them. I can add them 
image

src/components/Map/Map/Map.tsx
```diff
@@ -31,11 +31,11 @@ type RegionArea = {
...
     /** The default parameters to determine the viewport when no circular or point markers are present. */
-    defaultArea:
+    defaultArea?:
         | {
-              address: string;
+              address: string | string[] | undefined;
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: :+1:

src/components/Map/MapWithGoogleLoader/MapWithGoogleLoader.tsx
```diff
@@ -1,8 +1,7 @@
 import * as React from 'react';
-import PropTypes from 'prop-types';
-import { LoadScriptNext } from '@react-google-maps/api';
+import { GoogleMap, LoadScriptNext } from '@react-google-maps/api';
 import { LoadingSpinner } from '../../LoadingSpinner';
-import { Map } from '../Map/Map';
+import { Map, MapProps } from '../Map/Map';
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: :+1:

src/components/Map/Map/Map.tsx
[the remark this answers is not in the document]
PavelKirvelTextkernel: Because in LocationSelectorDialog.tsx  use     
```
function getDefaultArea() {
        if (initialMapAddress || country) {
            return { address: initialMapAddress || country };
        }

        return undefined;
    }
```
image

src/components/Map/__tests__/Map.spec.tsx
PavelKirvelTextkernel: I don't know. It was here, i'll check it

src/components/Map/__tests__/__snapshots__/Map.spec.tsx.snap
PavelKirvelTextkernel: :+1:

src/components/Map/MapWithGoogleLoader/MapWithGoogleLoader.tsx
```diff
@@ -12,41 +11,38 @@ import { Map } from '../Map/Map';
...
+    language?: string;
     /** Regional setting for the map. By default Google uses US.
      * For details see: https://developers.google.com/maps/documentation/javascript/localization#Region
      */
-    region: PropTypes.string,
+    region?: string;
     /** other props to pass to the google loader. For details see: https://react-google-maps-api-docs.netlify.com/#loadscriptnext */
-    additionalGoogleProps: PropTypes.object, // eslint-disable-line react/forbid-prop-types
-};
+    additionalGoogleProps?: object;
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: I try GoogleMapProps but it doesn't allow to use in LoadScriptNext

src/components/Map/__tests__/Map.spec.tsx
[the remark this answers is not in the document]
PavelKirvelTextkernel: I've not found  solution. 
Only about enzyme:
https://stackoverflow.com/questions/54960750/test-that-react-prop-method-has-been-called-with-jest

src/components/Map/__tests__/Map.spec.tsx
PavelKirvelTextkernel: I can off this rule. But as i remember it's not good solution

eszthoff: Sorry, I don't get your answer. Why do we need this lint exception here?

src/components/Map/__tests__/MapWithGoogleLoader.spec.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { render, screen } from '@testing-library/react';
+import { MapWithGoogleLoader } from '..';
+import '@testing-library/jest-dom';
+
+describe('<MapWithGoogleLoader/> that loads google api and renders a Map', () => {
+    it('should render with default props', () => {
+        const view = render(<MapWithGoogleLoader apiKey="someKey" />);
+
+        expect(view.container).toMatchSnapshot();
+        expect(screen.getByRole('status')).toBeInTheDocument();
```
eszthoff, line 11: The snapshot shows a loader. Maybe that is expected, then we should adjust the title of this test. 
If we can render something more specific to the map, that would be great to test. However, since we don't actually want to call google from unit test, but stab the object, I'm not sure what tests we can do here at all. Any ideas?

eszthoff (changes requested):

src/components/Map/Map/Map.tsx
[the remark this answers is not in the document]
mukiienko: Instead of `GoogleMap` we need to use `GoogleMapProps` for props. Fixed.

src/components/Map/__tests__/MapWithGoogleLoader.spec.tsx
mukiienko: In fact, I believe we don't require tests specifically for `LoadScriptNext` since it serves as a wrapper for third-party component and `testing-library.com` guideline principle says "you should avoid mocking out component". Thus, I'd like to remove that test if you don't mind.

mukiienko: Hmm, it seems that even the `react-google-maps/api` library lacks comprehensive tests. :) https://github.com/JustFly1984/react-google-maps-api/tree/develop/packages/react-google-maps-api/src

src/components/Map/__tests__/Map.spec.tsx
```diff
@@ -23,84 +26,129 @@ describe('<Map/> that renders a Map with markers', () => {
...
         },
-    };
+    } as CircularMarker;
 
     it('should render with default props', () => {
-        const wrapper = mount(<Map />);
-        expect(toJson(wrapper)).toMatchSnapshot();
+        const view = render(<Map />);
+
+        expect(view.container).toMatchSnapshot();
```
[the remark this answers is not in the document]
mukiienko: It's possible to test props by mocking `Map` component, but I agree with you, it's  against best practices, so let's avoid this. I'll remove the test if you don't mind.

src/components/Map/__tests__/Map.spec.tsx
mukiienko: Removed

package.json
```diff
@@ -150,9 +151,9 @@
         "react-dom": "16.14.0",
         "react-icons": "3.11.0",
         "react-router-dom": "6.9.0",
+        "react-test-renderer": "16.14.0",
```
eszthoff, line 154: not sure why this change happens in this PR, but since it is here: do we need this package at all?

src/components/Map/__tests__/Map.spec.tsx
```diff
@@ -23,84 +25,123 @@ describe('<Map/> that renders a Map with markers', () => {
...
+        );
         // will be called as many times as the length of the returned array from addGeoJsonMock
         expect(removeDataMock).toHaveBeenCalledTimes(1);
     });
+
     it('should render with markers', () => {
-        const wrapper = mount(<Map markers={[pointMarker, regionMarker]} />);
-        expect(wrapper.find('Marker')).toHaveLength(2);
-        expect(wrapper.find('Circle')).toHaveLength(1);
+        render(<Map markers={[pointMarker, regionMarker]} />);
```
eszthoff, line 75: can we test something here?

src/components/Map/__tests__/Map.spec.tsx
```diff
@@ -23,84 +25,123 @@ describe('<Map/> that renders a Map with markers', () => {
...
         expect(setZoomMock).toHaveBeenCalledTimes(2);
         expect(setCenterMock).toHaveBeenCalledTimes(2);
     });
+
     it('should add default highlight when geoJson markers are reset', () => {
-        const wrapper = mount(<Map defaultHighlight={NL_PATHS} markers={[FR_FRIESLAND]} />);
+        const view = render(
+            <Map
+                defaultHighlight={NL_PATHS as GeoJSON.GeoJsonObject}
+                markers={[FR_FRIESLAND as FeatureCollection]}
```
eszthoff, line 137: I'm confused about these types.. I'd think both these files are GeoJSON files, so why we cast them differently. Also a few lines above it reads `FR_FRIESLAND as RegionArea`. Can these be more aligned, or clear when we use which?

eszthoff (changes requested):

src/components/Map/__tests__/MapWithGoogleLoader.spec.tsx
eszthoff: yes, that makes sense to me too

src/components/Map/__tests__/Map.spec.tsx
mukiienko: Actually no. I'll remove the test.

src/components/Map/__tests__/Map.spec.tsx
mukiienko: I'm also confused about all these types. What I got from the [documentation](https://www.grapecity.com/wijmo/api/interfaces/wijmo_chart_map.geojson.geojsonobject.html):

- `GeoJsonObject` is the base type for GeoJson that includes other types `FeatureCollection `, `Point`, `MultiPoint` etc.
-  `FeatureCollection` is the list of [features](https://rdrr.io/cran/geoops/man/Feature.html). 

Probably for `markers` prop, we need only `FeatureCollection` type that can draw the circles by coordinates. 
We can create a follow-up task to investigate it later.
Also, I removed `geojson` library. We can use `GeoJSON.FeatureCollection` instead.

stories/atoms/Map.tsx
```diff
@@ -91,14 +91,19 @@ export default {
     },
 };
 
-export const _Map = ({ width, height, ...args }) => (
+type MapArguments = {
+    width: string;
+    height: string;
+} & MapProps;
+
+export const _Map = ({ width, height, ...mapProps }: MapArguments) => (
```
eszthoff, line 99: why to rename `args`? 
`args` is what we call it in all the stories and it easy to understand that they are coming from the `_Map.args` object below. I would keep it consistent

src/components/Map/__tests__/Map.spec.tsx
eszthoff: I think the test did have a purpose, as it checked, at least to some extent, that the props were correctly parsed. Now we cannot check React props anymore (based on other comments in this PR). I wonder if we should at least leave a small TODO note to point out that we missing tests for these parts of the code. 🤔

mukiienko: We already have the test for checking `markers` props with the title `should fit map to markers'. The current test tested how the markers are rendered. For this kind of tests we need to mock Map component.

stories/atoms/Map.tsx
mukiienko: I don't mind changing, but personally, I find using `...mapProps` and `...props` to be more explicit in conveying the purpose of these variables.

PR #986, opened 2023-05-29
One UI 370 migrate popup base tests

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-370)


src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -14,28 +16,32 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
-            expect(wrapper.find('Popover')).toHaveLength(1);
+            expect(view.container).toMatchSnapshot();
+            expect(screen.getByRole('group')).toBeInTheDocument();
         });
-        it('should render popup in portal when requested', () => {
-            wrapper = mount(
+
+        it('should render popup in portal when requested', async () => {
+            const user = userEvent.setup();
+            view = render(
```
andrentk, line 44: Do we need this render? Looks like it is the same component as the one that is being render first in the beforeEach. But for some reason if we need we should use rerender.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -14,28 +16,32 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
+
+        it('should render popup when requested', async () => {
+            const user = userEvent.setup();
             // trigger setPopupVisibility(true) through our dummy component
-            wrapper.find('button').simulate('click');
+            await user.click(screen.getByRole('button'));
 
-            expect(toJson(wrapper)).toMatchSnapshot();
-            expect(wrapper.find('Popover')).toHaveLength(1);
+            expect(view.container).toMatchSnapshot();
```
andrentk, line 38: the snapshot will not show anything popup related because it is a portal.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -44,119 +50,127 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
-            expect(wrapper.find('Portal')).toHaveLength(1);
+            expect(view.container).toMatchSnapshot();
+            expect(screen.getByRole('group')).toBeInTheDocument();
         });
-        it('should support no popup content from renderer', () => {
-            wrapper = mount(
+
+        it('should support no popup content from renderer', async () => {
+            const user = userEvent.setup();
+            view = render(
```
andrentk, line 61: rerender

src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -14,28 +16,32 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
         it('should render with minimal props correctly', () => {
-            expect(toJson(wrapper)).toMatchSnapshot();
-            expect(wrapper.find('Popover')).toHaveLength(0);
+            expect(view.container).toMatchSnapshot();
+            expect(screen.queryByRole('group')).not.toBeInTheDocument();
         });
-        it('should render popup when requested', () => {
+
+        it('should render popup when requested', async () => {
+            const user = userEvent.setup();
```
andrentk, line 34: you can do this in the beforeEach because it looks like it is being used at every test.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -44,119 +50,127 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
+            await togglePopup(user);
 
-            // @ts-ignore
-            clickDocument();
-            wrapper.update();
+            expect(screen.getByRole('group')).toBeInTheDocument();
 
-            expect(wrapper.find('Popover')).toHaveLength(0);
+            await user.keyboard(`[${ESCAPE_KEY}]`);
```
andrentk, line 127: I think it is more confusing to use the ESCAPE_KEY variable than simply use [Escape], what do you think?

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: For me using constant is better.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: I don't remember why but we should init this user in each test case. At least we do it for all tests

PavelKirvelTextkernel:
@andrentk Hello, I've addressed comments could you please check it

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: Also test cases contain this approach before

src/components/PopupBase/__tests__/PopupBase.spec.tsx
andrentk: Doing this in the beforeEach we also init this user in each test case, we just don't need to repeat the line of code in every test.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
andrentk: I'm fine with it.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
andrentk: I didn't mean to remove the snapshot. You can make it to appear in the snapshot if you use `view.baseElement` instead of the `view.container`. I think you added that to the TESTING_LIBRARY_MIGRATION.md file.

PavelKirvelTextkernel: I think we don't need it. The test case **should render popup when requested** and i just check it  expect(screen.getByRole('group')).toBeInTheDocument();

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -14,28 +16,32 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
+            view = render(
                 <PopupBase anchorRenderer={anchorRendererMock} popupRenderer={popupRendererMock} />
             );
+            user = userEvent.setup();
         });
 
         it('should render with minimal props correctly', () => {
-            expect(toJson(wrapper)).toMatchSnapshot();
-            expect(wrapper.find('Popover')).toHaveLength(0);
+            expect(view.container).toMatchSnapshot();
```
andrentk, line 31: I think this is not also producing the expected snapshot. Look at the .snap file.

src/components/PopupBase/__tests__/PopupBase.spec.tsx
```diff
@@ -44,119 +50,120 @@ describe('<PopupBase> that adds basic anchor/popup functionality to rendered com
...
                 <PopupBase anchorRenderer={anchorRendererMock} popupRenderer={() => null} />
             );
 
             // trigger setPopupVisibility(true) through our dummy component
-            wrapper.find('button').simulate('click');
+            await user.click(screen.getAllByRole('button')[1]);
 
-            expect(toJson(wrapper)).toMatchSnapshot();
-            expect(wrapper.find('Popover')).toHaveLength(0);
+            expect(view.container).toMatchSnapshot();
```
andrentk, line 67: same as above

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: I was original snapshot
image

PavelKirvelTextkernel: This new snapshot
image

PavelKirvelTextkernel: This one will be if use expect(view.baseElement).toMatchSnapshot()

src/components/PopupBase/__tests__/PopupBase.spec.tsx
PavelKirvelTextkernel: If i use baseElement it will render 
image

src/components/PopupBase/__tests__/PopupBase.spec.tsx
andrentk: Sorry, you're right, I think I mentioned the wrong test.

src/components/PopupBase/__tests__/__snapshots__/PopupBase.spec.tsx.snap
```diff
@@ -1,175 +1,67 @@
...
 
 exports[`<PopupBase> that adds basic anchor/popup functionality to rendered components rendering should render popup in portal when requested 1`] = `
-<PopupBase
-  anchorRenderer={[Function]}
-  popupRenderer={[Function]}
-  renderInPortal={true}
->
-  <Button
-    onClick={[Function]}
+<div>
```
andrentk, line 4: I meant this snapshot, probably mentioned in the wrong test.

PavelKirvelTextkernel: :+1:

andrentk (approved):

PR #987, opened 2023-05-30
fix(PageWidthRestrictor): improve semantic HTML | test: migrate WeightedResultBar, BlockWidthRestrictor and PageWidthRestrictor

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-400)


src/components/WidthRestrictor/BlockWidthRestrictor/__tests__/BlockWidthRestrictor.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+        expect(view.container).toBeEmptyDOMElement();
+    });
+
+    it('should render correctly', () => {
+        view = render(<BlockWidthRestrictor>Some children</BlockWidthRestrictor>);
+
+        expect(view.container).toMatchSnapshot();
+        expect(view.container).toHaveTextContent('Some children');
+        expect(view.container.children[0].tagName).toBe('DIV');
```
PavelKirvelTextkernel, line 20: I'm not sure about this approach. Could we use it ?

src/components/WidthRestrictor/BlockWidthRestrictor/__tests__/BlockWidthRestrictor.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    it('should render correctly when element type is defined', () => {
+        view = render(<BlockWidthRestrictor As="nav">Some children</BlockWidthRestrictor>);
+
+        expect(view.container.children).toHaveLength(1);
+        expect(view.container.children[0].tagName).toBe('NAV');
```
PavelKirvelTextkernel, line 27: s/a

src/components/WidthRestrictor/BlockWidthRestrictor/__tests__/BlockWidthRestrictor.spec.tsx
eszthoff: First, I would write it as `expect(screen.findByRole('...').toBeInTheDocument`. But we should always ask ourselves: is this test useful? what concerns does it answer? etc.

Here you suggest to check if there is a `div` somewhere in the DOM. This doesn't tell us anything about what the user experiences/sees. So I don't see value in it. The test you wrote before, that checks if the text is present is a lot more valuable. I think that is enough for this check. this is also what you would do if you do a manual test: you would look for the text to be somewhere on the page. So I think we can indeed drop this line.

src/components/WidthRestrictor/BlockWidthRestrictor/__tests__/BlockWidthRestrictor.spec.tsx
eszthoff: Since the intent here is to verify that the `As` prop indeed effects the HTML tag that is rendered, it is a good test I think. If you can look up what the implicit role for `nav` is and use `getByRole` instead, that would be better.

I'm not sure of the usefulness of the previous line though...

eszthoff (changes requested):

src/components/WidthRestrictor/PageWidthRestrictor/PageWidthRestrictor.tsx
```diff
@@ -15,7 +15,7 @@ export const PageWidthRestrictor: React.FC<Props> = ({ children, ...rest }) => {
     }
 
     return (
-        <div {...rest} {...block({ ...rest })}>
+        <div {...rest} {...block({ ...rest })} role="group">
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: When i use this: 
image
Snapshot doesn't show role for **section** doesn't contain role 
image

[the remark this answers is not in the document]
PavelKirvelTextkernel: But when i set role directly it shows  The element section has an implicit role of region. Defining this explicitly is redundant and should be avoided.(jsx-a11y/no-redundant-roles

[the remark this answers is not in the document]
PavelKirvelTextkernel: :+1:

src/components/WidthRestrictor/BlockWidthRestrictor/__tests__/BlockWidthRestrictor.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/WidthRestrictor/BlockWidthRestrictor/__tests__/BlockWidthRestrictor.spec.tsx
PavelKirvelTextkernel: :+1:

eszthoff:
LGTM. I changed the PR title to reflect changes in customer facing component. Please use this for the merge commit, so that the changes are reflected in the changelog.

eszthoff (approved):

PR #1008, opened 2023-07-06
refactor: address TODO comment

PavelKirvelTextkernel:
Link [task](https://textkernel.atlassian.net/browse/ONEUI-411)


src/components/Themeroller/ThemeTuner/__tests__/ThemeTuner.spec.tsx
```diff
@@ -88,21 +87,60 @@ describe('ThemeTuner component', () => {
...
+        //         target: {
+        //             value: '#00000',
+        //         },
+        //     });
+        //
+        const ch = screen.getAllByRole('log')[1];
+        await ch.focus();
+        // await ch.click();
+        await user.click(ch);
+        await user.type(ch, '#ffffff');
```
PavelKirvelTextkernel, line 140: After get input field just hard-code role for it. But input value doesn't change

src/components/Themeroller/ThemeTuner/__tests__/ThemeTuner.spec.tsx
```diff
@@ -88,21 +89,16 @@ describe('ThemeTuner component', () => {
...
 
-        const firstInput = screen.getByRole('spinbutton');
-
-        await user.type(firstInput, '#00000');
+        await user.type(screen.getByRole('textbox'), '#ffffff');
 
-        expect(onChangeMock).toHaveBeenCalledWith({
-            '--color-background': '#00000',
-        });
+        expect(onChangeMock).toHaveBeenCalled();
```
mukiienko, line 102: Can we use `toHaveBeenCalledWith`?

PavelKirvelTextkernel: It shows 
image

mukiienko: hmm, then something is wrong with the test itself.

PavelKirvelTextkernel: I think the test is correct. The main problem is that enzyme allow to set the hole value to input. But when we use RTL type function it types by element here. The same problem we have for SelectBase.spec.tsx it was 
image

And then we replace it to 
image

PavelKirvelTextkernel: The solution is to use fireEvent.change I'll apply it for both tests

mukiienko (approved):

PR #1010, opened 2023-07-07
One UI 404 improve pill tests

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-404)

### DIFF
--- a/src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
+++ b/src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
@@ -41,6 +41,8 @@ describe('<PillButton> component', () => {
                 'PillButton__button PillButton__button--isOpen'
             );
             expect(getButtonByName(downArrowLabel)).not.toHaveClass('PillButton__button--isOpen');
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
         });
 
         it('should trigger toggle state once when clicked', async () => {
@@ -50,6 +52,7 @@ describe('<PillButton> component', () => {
             expect(toggleDropdownMock).toHaveBeenCalledTimes(1);
             expect(getButtonByName(downArrowLabel)).toBeInTheDocument();
             expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
         });
 
         it('should trigger toggle state once on keyboard interaction', async () => {
@@ -64,6 +67,8 @@ describe('<PillButton> component', () => {
         it('should have arrow down label', () => {
             expect(screen.getByRole(imgRole)).toBeInTheDocument();
             expect(getButtonByName(downArrowLabel)).toBeInTheDocument();
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
         });
     });
 
@@ -87,9 +92,8 @@ describe('<PillButton> component', () => {
             expect(getButtonByName(upArrowLabel)).toHaveClass(
                 'PillButton__button PillButton__button--isOpen'
             );
-            expect(getButtonByName('Pill name up arrow')).not.toHaveClass(
-                'PillButton__button--isOpen'
-            );
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
         });
 
         it('should have arrow up label', () => {
@@ -164,6 +168,8 @@ describe('<PillButton> component', () => {
             expect(getButtonByName('This pill is in use up arrow')).toHaveClass(
                 'PillButton__pill PillButton__pill--isOpen PillButton__pill--isActive'
             );
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
         });
 
         it('should trigger toggle state once when button is clicked', async () => {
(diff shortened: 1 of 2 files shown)


src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -86,13 +92,12 @@ describe('<PillButton> component', () => {
             expect(getButtonByName(upArrowLabel)).toHaveClass(
                 'PillButton__button PillButton__button--isOpen'
             );
-            expect(getButtonByName('Pill name up arrow')).not.toHaveClass(
-                'PillButton__button--isOpen'
-            );
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
```
eszthoff, line 95: I'm unsure about this check. It is for the open state, so we should expect the arrow up to show. I think there is a problem with how you checking it. There is no text on the screen with the label. It is only a label :-)
I think you need to use `queryByLabelText` or with role and name here and also in the other tests in this document.
Then the arrow up should be in the document, and we can add a check that downArrowLabel isn't.

Also someone recently pointed out and it might be better to use `toBeVisible` then `toBeInTheDocument`.

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -13,6 +13,7 @@ describe('<PillButton> component', () => {
     const downArrowLabel = 'down arrow';
     const upArrowLabel = 'up arrow';
     const clearLabel = 'clear label';
+    const imgRole = 'img';
 
     let view: RenderResult;
```
eszthoff, line 19: The line below (I cannot comment there) can be written simpler: `{ name: inputName }`

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -163,13 +168,17 @@ describe('<PillButton> component', () => {
             expect(getButtonByName('This pill is in use up arrow')).toHaveClass(
                 'PillButton__pill PillButton__pill--isOpen PillButton__pill--isActive'
             );
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
```
eszthoff, line 172: s/a

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -163,13 +168,17 @@ describe('<PillButton> component', () => {
...
         });
 
         it('should trigger toggle state once when button is clicked', async () => {
             const user = userEvent.setup();
             await user.click(getButtonByName(upArrowLabel));
 
             expect(toggleDropdownMock).toHaveBeenCalledTimes(1);
+            expect(screen.queryByText(upArrowLabel)).not.toBeInTheDocument();
+            expect(screen.queryByText(clearLabel)).not.toBeInTheDocument();
```
eszthoff, line 181: I don't think we need to repeat these tests here, if we had them in the previous `it`

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -86,13 +92,12 @@ describe('<PillButton> component', () => {
...
         });
 
         it('should have arrow up label', () => {
-            expect(screen.getByRole('img')).toBeInTheDocument();
+            expect(screen.getByRole(imgRole)).toBeInTheDocument();
             expect(getButtonByName(upArrowLabel)).toBeInTheDocument();
         });
     });
```
eszthoff, line 103: In the below test `should render correctly` can you add a check for which arrow is showing?

eszthoff (changes requested):

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
PavelKirvelTextkernel: :+1:

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
PavelKirvelTextkernel: We can't use toBeVisible for null.
Error message:
**received value must be an HTMLElement or an SVGElement.
Received has value: null**

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
PavelKirvelTextkernel: :+1:

eszthoff: this is not yet addressed; note that line is not visible in the diff in the Github UI.

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -190,8 +192,14 @@ describe('<PillButton> component', () => {
         });
 
         it('should have arrow up label', () => {
-            expect(screen.getByRole(imgRole)).toBeInTheDocument();
-            expect(getButtonByName(upArrowLabel)).toBeInTheDocument();
+            expect(screen.getByRole(imgRole, { name: upArrowLabel })).toBeVisible();
```
eszthoff, line 195: this and the next line is the very same check, only once you call it through the function and once inline, am I correct (in which case this line can be removed)?

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
```diff
@@ -190,8 +192,14 @@ describe('<PillButton> component', () => {
...
+            expect(getButtonByName(upArrowLabel)).toBeVisible();
+            expect(getButtonByName(upArrowLabel)).toHaveClass(
```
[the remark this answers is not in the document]
PavelKirvelTextkernel: :+1:

src/components/Pill/PillButton/__tests__/PillButton.spec.tsx
PavelKirvelTextkernel: We have img and button with 'arrow up' label. I think we need to check both

eszthoff: I missed that. Thanks.

eszthoff (approved):

PR #1023, opened 2023-07-19
build: packages upgrade react and dependent components

PavelKirvelTextkernel:
Task [link](https://textkernel.atlassian.net/browse/ONEUI-289)


.eslintrc.json
```diff
@@ -58,7 +58,10 @@
         "@typescript-eslint/comma-dangle": ["error", "only-multiline"],
         "@typescript-eslint/explicit-module-boundary-types": "off",
         "@typescript-eslint/ban-ts-comment": "off",
-        "@typescript-eslint/ban-types": "off"
+        "@typescript-eslint/ban-types": "off",
+        "react/require-default-props": "off",
+        "react/no-unused-prop-types": "off",
```
antipin, line 63: eslint is only unhappy about`no-unused-prop-types` in `components/Navigation/NavItem/NavItem.tsx` component.

It's better to disable it just for that component (until it's fixed).

You can make it my wrapping the `type Props = {}` block

```ts
/* eslint-disable react/no-unused-prop-types */
export interface Props extends React.HTMLAttributes<HTMLDivElement> {
    ...
}
/* eslint-enable react/no-unused-prop-types */

```

package.json
```diff
@@ -65,12 +65,12 @@
...
-        "randomcolor": "~0.6.0",
-        "rc-slider": "10.1.1",
-        "react-datepicker": "4.5.0",
-        "react-modal": "~3.11.2",
-        "react-popper": "2.2.5",
-        "react-virtualized": "~9.22.0",
+        "randomcolor": "~0.6.2",
+        "rc-slider": "10.2.1",
+        "react-datepicker": "4.16.0",
+        "react-modal": "~3.16.1",
```
antipin, line 71: Let's use exact versions everywhere. Please remove `~` prefix for all deps in `package.json`

.eslintrc.json
```diff
@@ -58,7 +58,10 @@
...
+        "react/jsx-no-bind": "off"
```
antipin, line 64: `react/jsx-no-bind` is a useful rule, it can prevent some nasty hard to debug errors with losing `this`.

Since `eslint` is only unhappy about it in 2 places, it's better to fix it.

* `LocationSelector/LocationSelector.tsx`
* `LocationSelector/LocationSelectorDialog/LocationSelectorDialog.tsx`

```ts
// Rewrite handler functions from functions to arrow functions.
// Example:
// src/components/LocationSelector/LocationSelector.tsx

// Before
function handleCloseModal() {
    if (isOpen && isBrowserTabVisible) {
        setIsOpen(false);
        onBlur?.();
    }
}

// After
const handleCloseModal = () => {
    if (isOpen && isBrowserTabVisible) {
        setIsOpen(false);
        onBlur?.();
    }
};
```

antipin (changes requested):
Looks good 👍 , just a few notes

package.json
PavelKirvelTextkernel: :+1:

.eslintrc.json
PavelKirvelTextkernel: :+1:

.eslintrc.json
PavelKirvelTextkernel: :+1:

PR #1059, opened 2023-09-22
refactor: migrate MarkedText to TS

andrentk:
Story: [ONEUI-412](https://textkernel.atlassian.net/browse/ONEUI-412)


src/components/Text/MarkedText/MarkedText.tsx
```diff
@@ -1,15 +1,23 @@
...
-import PropTypes from 'prop-types';
+import { Context } from '@textkernel/oneui/constants';
 import { bem } from '../../../utils';
 import { Text } from '../Text';
 import styles from './MarkedText.scss';
 
 const { block, elem } = bem('MarkedText', styles);
 
-export const MarkedText = (props) => {
+interface Props {
```
eszthoff, line 9: looking at it more, I think some of the props are actually TextProps. Maybe it makes better sense to use `extend TextProps` here, and not list `inline` and `context`

src/components/Text/MarkedText/MarkedText.tsx
```diff
@@ -1,15 +1,23 @@
...
+    children: string;
+    marker: string;
+    inline?: boolean;
+    context?: Context | 'primary';
```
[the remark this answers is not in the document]
andrentk: Done

src/components/Text/MarkedText/MarkedText.tsx
andrentk: You're right!! Fixed.

PR #1143, opened 2024-02-22
feat(Card): new component

andrentk:
Story: [ONEUI-435](https://textkernel.atlassian.net/browse/ONEUI-435)

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import { Text } from '../Text';
+import { Heading } from '../Heading';
+import styles from './Card.scss';
+
+export interface Props {
+    title: string;
+    subtitle?: string;
+    trailing?: string | number;
+}
+
+export const Card = ({ title, subtitle, trailing }: Props) => (
+    <div className={styles.container}>
+        <div className={styles.leftColumn}>
+            <Heading level="h3" className={styles.title}>
+                {title}
+            </Heading>
+            {subtitle && (
+                <Text inline className={styles.subtitle}>
+                    {subtitle}
+                </Text>
+            )}
+        </div>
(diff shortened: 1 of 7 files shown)


src/components/Card/__tests__/Card.spec.tsx
janseke10, line 1: Maybe you could add a test + storybook option with only title and subtitle, no trailing

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Text } from '../Text';
+import { Heading } from '../Heading';
+import styles from './Card.scss';
+
+export interface Props {
```
eszthoff, line 6: This should extend usual Div attributes, which should then be passed to the top `div` via `...rest`. This will make sure that e.g. `style` or aria attributes can be passed by the 3rd party user.

src/index.ts
```diff
@@ -15,6 +15,7 @@ export * from './customTypes/types';
 export * from './components/Buttons';
 export * from './components/Callout';
 export * from './components/CandidateAvatar';
+export * from './components/Card';
```
eszthoff, line 18: this component uses Heading and Text, so should be Molecule, not an Atom.

stories/atoms/Card.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import type { Meta, StoryObj } from '@storybook/react';
+import { Card } from '@textkernel/oneui';
+
+const meta: Meta<typeof Card> = {
+    title: 'Atoms/Card',
```
eszthoff, line 6: This should be a Molecule, not an Atom (see comment above)

eszthoff (changes requested):

PR #1192, opened 2024-06-07
feat(ExpandableText): new component

eszthoff:
Story:[ONEUI-442](https://textkernel.atlassian.net/browse/ONEUI-442)

# Checklist

### DIFF
--- a/src/components/ExpandableText/ExpandableText.tsx
+++ b/src/components/ExpandableText/ExpandableText.tsx
@@ -1,0 +1,83 @@
+import * as React from 'react';
+import { MdKeyboardArrowUp } from 'react-icons/md';
+import { bem } from '../../utils';
+import styles from './ExpandableText.scss';
+import { Button } from '../Buttons';
+
+export interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** The content of the block */
+    children: React.ReactNode;
+    /** The maximum number of lines to be shown when displaying the text in collapsed mode  */
+    threshold?: number;
+    /** Label for the button when the element is collapsed */
+    showMoreButtonLabel: string;
+    /** Label for the button when the element is expanded */
+    showLessButtonLabel: string;
+}
+
+const { block, elem } = bem('ExpandableText', styles);
+
+export const ExpandableText = React.forwardRef<HTMLDivElement, Props>((props, ref) => {
+    const { children, threshold = 5, showMoreButtonLabel, showLessButtonLabel, ...rest } = props;
+    const contentRef = React.createRef<HTMLDivElement>();
+    const [isExpandable, setIsExpandable] = React.useState(false);
+    const [isCollapsed, setIsCollapsed] = React.useState(true);
+
+    const collapsedHeight = `calc(var(--line-height-normal) * ${threshold})`;
+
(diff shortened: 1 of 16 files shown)


antipin (approved):

src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,82 @@
...
+        if (isCollapsed) {
+            expandSection();
+        } else {
+            collapseSection();
+        }
+        setIsCollapsed((prevIsCollapsed) => !prevIsCollapsed);
+    };
+
+    return (
+        <div ref={ref} {...rest} {...block(props)}>
```
mukiienko, line 68: Tell me please why do we pass props to `...block(props)`.?

src/components/ExpandableText/ExpandableText.tsx
```diff
@@ -0,0 +1,82 @@
...
+        contentRef.current.style.height = `${sectionHeight}px`;
+    };
+
+    const handleToggleClick = () => {
+        if (isCollapsed) {
+            expandSection();
+        } else {
+            collapseSection();
+        }
+        setIsCollapsed((prevIsCollapsed) => !prevIsCollapsed);
```
mukiienko, line 64: I'd move run function `setIsCollapsed` to the methods `expandSection` and `collapseSection`, otherwise running those function without `setIsCollapsed` will be break UI.

mukiienko (approved):

src/components/ExpandableText/ExpandableText.tsx
eszthoff: Sorry, I'm not sure I understand what do you mean. Can you elaborate?

src/components/ExpandableText/ExpandableText.tsx
eszthoff: This is the usual way we always pass them. E.g. if `classname` is passed as a prop, that will be used by `block`. Or am I missing something?

mukiienko: You are right! I forgot about the possibility of passing `classname ` :)

src/components/ExpandableText/ExpandableText.tsx
mukiienko: Imagine I need to run functions `expandSection` and `collapseSection` also from another place. In this case, the height will be changed, but state `isCollapsed` will be the same. So the idea to add `setIsCollapsed` inside those methods:


```
    const expandSection = () => {
        if (!contentRef.current) {
            return;
        }

        const sectionHeight = contentRef.current.scrollHeight;
        contentRef.current.style.height = `${sectionHeight}px`;
        setIsCollapsed(true);
    };
```

eszthoff: but it is not possible to call these functions from somewhere else. 🤔 They aren't exposed, and that was not a requirement either. How would you call these from outside?

mukiienko: I understand they aren't exposed, but the concern is about maintaining consistency within the component. By integrating setIsCollapsed within these methods, we ensure that the component's state remains consistent regardless of how these methods are invoked.

mukiienko (approved):

PR #1213, opened 2024-07-17
feat(Checkbox): migrate to new design

eszthoff:
Story: [ONEUI-444](https://textkernel.atlassian.net/browse/ONEUI-444)

### DIFF
--- a/src/components/Checkbox/Checkbox.scss
+++ b/src/components/Checkbox/Checkbox.scss
@@ -73,6 +73,8 @@
         margin: 0;
         width: 0;
         height: 0;
+        border: none;
+        appearance: none;
 
         &:focus {
             + .Checkbox__label {
--- a/src/components/Checkbox/Checkbox.scss
+++ b/src/components/Checkbox/Checkbox.scss
@@ -134,6 +134,10 @@
                 }
             }
         }
+        /* stylelint-enable no-duplicate-selectors, no-descending-specificity */
+
+        .Checkbox__text {
+            color: var(--color-text-disabled);
+        }
     }
-    /* stylelint-enable no-duplicate-selectors, no-descending-specificity */
 }
--- a/src/components/Checkbox/Checkbox.scss
+++ b/src/components/Checkbox/Checkbox.scss
@@ -23,10 +23,7 @@
     }
 
     &__label {
-        margin: auto;
-        user-select: none;
         cursor: pointer;
-        display: block;
 
         &--asFlexbox {
             display: flex;
@@ -73,6 +70,7 @@
     }
 
     &__input {
+        margin: 0;
         width: 0;
         height: 0;
 
--- a/src/components/Checkbox/Checkbox.scss
+++ b/src/components/Checkbox/Checkbox.scss
@@ -76,7 +76,7 @@
         border: none;
         appearance: none;
 
-        &:focus {
+        &:focus-visible {
             + .Checkbox__label {
                 .Checkbox__box {
                     box-shadow: 0px 0px 0px 2px var(--color-info-20, #99CAFF);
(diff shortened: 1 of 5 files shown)


src/components/Checkbox/Checkbox.scss
```diff
@@ -47,76 +32,108 @@
...
+                    box-shadow: none;
+                }
+            }
+        }
+    }
+
     &__input {
-        display: none;
+        width: 0;
+        height: 0;
```
AnastasiiaMP, line 77: it looks like specifying width and height instead of using `display: none` led to the extra space 🤔 the best you could see it for ThreeCheckboxState. Is it intended?

src/components/Checkbox/Checkbox.scss
```diff
@@ -47,76 +32,108 @@
...
 
-        &--indeterminate {
```
AnastasiiaMP, line 66: I also couldn't get mixed checkbox while testing... isn't indeterminate was responsible for it or am I missing something?

src/components/Checkbox/Checkbox.scss
eszthoff: nice catch!
note: I changed the display option so that keyboard navigation is possible, and you can use `TAB` to reach the checkbox.

src/components/Checkbox/Checkbox.scss
eszthoff: fixed the story

src/components/Checkbox/Checkbox.scss
AnastasiiaMP: Thank you for the explanation and making changes. 
I tested it in several browsers and it looks good only in Chrome to me. 
Firefox:
Screenshot 2024-07-22 at 14 56 22

Safari:
Screenshot 2024-07-22 at 14 56 58

Could you please have a look?

Not sure what is happening in Microsoft Edge, though 🫣

AnastasiiaMP (changes requested):

src/components/Checkbox/Checkbox.scss
AnastasiiaMP: works perfectly now :-) Thank you

src/components/Checkbox/Checkbox.scss
eszthoff: Nice catch again! 
And here I was, thinking those differences are a problem of the past. I learned [something](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) new today.  From the look of this attribute, I think it should be ok in Edge too
I managed to use Browserstack to test Edge too and it looks good. but when I tried to record it, it all goes confused (I think it is browserstack). I can PM you the credentials if you want.

AnastasiiaMP:
LGTM 🎉 reviewed and QAed

AnastasiiaMP (approved):

src/components/Checkbox/Checkbox.scss
```diff
@@ -1,122 +1,143 @@
...
     &__svg {
         position: absolute;
-        top: 0;
+        top: var(--border-width-thick);
         left: 0;
         fill: none;
-        stroke: var(--color-background);
+        fill-rule:evenodd;
+        clip-rule:evenodd;
+        stroke: var(--color-background-information-subtlest-default);
```
AnastasiiaMP, line 53: shouldn't it be white as mentioned in Dima's comment? 👀

AnastasiiaMP (changes requested):

src/components/Checkbox/Checkbox.scss
eszthoff: When I talked to Dima he said it actually does look ok but I forgot to check after that. Nice, thanks.

AnastasiiaMP (approved):

PR #1214, opened 2024-07-18
feat(Toast): New Component

janseke10:
[ONEUI-451](https://textkernel.atlassian.net/browse/ONEUI-451)

feat(Toast): new component

# Checklist


src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** Short and descriptive title */
+    title: string;
+    /** Body content */
+    content?: string;
+    /** The Toast context (e.g. info, critical, success or cautious - defaults to info) */
+    context?: Context;
+    /** If there are actions (links or buttons) in the toast */
+    actionable?: boolean;
```
eszthoff, line 14: Instead of manually setting it, could we drop this prop, and deduct its value from the presents/absence of actions?

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,25 @@
...
+    /** Action object, containing item text and href to action */
+    actionOne?: { text: string; href: string };
+    /** Action object, containing item text and href to action */
+    actionTwo?: { text: string; href: string };
```
eszthoff, line 18: Maybe instead of going with 1 and 2, we could have it as one prop `actions?: {text: string; callback: () => void; href: string}[]`. Also note that I added `callback` here. I think an action can be a function not only a navigation item. We can use some `if` statement maybe to see if we need to render a link (`<a>`) or a button component

stories/molecules/Toast.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+export default meta;
+
+type Story = StoryObj<typeof Toast>;
+
+export const NotActionableToast: Story = {
+    name: 'Normal Toast',
+    args: {
+        title: 'Success',
+        content: 'This is a toast',
+        context: 'success',
```
eszthoff, line 19: If we can use `CONTEXT[1]` it would be better, so that if it changes in the future we don't have to change it manually here

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import * as React from 'react';
+import { bem } from '../../utils';
+import { Context } from '../../constants';
+
+export interface Props extends React.HTMLAttributes<HTMLDivElement> {
+    /** Short and descriptive title */
+    title: string;
```
eszthoff, line 8: Seems it can be optional (see design)

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,25 @@
...
+    /** Body content */
+    content?: string;
```
eszthoff, line 10: This seems to be mandatory (see designs)

src/components/Toast/Toast.tsx
janseke10: Ah, I thought it was more like 'if it's actionable, **then** it could include links and buttons. And that a toast can also be actionable without links and buttons, causing it to be focused on, in case it's really important. 
But maybe I understood it wrong. 
https://zeroheight.com/9e76adf4e/p/08cfd0-toast , see 'Action (actionable only)'

src/components/Toast/Toast.tsx
janseke10: Yes, very good idea. Should we put a limit on the amount of actions, though?

src/components/Toast/Toast.tsx
eszthoff: I think if it is actionable then there will be actions. But maybe best if you check with Dima.

src/components/Toast/Toast.tsx
eszthoff: Please check with Dima, but my 2 cents is there that would be over engineering. Let's hope people are resonable enough not to offer 30 actions to a user in the context of a toast.
Also I don't know how would you limit that in the TS definition. In the code you could make sure to render only the first few, but that is really opinionated, and sounds like a source of bugs to me :-D

janseke10: There should indeed be a limit of 2 actions according to Dima. We could still keep the actionOne & actionTwo, just adding the callback. But a single prop does look cleaner indeed

eszthoff: I'd go with the array, and maybe say in the prop description that more then 2 actions is not recommended. But I wouldn't enforce it

andrentk (commented):
You can already remove the Draft status

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,160 @@
+/* eslint-disable import/no-unresolved */
+import * as React from 'react';
+import { Toaster, toast } from 'sonner';
+import { MdClose } from 'react-icons/md';
```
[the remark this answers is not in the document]
janseke10: I implemented this. However, this material design icon has a size of 24px, and I cannot change this (setting font-size to 20px doesn't work, it just messes with the alignment). So the question is: is it a problem if this is 24px instead?

src/components/Toast/Toast.scss
```diff
@@ -0,0 +1,108 @@
...
+            font: {
+                size: 20px;
+                weight: var(--font-weight-bold);
+                family: var(--font-family-primary);
+            }
+            line-height: 24px;
+            margin: 0;  
+        }
+
+        p {
```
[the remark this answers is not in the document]
janseke10: Without applying them I didn't get the right values. What do you mean exactly with 'use and a instead of the h3 and p tags styled'?

mukiienko:
I see that element has `--z-index: 2;`. That's not aligned with our components. For example, we have Drawer component that has    `--drawer-z-index: 800;`, but I believe Toast should overlap the Drawer:
Screenshot 2024-08-15 at 11 46 23

mukiienko:
I like it. Nice animation!

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,157 @@
...
+        case 'info':
+            return <InfoIcon data-testid="info-icon" />;
+        case 'success':
+            return <SuccessIcon data-testid="success-icon" />;
+        case 'cautious':
+            return <CautiousIcon data-testid="cautious-icon" />;
+        case 'critical':
+            return <CriticalIcon data-testid="critical-icon" />;
+        default:
+            return <InfoIcon />;
```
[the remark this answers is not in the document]
janseke10: I forgot about that! Yes :)

src/components/Toast/Toast.scss
```diff
@@ -0,0 +1,102 @@
...
+        flex-direction: column;
+        gap: var(--space-150);
+    }
+
+    &__message {
+        gap: var(--space-100);
+        display: flex; 
+        flex-direction: column;
+        margin: 0;
+        font-family: var(--font-family-primary);
```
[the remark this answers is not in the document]
janseke10: Somehow it gets the font from data-sonner-toaster, for the h3 it does. I've tried inherit here, but it doesn't work (it inherits it from the sonner toast)

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,157 @@
...
+    closeButtonLabel?: string;
+}
+
+interface ActionProps {
+    action: {
+        text: string;
+        callback?: () => void;
+        href?: string;
+    };
+    handleClose: () => void;
```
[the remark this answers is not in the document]
janseke10: We wouldn't necessarily want to close all toasts, which would happen if we did it on the toaster level. 
The thing is, in most situations the toast handles closing itself. Only when we add our own buttons and we want to close to toast onClick we can use the toast that is being passed as we use sonnerToast.custom
We can use this toast (t) to then call sonnerToast.dismiss(t)
It than removes that specific toast. As this is toast specific, and something that is only 'created' when calling toast() we can't really pass it as a prop, as 't' is not available yet. 
This is why I added it here. I could also pass the toast / t in question and call dismiss in the ActionItem, but I don't know if that's cleaner here.

mukiienko (approved):

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,157 @@
...
+                        {actions && actions[1] && (
+                            <>
+                                {' '}
+                                •{' '}
+                                <ActionItem
+                                    action={actions[1]}
+                                    handleClose={() => sonnerToast.dismiss(t)}
+                                />
+                            </>
+                        )}
```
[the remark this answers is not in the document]
eszthoff: nice workaround :-)

src/components/Toast/Toast.tsx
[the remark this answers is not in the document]
eszthoff: sorry, I still don't get why we don't have a function like:
```
const handleClose = (t: ToastInstance) => {
  onClose?.()
  sonner.dismiss(t)
}
```
Then we call this function when an action clicked, but also when the X button clicked, or (if possible) when the thing times out. What is the logic behind calling in only when an action was clicked, and not when X was clicked?

src/components/Toast/Toast.tsx
```diff
@@ -139,17 +136,21 @@ export const toast = ({
...
 
-export const Toaster = ({ ...props }) => <SonnerToaster {...props} className="Toaster" />;
-
-Toaster.defaultProps = {
-    // duration: 2500,
-    closeButton: true,
-    offset: 16,
-    gap: 8,
-};
+export const Toaster = ({ duration = 2500, ...props }) => (
```
eszthoff, line 142: Sorry, only noticed it now. We never really put 2 components in 1 file. Can you move it to a separate one?

janseke10: but in this case it's not a 'component', toast() is a function... should i still seperate them?

src/components/Toast/Toast.tsx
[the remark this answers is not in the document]
eszthoff: Clarified f2f. This is an internal prop, so no need to be generic. If in the future we have a usecase to report to the application when a toast closes, we will add new prop for that.

src/components/Toast/Toast.tsx
eszthoff: Clarified f2f. 
`ActionItem` will be separated to its own file/folder.
Main folder and file will be renamed to `Toaster`

src/components/Toaster/Toaster.tsx
```diff
@@ -139,20 +91,21 @@ export const toast = ({
         actions ? { important: true } : {} // focused if actionable
     );
 
+const bemToaster = bem('Toaster', styles);
+
 export const Toaster = ({ duration = 2500, ...props }) => (
     <SonnerToaster
+        {...bemToaster.block()}
```
eszthoff, line 98: This can lead to a bug. Basically, this thing will set the prop `className`. The way the package works is that it takes the props passed and calculates what the classes should be. If the props include a `classname` already, it will add that too, to allow users to add additional styles if needed. So with the current code we can have 2 bugs:
1. Props aren't passed to `block()` so if classname is defined it will not be included in the applied styles
2. `props` are spread after `block` is spread. if `classname` is defined in the `props` it will overwrite the classname added by `block`
So we need it to be last in the spread, and called with `props` as an argument

src/components/Toaster/ActionItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+            if (e.key === ENTER_KEY) {
+                handleOnClick();
+            }
+        };
+        return (
+            <p
+                onClick={handleOnClick}
+                onKeyDown={handleKeyDown}
+                {...elem('toastAction')}
+                role="presentation"
```
eszthoff, line 32: why did you decide to use `<p role="presentation" />`? 
I'd think it is a button, so should be `button`. Not sure we need to set the role manually, but [presentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role) normally mean it is not interactive at all. 
If using `<button />` is complicated in terms of styling it, we can use `<div role="button" />`

eszthoff (changes requested):

eszthoff (approved):

src/components/Toaster/Toaster.tsx
```diff
@@ -0,0 +1,111 @@
...
+            duration,
+            closeButton: true,
+            offset: 16,
+            gap: 8,
+            ...bemToaster.block(props),
+        }}
+    />
+);
+
+toast.displayName = 'toast';
```
mukiienko, line 109: Is it react component? I wondering why it starts from lowercase?

mukiienko (approved):

src/components/Toaster/Toaster.tsx
janseke10: It's actually a function! But it is a function we need to call from anywhere, hence the displayName. We use the same structure as the Sonner library: Toaster component, and a toast function to make toast appear

mukiienko: hmm, interesting :) so it's kind of imperative way showing notifications.

PR #1231, opened 2024-08-08
feat(Toggle): migrate to new design

eszthoff:
Story: [ONEUI-449](https://textkernel.atlassian.net/browse/ONEUI-449)

### DIFF
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -1,5 +1,5 @@
 .Toggle {
-    display: flex;
+    display: inline-flex;
     align-items: center;
     cursor: pointer;
 
--- a/src/components/Toggle/Toggle.scss
+++ b/src/components/Toggle/Toggle.scss
@@ -6,9 +6,9 @@
     &__toggleContainer {
         width: 36px;
         height: 20px;
-        border-radius: 20px;
+        border-radius: 24px;
         position: relative;
-        background-color: var(--color-neutral-30);
+        background-color: var(--color-background-neutral-subtlest-default);
         transition: border-color 300ms;
         flex-shrink: 0;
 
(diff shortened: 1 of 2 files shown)


mukiienko:
@eszthoff, I think when element is focused, pushing enter button should toggle component.

mukiienko:
@eszthoff Clicking on empty space switch the toggle. It looks strange:

https://github.com/user-attachments/assets/39b920f7-edb2-405e-8232-eb0604c94f51

eszthoff:
> @eszthoff, I think when element is focused, pushing enter button should toggle component.

Do you have some accessibility guidelines that support that? 
I believe now it is work as a native checkbox would, and you need to press the space bar to move the toggle.

mukiienko:
> > @eszthoff, I think when element is focused, pushing enter button should toggle component.
> 
> Do you have some accessibility guidelines that support that? I believe now it is work as a native checkbox would, and you need to press the space bar to move the toggle.

I thought the Enter key should also toggle the switch, but I was mistaken. The Space key works well.

eszthoff:
> @eszthoff Clicking on empty space switch the toggle. It looks strange:

Not sure how you would like to address it. It is a "problem" with the story. Basically you are clicking on the empty label. In a real application the `div` would be smaller and hence it would not happen so much. Please advise

mukiienko:
> > @eszthoff Clicking on empty space switch the toggle. It looks strange:
> 
> Not sure how you would like to address it. It is a "problem" with the story. Basically you are clicking on the empty label. In a real application the `div` would be smaller and hence it would not happen so much. Please advise

in the application level the div can be smaller, or not :) try to use `display: inline-flex;` for `Toggle` class.

mukiienko (approved):

PR #1234, opened 2024-08-12
feat(Button): button redesign

janseke10:

BREAKING CHANGE: button props isLink & isInline & isPrimary are removed

- inline button should be replaced by size="small"
- isLink should be replaced by variant="ghost"
- isPrimary should be replaced with variant="filled" (default)

# Checklist


src/components/Buttons/Button/Button.tsx
```diff
@@ -39,17 +39,25 @@ export const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, Props>(
...
             isBlock = false,
-            isInline = false,
             isLoading = false,
-            isLink = false,
-            isPrimary = false,
             type = 'button',
             href,
             size = 'normal',
+            variant = 'filled',
+            context = 'primary',
```
eszthoff, line 47: I'm not sure these are good default. It will change the default button from a neutral to a primary look. this will make migration more difficult. But if this is the main button we are going to use from here on, it might make sense. Can you check with Dima?

I also think the BREAKING CHANGE description of the PR is not explicit about it, we should add it once we are sure.

src/components/Buttons/Button/__tests__/Button.spec.tsx
```diff
@@ -23,17 +23,17 @@ describe('<Button> that renders a button', () => {
...
         );
 
         expect(view.container).toMatchSnapshot();
         const button = screen.getByRole('button');
-        expect(button).toHaveClass('Button--isPrimary');
+        expect(button).toHaveClass('Button--context_primary');
         expect(button).toHaveClass('Button--size_large');
         expect(button).toHaveClass('Button--isBlock');
-        expect(button).toHaveClass('Button--isLoading');
+        // TODO: expect(button).toHaveClass('Button--isLoading');
```
eszthoff, line 36: ;-)

src/components/Buttons/Button/__tests__/__snapshots__/Button.spec.tsx.snap
```diff
@@ -3,21 +3,42 @@
 exports[`<Button> that renders a button should add classes when props are changed 1`] = `
 <div>
   <button
-    class="Button Button--isPrimary Button--isBlock Button--isLoading Button--size_large"
+    class="Button Button--isBlock Button--size_large Button--variant_filled Button--context_primary"
+    disabled=""
```
eszthoff, line 7: wouldn't we expect this to be `true` for a loading button?

src/components/DatePicker/CalendarHeader/__tests__/__snapshots__/CalendarHeader.spec.tsx.snap
```diff
@@ -6,7 +6,8 @@ exports[`CalendarHeader should render correctly 1`] = `
     class="CalendarHeader"
   >
     <button
-      class="Button CalendarHeader__navButton"
+      class="Button Button--variant_filled Button--context_primary CalendarHeader__navButton"
```
eszthoff, line 9: I think these should stay secondary buttons

src/components/Input/Input.tsx
```diff
@@ -6,7 +6,7 @@ import { ValidationContext, InputType, Size } from '../../constants';
 // Any other attributes (onChange, onKeyUp etc.) are
 // supported although not defined in props type definition
 export interface Props extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'> {
-    /** The input field context (e.g. brand, critical, success etc. - defaults to brand) */
+    /** The input field context (e.g. primary, critical, success etc. - defaults to primary) */
```
eszthoff, line 9: Let's keep these changes to ticket that will update this component

src/components/Link/Link.tsx
```diff
@@ -6,15 +6,15 @@ export interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
     /** Link content */
     children?: React.ReactNode;
     /** Color context */
-    context?: 'brand' | 'neutral';
+    context?: 'primary' | 'neutral';
```
eszthoff, line 9: let's keep these changes to the ticket that will migrate this component

src/components/Link/Link.tsx
```diff
@@ -6,15 +6,15 @@ export interface Props extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
...
     /** Do not underline text on hover */
     dontDecorateOnHover?: boolean;
 }
 
 const { block } = bem('Link', styles);
 
 export const Link = React.forwardRef<HTMLAnchorElement, Props>(
-    ({ children, context = 'brand', dontDecorateOnHover = false, ...rest }, ref) => {
+    ({ children, context = 'primary', dontDecorateOnHover = false, ...rest }, ref) => {
```
eszthoff, line 17: s/a

src/components/List/ListItem/ListItem.scss
```diff
@@ -1,4 +1,4 @@
-$contexts: (neutral, brand, info, success, cautious, critical);
+$contexts: (neutral, primary, info, success, cautious, critical);
```
eszthoff, line 1: s/a

src/components/LoadingSpinner/LoadingSpinner.scss
```diff
@@ -1,3 +1,13 @@
+@use "sass:map";
+$contexts: (primary, secondary, critical, whiteSpinner);
```
eszthoff, line 2: I think better to call the last one `invert` as this is the word used for icons displayed on an inverted BG. E.g. in a dark theme it will more like to be black then white.

src/components/LoadingSpinner/LoadingSpinner.scss
```diff
@@ -1,3 +1,13 @@
...
+
+$contextMap: (
+    'secondary': 'neutral',
+    'primary': 'primary',
+    'critical': 'critical',
+    'whiteSpinner': 'whiteSpinner'
```
eszthoff, line 8: s/a

src/components/LoadingSpinner/LoadingSpinner.scss
```diff
@@ -38,12 +48,22 @@
...
         stroke-dashoffset: 0;
         stroke-linecap: round;
         animation: dash 1.5s ease-in-out infinite;
         transition: stroke var(--transition-duration);
-        stroke: var(--color-brand-50);
+        @each $context in $contexts {
+            &--context_#{ $context } {
+                $mappedContext: map.get($contextMap, $context);
+                @if ($mappedContext == 'whiteSpinner') {
+                    stroke: var(--color-background);
```
eszthoff, line 61: please check what is used in figma. `--color-background` is a "left over" from previous tokes we had, we should use the new one when migrating things.

src/components/LoadingSpinner/LoadingSpinner.scss
```diff
@@ -38,12 +48,22 @@
...
+                } @else {
+                    stroke: var(--color-#{$mappedContext}-30);
```
eszthoff, line 63: please use semantic token here too.

src/components/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -8,11 +9,12 @@ export interface Props extends React.HTMLAttributes<HTMLDivElement> {
     centerIn?: 'parent' | 'viewport';
     /** Loading text */
     children?: React.ReactNode;
-
     /** Hides the spinner when true */
     hidden?: boolean;
     /** Custom spinner size (will affect both width and height) */
     size?: number;
+    /**  Define the context, eg. primary, secondary, critical */
+    context?: ButtonContext | 'whiteSpinner';
```
eszthoff, line 17: s/a. let's name it `inverted` or similar

src/components/PopupBase/__tests__/__snapshots__/PopupBase.spec.tsx.snap
```diff
@@ -4,7 +4,8 @@ exports[`<PopupBase> that adds basic anchor/popup functionality to rendered comp
 <body>
   <div>
     <button
-      class="Button"
+      class="Button Button--variant_filled Button--context_primary"
```
eszthoff, line 7: can you check if these buttons are only set in the test, or if we should preserve the neutral look for them?

src/components/Text/MarkedText/__tests__/MarkedText.spec.tsx
```diff
@@ -47,7 +47,7 @@ describe('<MarkedText> that renders a text block while marking matched substring
 
     it('should pass props to Text', () => {
         view = render(
-            <MarkedText marker="text" inline context="brand">
+            <MarkedText marker="text" inline context="primary">
```
eszthoff, line 50: irrelevant change

src/components/Text/Text.scss
```diff
@@ -11,7 +11,7 @@
         }
     }
 
-    @include context-attribute-with-color('color', (brand, info));
+    @include context-attribute-with-color('color', (primary, info));
```
eszthoff, line 14: s/a

src/components/Text/Text.tsx
```diff
@@ -8,8 +8,8 @@ export interface Props extends React.HTMLAttributes<HTMLElement> {
     children: React.ReactNode;
     /** Text should be rendered inline */
     inline?: boolean;
-    /** The context of the text, effecting its color (e.g. brand, info, critical, success etc. 'neutral' added as special context here) */
-    context?: Context | 'neutral' | 'brand' | 'default';
+    /** The context of the text, effecting its color (e.g. primary, info, critical, success etc. 'neutral' added as special context here) */
+    context?: Context | 'neutral' | 'primary' | 'default';
```
eszthoff, line 12: s/a

eszthoff (changes requested):

eszthoff:
Check the [docs](https://github.com/textkernel/oneui/pull/1234/commits/1a84a26130f22ab466c89c32847d209ac2ca0d9c) about breaking change comments, the syntax is important for the library to pick it up. (singular change, I think it should be on one line). Please adjust the description to match, then when squashing the commits, you can just copy paste it.

src/components/Buttons/Button/__tests__/__snapshots__/Button.spec.tsx.snap
janseke10: As I understand it, in HTML, boolean attributes are considered truthy if the value is an empty string. So this is indeed true. If it isn't there, it is false.

src/components/Buttons/Button/Button.tsx
```diff
@@ -39,17 +39,25 @@ export const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, Props>(
...
         ref
     ) => {
+        const buttonRef = React.useRef<HTMLButtonElement>();
+        const [buttonWidth, setButtonWidth] = React.useState<number>();
+        // figure out width of button, and add that width as styling
+        React.useLayoutEffect(() => {
+            if (buttonRef && buttonRef.current) {
+                setButtonWidth(buttonRef.current.offsetWidth);
+            }
+        }, [buttonRef]);
```
mukiienko, line 59: In this case, we only measure the element's width during the initial phase. If the width or text changes during the button's lifecycle, it could result in an incorrect width. For example, if we switch the language and all the text changes, the button may have the wrong width. Although this might be an edge case, what do you think about measuring the element's width on demand—specifically when the loading state is true? This way, we will always have the correct width for both the button and the loading state:
```
....
        const buttonRef = React.useRef<HTMLButtonElement | HTMLAnchorElement>(null);
        const [buttonWidth, setButtonWidth] = React.useState<number | undefined>();
        const prevIsLoadingRef = React.useRef(isLoading);

        if (buttonRef.current && isLoading && !buttonWidth) {
            setButtonWidth(buttonRef.current.offsetWidth);
        }

        if (!isLoading && buttonWidth) {
            setButtonWidth(undefined); // Reset to auto width when not loading
        }

        if (typeof children !== 'number' && !children) {
            return null;
        }
...
style={{ width: buttonWidth ? `${buttonWidth}px` : 'auto' }}
....
```

mukiienko (changes requested):

mukiienko:
Loading state doesn't work for link:
@

src/components/Tag/Tag.tsx
```diff
@@ -13,7 +13,7 @@ export interface Props extends React.HTMLAttributes<HTMLDivElement> {
     /** Max-width of a component */
     maxWidth?: string;
     /** Size of the text */
-    size?: 'small' | 'normal' | 'large';
+    size?: Size;
```
eszthoff, line 16: nice

eszthoff (changes requested):

src/components/Buttons/Button/Button.scss
```diff
@@ -1,169 +1,186 @@
...
-    background-color: var(--color-background);
-    border-color: var(--color-neutral-80);
     cursor: pointer;
-    font: {
-        size: var(--font-size-base);
-        weight: var(--font-weight-bold);
-        family: var(--font-family-primary);
-    }
+    composes: OneUI-label-text-bold from global;
+    font-family: var(--font-family-primary);
```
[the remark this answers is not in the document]
janseke10: But... it looks different :) 
Screenshot 2024-08-19 at 11 45 12
Screenshot 2024-08-19 at 11 45 24

eszthoff (approved):

src/components/Buttons/Button/Button.tsx
mukiienko: I still think we don't need `React.useLayoutEffect` for calculation width.

mukiienko:
> Loading state doesn't work for link: Screenshot 2024-08-15 at 11 29 14

Loading state has wrong width:

https://github.com/user-attachments/assets/0516e42f-8ac5-4322-9bf1-26386d73dfc9

mukiienko (approved):

src/components/Buttons/Button/Button.scss
```diff
@@ -1,169 +1,213 @@
...
+                    &[disabled],
+                    &.Button--isLoading {
+                        color: var(--color-text-disabled);
+                        background-color: var(--transparent);
+                        &:hover,
+                        &:active {
+                            background-color: var(--transparent);
+                        }
+                    }
+                } @else if $variant == ghost {
```
eszthoff, line 192: we are already in the context of `if $variant == ghost` so it will always be `true`, so maybe we can just have `@else {...`.

mukiienko: Good catch!

PR #1249, opened 2024-09-05
feat(Link): implement new design

eszthoff:
Story: [ONEUI-477](https://textkernel.atlassian.net/browse/ONEUI-477)

BREAKING CHANGE: `context` and `doNotDecorateOnHover` props are no longer supported for Link component

### DIFF
--- a/src/components/Link/Link.scss
+++ b/src/components/Link/Link.scss
@@ -2,7 +2,7 @@
     color: var(--color-link-default);
     text-decoration: underline;
 
-    &:focus {
+    &:focus-visible {
         outline: 2px solid var(--color-border-information-subtle-default, #66AFFF);
     }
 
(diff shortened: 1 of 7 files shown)


src/components/Link/Link.scss
```diff
@@ -1,33 +1,26 @@
...
-    text-decoration: var(--link-decoration-normal);
-    transition: color var(--transition-duration);
+    color: var(--color-link-default);
+    text-decoration: underline;
 
-    &:hover,
     &:focus {
-        color: var(--color-brand-60);
-        text-decoration: var(--link-decoration-hover);
+        outline: 2px solid var(--color-border-information-subtle-default, #66AFFF);
```
AnastasiiaMP, line 6: I learned today about `hover-visible` (thank you 🙏) and It seems to be a perfect place to use it instead of just `:focus` here. 
This way, we can avoid showing an outline on mouse clicks and only display it during keyboard interactions as specified in the ticket.

AnastasiiaMP (changes requested):
There is just a small comment (today is a `:focus-visible` day 😅), other than that everything looks great

AnastasiiaMP (approved):

PR #1263, opened 2024-09-18
feat(IconButton): new component

andrentk:
Story: [ONEUI-473](https://textkernel.atlassian.net/browse/ONEUI-473)

### DIFF
--- a/src/components/Buttons/IconButton/IconButton.scss
+++ b/src/components/Buttons/IconButton/IconButton.scss
@@ -1,0 +1,32 @@
+@import '../Button/Button';
+
+.IconButton {
+    &--size {
+        &_small {
+            padding: var(--space-75);
+        }
+        &_medium {
+            padding: var(--space-100);
+        }
+        &_large {
+            padding: 10px;
+        }
+    }
+}
+
+.IconButton__icon {
+    &--size {
+        &_small {
+            width: var(--space-150);
+            height: var(--space-150);
+        }
+        &_medium {
+            width: var(--space-200);
+            height: var(--space-200);
+        }
+        &_large {
+            width: var(--space-250);
+            height: var(--space-250);
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/components/Buttons/IconButton/IconButton.scss
```diff
@@ -0,0 +1,32 @@
...
+.IconButton {
+    &--size {
+        &_small {
+            padding: var(--space-75);
+        }
+        &_medium {
+            padding: var(--space-100);
+        }
+        &_large {
+            padding: 10px;
```
eszthoff, line 12: Please double check with Dima. We shouldn't use anything that isn't in the tokens.

src/components/Buttons/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
+import { bem } from '../../../utils';
+import { ButtonContext, ButtonType, ButtonVariant, Size } from '../../../constants';
+import { Button } from '../Button';
+import styles from './IconButton.scss';
+import { LoadingSpinner } from '../../LoadingSpinner';
+
+export interface Props extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'> {
```
eszthoff, line 8: would it make sense to extend the ButtonProps here, and reuse that even more?

src/components/Buttons/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Button
+                ref={ref}
+                disabled={disabled}
+                size={size}
+                variant={variant}
+                context={context}
+                href={href}
+                type={type}
+                {...rest}
+                {...block({ size })}
```
eszthoff, line 66: you should pass `rest` here too, so that className is added if it is passed as a prop (e.g. spacing or something from the application side)

eszthoff (changes requested):

eszthoff (approved):

eszthoff (approved):

PR #1266, opened 2024-09-24
feat(Teaser): redesign Teaser

andrentk:
Story: [ONEUI-447](https://textkernel.atlassian.net/browse/ONEUI-447)


src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
+    timestamp?: string;
+    /** Render tags */
+    tags?: React.ReactElement<TagProps>[];
+    /** Matching indicator percentage */
+    matchingIndicatorPercentage?: number;
 }
 
 const { block, elem } = bem('Teaser', styles);
 
+const InfoComponent = ({ text, href, icon }: Info) => {
```
eszthoff, line 50: I see 2 patters in the component, one reuses `textComponent` perhaps with additional classes using React.cloneElement, two writing the text component from scratch. I'm good with either by perhaps not with both, if you see what I mean. 

Also, maybe it makes sense to move it to a separate file, so that it is easier to test?

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
     title,
-    subTitle = '',
-    location = '',
-    details = '',
-    statuses,
-    disabled = false,
+    subtitle,
+    hasCheckbox = false,
+    id,
+    isSelected,
```
eszthoff, line 84: I think should be `false` by default, no?

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
-                    <Text inline context="neutral" {...elem('location')}>
-                        {location}
+            {hasCheckbox && <Checkbox id={checkboxId} checked={isSelected} onChange={onChange} />}
+            <div {...elem('column')}>
+                <div {...elem(['row', 'titleRow'])}>
+                    <div {...elem('row')}>
+                        <Text inline title={title} {...elem('title')}>
+                            {title}
+                        </Text>
+                        {matchingIndicatorPercentage !== undefined &&
```
eszthoff, line 118: maybe this could be a separate component? I can see how we might want to reuse it at some point...

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
-            {...elem('details')}
-        >
-            {details}
-        </Text>
-    );
+    const useId = React.useId();
+    const checkboxId = id ?? useId;
+
+    const computeMatchingIndicatorPercentageColor = (value: number) => {
+        if (value <= 30) {
```
eszthoff, line 100: maybe you can define some constants for it, so that it is not magical?

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
+    onChange?: React.ChangeEventHandler<HTMLInputElement> | undefined;
+    /** First line of information */
+    primaryInfo?: Info;
+    /** Second line of information */
+    secondaryInfo?: Info;
+    /** Third line of information */
+    tercearyInfo?: Info;
+    /** Line containing the source of the information */
+    sourceInfo?: Info;
+    /** Render little blue bubble */
```
eszthoff, line 38: please extend this description to explicitly say that `undefined` means it won't be shown at all.

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
-                            </span>
-                        ))}
+                {primaryInfo && <InfoComponent text={primaryInfo.text} href={primaryInfo.href} />}
+                {secondaryInfo && (
+                    <InfoComponent text={secondaryInfo.text} href={secondaryInfo.href} />
+                )}
+                {tercearyInfo && (
+                    <InfoComponent text={tercearyInfo.text} href={tercearyInfo.href} />
+                )}
+                {sourceInfo && <InfoComponent text={sourceInfo.text} icon={sourceInfo.icon} />}
```
eszthoff, line 173: wouldn't source be clickable e.g. in Jobfeed? How will that work?

stories/molecules/Teaser.stories.tsx
```diff
@@ -16,48 +17,23 @@ type Story = StoryObj<typeof Teaser>;
...
-    name: 'Without statuses',
-    args: {
-        title: <span title={titleText}>{titleText}</span>,
-        subTitle: 'Awsome inc.',
-        location: 'Melbourne',
-        details: 'It was posted here',
+        return (
+            <Teaser {...args} isSelected={isSelected} onChange={() => setIsSelected(!isSelected)} />
+        );
     },
```
eszthoff, line 38: I think it would be good to add some example that show how it might look like for Search and for Jobs Data and duplicate some examples shown in Figma. Now it is pretty abstract for me when looking at storybook. 
It can be one story or several, but some more "real world" examples would be nice to see.

eszthoff (changes requested):

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
+type Info = {
+    text: string;
+    href?: string;
+    icon?: React.ReactElement;
 };
 
 export interface Props extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title'> {
     /** The title of entity. Note: it is rendered within a <span> */
-    title: React.ReactNode;
+    title: string;
```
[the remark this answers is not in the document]
andrentk: In figma there is no variation for title and subtitle besides a string. With all the 3 whole component variations present in figma not containing anything else, I assumed that is not needed.

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
     /** The subtitle of the entity. Note: it is rendered within a <span> */
-    subTitle?: React.ReactNode;
-    /** The location of the entity. Note: it is rendered within a <span> */
-    location?: React.ReactNode;
-    /** Details related to the entity. Note: it is rendered within a <span> */
-    details?: React.ReactNode;
-    /** Statuses of entity containing status label and tooltip message */
-    statuses?: Status[];
+    subtitle?: string;
```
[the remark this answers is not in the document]
andrentk: https://github.com/textkernel/oneui/pull/1266#discussion_r1778852420

src/components/Teaser/Teaser.tsx
```diff
@@ -1,104 +1,182 @@
...
+    /** Second line of information */
+    secondaryInfo?: Info;
+    /** Third line of information */
+    tercearyInfo?: Info;
+    /** Line containing the source of the information */
+    sourceInfo?: Info;
+    /** Render little blue bubble */
+    isVisited?: boolean;
+    /** Render timestamp information */
+    timestamp?: string;
```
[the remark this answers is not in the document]
andrentk: What do you think about `date` or `dateString`?

src/components/Teaser/Teaser.tsx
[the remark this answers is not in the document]
eszthoff: ok, we can try. I just recall we had to change that in the current component, but maybe those use cases are now covered by having more lines..

src/components/Teaser/Teaser.tsx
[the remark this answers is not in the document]
eszthoff: not `dateString` that's also something not-human-readable. `date` or `time` alone is ok for me.

src/components/Teaser/Teaser.tsx
andrentk: You can pass a href as well to the source if there is a need to be a Link

stories/molecules/Teaser.stories.tsx
andrentk: Added

src/components/Teaser/Teaser.tsx
```diff
@@ -35,75 +35,75 @@ export interface Props extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title
     tercearyInfo?: Info;
     /** Line containing the source of the information */
     sourceInfo?: Info;
-    /** Render little blue bubble */
+    /** Render little blue bubble, if undefined it will not be rendered */
```
eszthoff, line 38:
```suggestion
    /** Render visit indicator: `true` - little blue bubble | `false` - empty circle | `undefined` - nothing will be rendered */
```

stories/molecules/Teaser.stories.tsx
```diff
@@ -20,14 +20,98 @@ export const WithAllFields: Story = {
...
+    render: (args) => {
+        const [isSelected, setIsSelected] = React.useState(false);
+
+        return (
+            <Teaser {...args} isSelected={isSelected} onChange={() => setIsSelected(!isSelected)} />
+        );
+    },
+};
+
+export const VariantThree: Story = {
```
eszthoff, line 95: It would be good to see a realistic example for JD. Sorry I still have difficulty seeing how it will look like. Could you render it in a div with some given width (500px?)? Can you use some job title (maybe even a long one to see how it will be truncated); company name, show multiple sources like `jobboard.com (+ 4 more)`, longer time indicator like `about 1 month ago`; no visitation indication. Maybe at least one variant could also loose the checkbox.

stories/molecules/Teaser.stories.tsx
```diff
@@ -20,14 +20,98 @@ export const WithAllFields: Story = {
...
+            <div style={{ display: 'flex', gap: '8px' }}>
+                <Tag bgColor="yellow" size="small">
+                    First
+                </Tag>
+                <Tag bgColor="lightblue" size="small">
+                    Second
+                </Tag>
+            </div>
+        ),
+    },
```
eszthoff, line 40: I cannot see any effect on the UI when I set `disabled` or `isSelected`. Also, there is no callback on the checkbox, is there?

eszthoff (changes requested):

stories/molecules/Teaser.stories.tsx
andrentk: disabled it was a mistake, that prop should not exist. at least there is no design for disabled. isSelected is the controllable variable of the checkbox status. There is the onChange that is triggered when checkbox status changes.

eszthoff: disabled is used, so we need to keep it, otherwise we have a regression. Can you ask Dima to give you a design? Maybe just make the title gray will do...
For example on where we use it, I think it is for some expired jobs in some of our lists

src/components/Teaser/Teaser.scss
```diff
@@ -80,4 +85,12 @@
     &__bottom {
         margin-top: var(--space-100);
     }
+
+    &__lineClamp {
```
eszthoff, line 89: nice approach :)

src/components/MatchingIndicator/MatchingIndicator.tsx
```diff
@@ -31,27 +31,29 @@ export const MatchingIndicator: React.FC<Props> = ({ percentage }) => {
...
-                cx="50%"
-                cy="50%"
-                stroke={computeMatchingIndicatorPercentageColor(percentage)}
-                fill="none"
-                strokeWidth="3"
-                strokeLinecap="round"
-                strokeDasharray={`${(percentage / 100) * 56}, 56`}
-            />
-        </svg>
+        <div {...block()}>
```
eszthoff, line 34: if you define the props with `extends HTMLElementAttributes` (check syntax!) and then pass `rest` to block it will reuse any `classname` passed to the props. This we normally should do in all components. Also spreading `rest` on this div will ensure that other attributes (e.g. style, or data-x) will not get lost. 
Also using this approach you can save the extra `div` you use for adding a class to the wrapper, and can pass `elem` to the indicator directly.

stories/molecules/Teaser.stories.tsx
```diff
@@ -16,48 +15,152 @@ type Story = StoryObj<typeof Teaser>;
...
+            <Teaser {...args} isSelected={isSelected} onChange={() => setIsSelected(!isSelected)} />
+        );
+    },
+};
+
+export const VariantThree: Story = {
+    name: 'Variant Three',
+    args: {
+        title: 'John Doe',
+        subtitle: 'Surgeon',
```
eszthoff, line 142:
```suggestion
        subtitle: 'Super company you want to work for',
```

eszthoff (approved):

PR #1267, opened 2024-09-26
feat(SearchInput): new component

janseke10:
[ONEUI-456](https://textkernel.atlassian.net/jira/software/c/projects/JF/boards/196?assignee=712020%3A84543f8b-4a3d-4e77-9457-65010a0ec912&selectedIssue=ONEUI-456)

# Checklist


eszthoff:
@janseke10 Please use the SemVer convention for naming the PR (`feat(component): change`) and make sure to use that text as the commit message when merging the PR (this is important for the CI/CD to produce the change-logs correctly). Also please add a link to the story in the description. Thanks

eszthoff:
@janseke10 I thought the idea here was to reuse the Input component, and just make it more specific (set props, add some children etc.) similar for how IconButton reuses Button. Is there a reason that you decided to implement all features from scratch instead?

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+                    <Search
+                        {...elem('icon', { type: 'search', bold: hasText, size })}
+                        viewBox="0 0 24 24"
+                    />
+                    <input
+                        {...elem('input', { size })}
+                        {...rest}
+                        id={idRef}
+                        ref={ref}
+                        type="text"
```
eszthoff, line 96: why using type `text` for a `search` input?

janseke10: type="search" adds a 'clear' icon itself. That's why I chose to go with text instead. I could add something like this:
input[type="search"]::-webkit-search-cancel-button {
  -webkit-appearance: none;
  height: 1em;
  width: 1em;
  border-radius: 50em;
  background: url(https://pro.fontawesome.com/releases/v5.10.0/svgs/solid/times-circle.svg) no-repeat 50% 50%;
  background-size: contain;
  opacity: 0;
  pointer-events: none;
}

To get rid of the button... But i just didn't think that was the neatest way of doing things.

janseke10:
> @janseke10 I thought the idea here was to reuse the Input component, and just make it more specific (set props, add some children etc.) similar for how IconButton reuses Button. Is there a reason that you decided to implement all features from scratch instead?

I decided on this method, because so much had to change regarding the styling. <Input /> doesn't take any children, so adding an icon had to happen by adding it 'next' to the input, which means the borders had to change, etc. And then there were other requirements with the colors of the icons, or how they should appear or disappear when there is text or not. I thought using the Input component as inspiration and then changing it around to meet the new requirements was the easiest and cleanest way of creating this component.
@eszthoff

janseke10:
> @janseke10 Please use the SemVer convention for naming the PR (`feat(component): change`) and make sure to use that text as the commit message when merging the PR (this is important for the CI/CD to produce the change-logs correctly). Also please add a link to the story in the description. Thanks

Story link is already there! I thought it was purely the commits that needed to follow the SemVer. I'll change it, thanks.

src/components/Input/SearchInput/SearchInput.tsx
eszthoff: I think you can use `appearance: none` (see also Checkbox and Radio button uses that) to cancel browser specific looks and add your own. 
I think `type="search"` does more then the clear button though. E.g. on mobile (which we might want to support one day) it changes the enter button; also there is a different behaviour about caching things people typed (that's for desktop too).

src/components/Input/__tests__/Input.spec.tsx
```diff
@@ -89,16 +92,100 @@ describe('<Input> that renders an input field', () => {
...
     it('should add string html attributes correctly', () => {
-        view = renderComponent({ 'data-test': 'something' });
+        view = renderInputComponent({ 'data-test': 'something' });
 
         expect(screen.getByRole('textbox')).toHaveAttribute('data-test', 'something');
     });
 });
+
+describe('<SearchInput> that renders an input field', () => {
+    let view: RenderResult;
```
andrentk, line 110: since there is no beforeEach rendering the result, it may be better to have a local variable inside of each test, not sure if with test parallelisation it would conflict somehow.

src/components/Input/__tests__/Input.spec.tsx
```diff
@@ -11,14 +12,16 @@ const defaultProps: Props = {
     context: undefined,
 };
 
-const renderComponent = (props = {}) => render(<Input {...defaultProps} {...props} />);
+const renderInputComponent = (props = {}) => render(<Input {...defaultProps} {...props} />);
+const renderSearchInputComponent = (props = {}) =>
+    render(<SearchInput {...defaultProps} {...props} />);
 
 describe('<Input> that renders an input field', () => {
     let view: RenderResult;
```
andrentk, line 20: same as below

src/components/Input/Input.tsx
```diff
@@ -30,14 +30,19 @@ interface BaseProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 's
...
      *  to avoid "jumping" UI. */
     reserveErrorMessageSpace?: boolean;
 }
 
-export type Props = BaseProps & (ErrorStateProps | { context?: undefined; errorMessage?: never });
+export type Props = BaseProps &
+    (ErrorStateProps | { context?: undefined | string; errorMessage?: undefined | string });
 
-const { block, elem } = bem('Input', styles);
+const { block, elem } = bem('InputContainer', styles);
```
andrentk, line 45: I would say to keep the name as Input here and use the block in the upper element, maybe transforming it into a div.

andrentk (commented):
for some reason the font-size applied in the input is 13.333px
image

eszthoff (changes requested):

src/components/Input/Input.tsx
```diff
@@ -30,14 +30,19 @@ interface BaseProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 's
...
+    /** Icon after input, which can trigger callback as a button, optional */
+    trailingIcon?: { icon: React.ReactElement; callback: () => void };
     /**  Controls whether space is reserved for error messages under the input field when validation is expected
      *  to avoid "jumping" UI. */
     reserveErrorMessageSpace?: boolean;
 }
 
-export type Props = BaseProps & (ErrorStateProps | { context?: undefined; errorMessage?: never });
+export type Props = BaseProps &
+    (ErrorStateProps | { context?: undefined | string; errorMessage?: undefined | string });
```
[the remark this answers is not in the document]
janseke10: Because it could be that the error is passed to the SearchInput as a prop, i thought. And in that case it has to be passed on to Input... 
I was in doubt when I did this though ;) I thought it was a bit strange anyway, that it is set to undefined or never, but you also have this:
type ErrorStateProps = {
    /** The input error field context (critical) */
    context: ErrorContext;
    /** This message will be rendered under the input when context critical will be applied */
    errorMessage: string;
};

I'm not familiar with this structure!

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import Search from '@material-design-icons/svg/round/search.svg';
+import Clear from '@material-design-icons/svg/round/cancel.svg';
+
+import { Input, Props } from '../Input';
```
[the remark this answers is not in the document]
janseke10: But if you can't pass a label here or errormessage, it won't be passed to the input either. I thought because of the sentence 'All variants, states, sizes and types are supported' in the ticket, these things should be passed on.

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+                readOnly={readOnly}
+                isBlock={isBlock}
+                context={context}
+                size={size}
+                label={label}
+                helperText={helperText}
+                type="search"
+                value={value}
+                onChange={(e) => {
+                    setValue(e.target.value);
```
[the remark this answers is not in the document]
janseke10: Uhm i don't think so. I could add to trigger rest.onChange if it exists? Like I did in Input.tsx?

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+    (
+        {
+            id,
+            children,
+            context = undefined,
+            disabled = false,
+            readOnly = false,
+            isBlock = false,
+            size = 'medium',
+            value: initialValue = '',
```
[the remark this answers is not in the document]
janseke10: What do you mean exactly? Try the SearchInput? It's a component in input

src/components/Input/Input.tsx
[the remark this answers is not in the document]
eszthoff: the idea critical part here is that you have `ErrorStateProps | {...}`. This means you either pass the ErrorStateProps which will defined those details, or you don't pass neither. So in this way we group props into a logical unit that means if context is undefined, errorMessage will also not be there. Something like that. Does that clarify things?
@AnastasiiaMP came up with this pattern recently, I think we used in it a 2-3 components now.

With your change, this logic is overwritten, as you allow for the error message to be there even in the second option. so this change should be reverted.

src/components/Input/SearchInput/SearchInput.tsx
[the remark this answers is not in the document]
eszthoff: yes, it should. How else will the application actually react to what the user types?

src/components/Input/SearchInput/SearchInput.tsx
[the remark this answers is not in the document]
eszthoff: how can I set `initialValue` in storybook? I wanted to test if this assignment here won't make a bug. Like if you write something and then clear the field, will it be empty?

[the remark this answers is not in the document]
janseke10: InitialValue takes the value of 'value', unless value is undefined, then it becomes ''. As it isn't part of the props, you can't set it in storybook, just like you can't set it for input, without simply typing it. I don't know if that should be a possibility? Then I need to add value to the props. 

I used this method so I could then create a state of value, setValue and directly use that in the input.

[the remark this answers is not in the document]
eszthoff: Input has the following props: `interface BaseProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'> {...}`. So it has all the native attributes possible, including value. This way you can implement a controlled component. These are passed to Input correctly via `rest` that is applied to the `<input>` 
I'm a bit confused by the code, and I wonder how it will work if someone implements the component as controlled. Maybe you could add a story where you pass `value`. But even if not, we will be sure that it is possible to implement and it works as expected

src/components/Input/SearchInput/SearchInput.tsx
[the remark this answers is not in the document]
janseke10: I did change it like you suggested, but storybook still shows those options. I can't change it, I tried all kinds of things... Most common you use argTypes to set it to disabled, but it is aware that those props are not longer part of searchInput, so I can't do that :') 

Anyway, hope it's okay like this

src/components/Input/SearchInput/SearchInput.tsx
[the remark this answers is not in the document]
janseke10: Tell me about it, haha. I'm also confused. But I can't keep SearchInput as a controlled component, right? I wouldn't know how to, with the icons influencing the value...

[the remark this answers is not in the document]
eszthoff: hhhmmm... In JF we almost always use controlled components, since we need to do something with the value, like filter results that we show and so on. Once we have it in the code, it easier to use controlled components.

Maybe you can think of an other approach, and use ref on input to see if it has a value or not, and set the states based on that? Would that work?

src/components/Input/Input.tsx
```diff
@@ -84,27 +83,37 @@ export const Input = React.forwardRef<HTMLInputElement, Props>(
...
-            <>
+            <div
+                {...block({
+                    withIcon: leadingIcon,
+                    context,
+                    size,
+                    isBlock,
+                    disabled,
+                    readOnly,
+                })}
```
eszthoff, line 94: either the whole `rest` or at least `classname` from there should be passed here, so that classes defined on the application side are also applied. 
Also probably `style` should be applied to this element, rather then to input (even though other rest attributes do belong to there)

src/components/Input/Input.scss
```diff
@@ -83,40 +87,36 @@
         background-color: inherit;
         color: inherit;
         box-sizing: border-box;
+        composes: OneUI-label-text from global;
 
         &--type_search::-webkit-search-cancel-button {
             display: none;
```
eszthoff, line 93: doesn't `appearance: none` works better? I should check it in all browsers, did you try them?

src/components/Input/SearchInput/__tests__/SearchInput.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+        );
+    });
+
+    it('should render the search icon bold when there is a value', () => {
+        renderSearchInputComponent({ value: 'Some value' });
+
+        const searchIcon = screen.getAllByTestId('default-icon')[0];
+        expect(searchIcon).toHaveClass('Input__icon--bold');
+    });
+    it('should not render the search icon bold when there is a value', () => {
```
eszthoff, line 51:
```    it('should not render the search icon bold when there is no value', () => {
```

src/components/Input/Input.tsx
eszthoff: Sorry, I wasn't clear. The `...rest` should go into the list of thing `block` called with. It knows how to combine the classname there. It is indeed a good idea to pass the `styles` here though. I don't think block manages that.

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -51,7 +46,19 @@ export const SearchInput = React.forwardRef<HTMLInputElement, SearchInputProps>(
...
                 onChange={handleOnChange}
                 leadingIcon={{ icon: <Search /> }}
-                trailingIcon={{ icon: <Clear />, callback: () => setValue('') }}
+                trailingIcon={{
+                    icon: <Clear />,
+                    callback: () => {
+                        if (inputRef && 'current' in inputRef && inputRef.current) {
+                            (inputRef.current as HTMLInputElement).value = '';
+                            if (rest.onChange) {
+                                rest.onChange({
```
eszthoff, line 55: why do you prefer this `rest.prop` notation instead of destructuring it from the start from the props? That what normally we do.
Also I think here, you don't have to destructure things like `IsBlock` and whatever else is not used in the code for this component, and just pass those to Input by spreading them here (`...rest`).

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+                                } as React.ChangeEvent<HTMLInputElement>);
+                            }
+                        }
+                    },
+                }}
+                reserveErrorMessageSpace={reserveErrorMessageSpace}
+                {...(context && errorMessage
+                    ? { context, errorMessage }
+                    : { context: undefined, errorMessage: undefined })}
+                {...rest}
```
eszthoff, line 66: A way to make sure that things aren't overwritten by props if e.g. the applicaiton doesn't uses TS and props like `type` is defined even though it shouldn't is by adding this spread at the beginning.

stories/atoms/Input.stories.tsx
```diff
@@ -27,3 +27,26 @@ export const _Input: StoryInput = {
...
+        label: 'Label',
+        helperText: 'Optional helper text',
+        errorMessage: 'Error message',
+        context: undefined,
+        size: 'medium',
+    },
+    render: (args) => {
+        const [value, setValue] = React.useState('');
+        const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+            console.log(e.target.value);
```
eszthoff, line 46:
```suggestion
            console.log('onChange is called with target.value:', e.target.value);
```

eszthoff (changes requested):

src/components/Input/Input.tsx
```diff
@@ -67,12 +67,11 @@ export const Input = React.forwardRef<HTMLInputElement, Props>(
     ) => {
         const fallbackId = React.useId();
         const idRef = id || fallbackId;
-        const [hasText, setHasText] = React.useState<boolean>(!!value);
+        const { onChange, style } = rest;
```
eszthoff, line 70: you don't need it to do here. You can just add these where the props are destructured (line 63 or so)

src/components/Input/Input.tsx
```diff
@@ -67,12 +67,11 @@ export const Input = React.forwardRef<HTMLInputElement, Props>(
...
 
         const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-            setHasText(!!e.target.value);
-            if (rest.onChange) {
-                rest.onChange(e);
+            if (onChange) {
+                onChange(e);
```
eszthoff, line 74: Looks like now this function does what the native callback would do, so maybe we can just leave it out, no?

src/components/Input/Input.tsx
```diff
@@ -106,12 +105,7 @@ export const Input = React.forwardRef<HTMLInputElement, Props>(
                         readOnly,
                     })}
                 >
-                    {leadingIcon &&
-                        React.cloneElement(leadingIcon.icon, {
-                            viewBox: '0 0 24 24',
-                            ...elem('icon', { type: 'leading', bold: hasText, size }),
-                            onClick: leadingIcon.callback ? leadingIcon.callback : undefined,
-                        })}
+                    {leadingIcon || <></>}
```
eszthoff, line 108: Why do we need this fallback? I mean if `leadingIcon: ReactElement` then it should render in the DOM ok. What am I missing?

src/components/Input/Input.tsx
```diff
@@ -125,12 +119,7 @@ export const Input = React.forwardRef<HTMLInputElement, Props>(
                         aria-invalid={context === 'critical' ? 'true' : undefined}
                         data-lpignore={isLastPassDisabled}
                     />
-                    {trailingIcon &&
-                        React.cloneElement(trailingIcon.icon, {
-                            viewBox: '0 0 24 24',
-                            ...elem('icon', { type: 'trailing', visible: hasText, size }),
-                            onClick: trailingIcon.callback ? trailingIcon.callback : undefined,
-                        })}
+                    {trailingIcon || <></>}
```
eszthoff, line 122: s/a

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -2,68 +2,79 @@ import * as React from 'react';
...
-            helperText,
-            errorMessage,
-            reserveErrorMessageSpace = false,
-            ...rest
-        },
-        ref
-    ) => {
+    ({ context, errorMessage, size, value, ...rest }, ref) => {
         const inputRef = React.useRef<HTMLInputElement | null>(null);
+        const { onChange } = rest;
```
eszthoff, line 17: s/a

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -2,68 +2,79 @@ import * as React from 'react';
...
+        const [hasText, setHasText] = React.useState<boolean>(!!inputRef.current?.value);
+
+        React.useEffect(() => {
+            setHasText(!!value);
+        }, [value]);
+
         const handleOnChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
eszthoff, line 24: s/a

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -2,68 +2,79 @@ import * as React from 'react';
...
+                    'value'
+                )?.set;
+                if (nativeInputValueSetter) {
+                    // Use the setter to update the value
+                    nativeInputValueSetter.call(inputRef.current, '');
+
+                    // Dispatch the 'input' event so React detects the change
+                    const event = new Event('input', { bubbles: true });
+                    inputRef.current.dispatchEvent(event);
+                }
```
eszthoff, line 43: wow. I did learn something today :-D

src/components/Input/SearchInput/SearchInput.tsx
```diff
@@ -2,68 +2,79 @@ import * as React from 'react';
...
+                {...elem('icon', { type: 'leading', bold: hasText, size })}
+                viewBox="0 0 24 24"
+            />
+        );
+        SearchIcon.displayName = 'SearchIcon';
+
         return (
             <Input
+                {...rest}
+                {...block}
```
eszthoff, line 68: Since there is no actual CSS for the top level, you can omit this. 
Note: I noticed this because normally it should be a function that is called, but there is no () at the end :D

eszthoff (changes requested):

src/components/Input/SearchInput/__tests__/SearchInput.spec.tsx
eszthoff: there is a type here. note the `no` before value.

eszthoff (approved):

andrentk:
> for some reason the font-size applied in the input is 13.333px image

Resolved

PR #1278, opened 2024-10-09
Radix Dropdown Menus POC

eszthoff:
[OneUI-496](https://textkernel.atlassian.net/browse/ONEUI-496)


stories/atoms/Menu.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+                                    >
+                                        {value}
+                                    </MultiSelectMenuItem>
+                                ))}
+                            </DropdownMenu.SubContent>
+                        </DropdownMenu.Portal>
+                    </DropdownMenu.Sub>
+                </DropdownMenu.Content>
+            </DropdownMenu.Portal>
+        </DropdownMenu.Root>
```
mukiienko, line 137: So `DropdownMenu` is not responsible for `onChange` of the items, the items have that logic. I like this approach!

mukiienko, line 137: We also can wrap items in another component the selection/navigations still works. Super!

PR #1279, opened 2024-10-09
feat(PrioritySelector): create new component

janseke10:
[ONEUI-489](https://textkernel.atlassian.net/browse/ONEUI-489)

# Checklist

### DIFF
--- a/src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
+++ b/src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
@@ -1,0 +1,111 @@
+import * as React from 'react';
+import Close from '@material-design-icons/svg/round/close.svg';
+import KeyboardDoubleArrowUp from '@material-design-icons/svg/round/keyboard_double_arrow_up.svg';
+import KeyboardArrowUp from '@material-design-icons/svg/round/keyboard_arrow_up.svg';
+import KeyboardArrowDown from '@material-design-icons/svg/round/keyboard_arrow_down.svg';
+import { bem } from '../../../utils/bem';
+import { Dropdown } from '../Dropdown';
+import { ListItem } from '../../List';
+import { Text } from '../../Text';
+import styles from './PrioritySelector.scss';
+import { IconButton } from '../../Buttons';
+
+const iconMap = {
+    mandatory: KeyboardDoubleArrowUp,
+    important: KeyboardArrowUp,
+    optional: KeyboardArrowDown,
+    exclude: Close,
+};
+
+export type Priority = 'mandatory' | 'important' | 'optional' | 'exclude';
+
+export type PriorityItem = {
+    priority: Priority;
+    label: string;
+    value?: string;
+};
+
+const defaultList: PriorityItem[] = [
+    { priority: 'mandatory', label: 'Mandatory' },
+    { priority: 'important', label: 'Important' },
+    { priority: 'optional', label: 'Optional' },
+    { priority: 'exclude', label: 'Exclude' },
+];
+
+export interface Props<>extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange'> {
+    /** Currently selected priority item that indicates the importance of the component. */
+    selectedItem: PriorityItem;
+    /** Array of availible priority items. */
+    list: PriorityItem[];
+    /** Callback function triggered when a new priority is selected. */
+    onChange: (newPriorityItem: PriorityItem) => void;
+    /** Boolean indicating whether the whole badge should be disabled. */
+    isDisabled?: boolean;
+    /** Priority button label name for ARIA labelling */
+    buttonLabel?: string;
+}
+
+const { block, elem } = bem('PrioritySelector', styles);
+export function PrioritySelector({
+    selectedItem,
+    isDisabled = false,
+    onChange,
+    buttonLabel,
+    list = defaultList,
+    ...rest
+}: Props) {
+    const badgeRef = React.useRef<HTMLDivElement | null>(null);
+
+    const hasPriorityList = list && list.length > 0;
+
+    const renderPriorityIcon = (priorityType?: Priority, disabled: boolean = false) => {
+        if (!priorityType) {
+            return null;
+        }
+
+        const IconComponent = iconMap[priorityType];
+
+        return IconComponent ? (
+            <IconComponent
+                {...elem('icon', { [priorityType]: true })}
+                disabled={disabled}
+                viewBox="0 0 24 24"
+            />
+        ) : null;
+    };
+
+    return (
+        <div {...block({ ...rest })} ref={badgeRef}>
+            {hasPriorityList && (
+                <Dropdown<PriorityItem>
+                    button={
+                        <IconButton
+                            {...elem('icon', { [selectedItem.priority]: true })}
+                            variant="ghost"
+                            aria-label={buttonLabel}
+                            disabled={isDisabled}
+                            type="button"
+                        >
+                            {renderPriorityIcon(selectedItem.priority, isDisabled)}
+                        </IconButton>
+                    }
+                    onChange={(newPriorityItem) => onChange(newPriorityItem)}
+                    placement="bottom-start"
+                    listClassName={styles.badgeDropdownList}
+                    refElement={badgeRef}
+                >
+                    {list.map((item) => (
+                        <ListItem className={styles.badgeListItem} key={item.priority} value={item}>
+                            {renderPriorityIcon(item.priority)}
+                            <Text inline size="small">
+                                {item.label}
+                            </Text>
+                        </ListItem>
+                    ))}
+                </Dropdown>
+            )}
+        </div>
+    );
+}
+
+PrioritySelector.displayName = 'PrioritySelector';
--- a/src/components/Dropdown/PrioritySelector/__tests__/PrioritySelector.spec.tsx
+++ b/src/components/Dropdown/PrioritySelector/__tests__/PrioritySelector.spec.tsx
@@ -1,0 +1,10 @@
+describe('PrioritySelector', () => {
+    it.todo('renders correctly with all props provided');
+    it.todo('should handle change correctly');
+    it.todo('should render correctly closed');
+    it.todo('should render correctly opened');
+    it.todo('should access only enabled items with value during keyboard navigation');
+    it.todo('onChange should return passed value');
+    it.todo('should close dropdown when item is chosen');
+    it.todo('should call callback when menu state is changed');
+});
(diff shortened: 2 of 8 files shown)


src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from 'react';
```
eszthoff, line 1: I would place it in its own, as an organism. I think inside other components mean, that they are used by that component, not that they use it. Then we can also change implementation details in the future if needed (I think you saw that design...)

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+const iconMap = {
+    mandatory: KeyboardDoubleArrowUp,
+    important: KeyboardArrowUp,
+    optional: KeyboardArrowDown,
+    exclude: Close,
+};
+
+export type Priority = 'mandatory' | 'important' | 'optional' | 'exclude';
+
+export type PriorityItem = {
```
eszthoff, line 22: can you add docs (comments) for what each means? I think value is the most obscure.

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+    buttonLabel?: string;
+}
+
+const { block, elem } = bem('PrioritySelector', styles);
+export function PrioritySelector({
+    selectedItem,
+    isDisabled = false,
+    onChange,
+    buttonLabel,
+    list = defaultList,
```
eszthoff, line 54: I don't think you should have this. It introduces language dependent elements, which OneUI shouldn't do. You can use it in the story, or tests, but not in the actual implementation.

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const renderPriorityIcon = (priorityType?: Priority, disabled: boolean = false) => {
+        if (!priorityType) {
+            return null;
+        }
+
+        const IconComponent = iconMap[priorityType];
+
+        return IconComponent ? (
+            <IconComponent
+                {...elem('icon', { [priorityType]: true })}
```
eszthoff, line 70: nice :-)

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+                disabled={disabled}
+                viewBox="0 0 24 24"
+            />
+        ) : null;
+    };
+
+    return (
+        <div {...block({ ...rest })} ref={badgeRef}>
```
eszthoff, line 78: 2 things:
* why do you need this extra `div` to wrap the Dropdown?
* will you also spread `rest` on the component here (or on the tigger element), so that other attributes like `style` or `data-x` will persist?

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+            {hasPriorityList && (
+                <Dropdown<PriorityItem>
+                    button={
+                        <IconButton
+                            {...elem('icon', { [selectedItem.priority]: true })}
+                            variant="ghost"
+                            aria-label={buttonLabel}
```
eszthoff, line 85: maybe this can default to `selectedItem.label`?

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+                        >
+                            {renderPriorityIcon(selectedItem.priority, isDisabled)}
+                        </IconButton>
+                    }
+                    onChange={(newPriorityItem) => onChange(newPriorityItem)}
+                    placement="bottom-start"
+                    listClassName={styles.badgeDropdownList}
+                    refElement={badgeRef}
+                >
+                    {list.map((item) => (
```
eszthoff, line 97: I wonder if this can cause a problem if the order is wrong, and if we should control for that or not?
It means looping on Priority, then looking up the correct PriorityItem from the list and rendering that. For that it would be easier if this was an object like thing, something like `{key[Priority]: PriorityItem}` even thought that is a bit of a duplication. What do you think?

BTW: if you do this, you could make `Priority` and Enum, so that you can use it both in JS and TS. (Cannot remember if we have that here in OneUI or only in JF if you want to see examples)

src/components/Dropdown/PrioritySelector/__tests__/PrioritySelector.spec.tsx
```diff
@@ -0,0 +1,10 @@
+describe('PrioritySelector', () => {
+    it.todo('renders correctly with all props provided');
+    it.todo('should handle change correctly');
+    it.todo('should render correctly closed');
+    it.todo('should render correctly opened');
+    it.todo('should access only enabled items with value during keyboard navigation');
```
eszthoff, line 6: I don't think this is relevant. There won't be anything without value, will there?

src/components/Dropdown/PrioritySelector/__tests__/PrioritySelector.spec.tsx
```diff
@@ -0,0 +1,10 @@
...
+    it.todo('onChange should return passed value');
+    it.todo('should close dropdown when item is chosen');
+    it.todo('should call callback when menu state is changed');
```
eszthoff, line 9: might not be needed either

src/components/Dropdown/PrioritySelector/__tests__/PrioritySelector.spec.tsx
```diff
@@ -0,0 +1,10 @@
+describe('PrioritySelector', () => {
+    it.todo('renders correctly with all props provided');
```
eszthoff, line 2: maybe add a test to see that priorities are in the correct order, if you want to control for that (comment above)

src/components/Dropdown/PrioritySelector/index.ts
```diff
@@ -0,0 +1,6 @@
+export {
+    PrioritySelector,
+    Props as PrioritySelectorProps,
+    PriorityItem,
+    Priority,
```
eszthoff, line 5: Should we say Type somewhere? It might be only me, but I'm always confused if things are JS or TS.

stories/molecules/Dropdown.stories.tsx
```diff
@@ -9,6 +9,10 @@ import {
     MultiSelectItem,
 } from '@textkernel/oneui';
 import { HiDotsVertical } from 'react-icons/hi';
+import {
+    PriorityItem,
+    PrioritySelector,
+} from '@textkernel/oneui/components/Dropdown/PrioritySelector/PrioritySelector';
```
eszthoff, line 15: normally you should be able to import things from `@textkenel/oneui` without the full path. now it doesn't work, because you didn't export them from the top `index.ts`

eszthoff (changes requested):

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+    optional: KeyboardArrowDown,
+    exclude: Close,
+};
+
+export type Priority = 'mandatory' | 'important' | 'optional' | 'exclude';
+
+export type PriorityItem = {
+    priority: Priority;
+    label: string;
+    value?: string;
```
AnastasiiaMP, line 25: It would be nice to use the generic type here as it was in `SelectedItemBadge`, because, for example, in Search we don't use string, we use custom type, so it will lead to unnecessary type errors

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+                {...elem('icon', { [priorityType]: true })}
+                disabled={disabled}
+                viewBox="0 0 24 24"
+            />
+        ) : null;
+    };
+
+    return (
+        <div {...block({ ...rest })} ref={badgeRef}>
+            {hasPriorityList && (
```
AnastasiiaMP, line 79: If we’re still planning to use the default list, this check is unnecessary. However, I agree with @eszthoff that we shouldn’t include it here. Additionally, it might be better to maintain single responsibility by leaving the decision to display the selector to the components that use it. For instance, if the priority prop is provided to `SelectedItemBadge`, then we render the `PrioritySelector`. What do you think?

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+const { block, elem } = bem('PrioritySelector', styles);
+export function PrioritySelector({
+    selectedItem,
+    isDisabled = false,
+    onChange,
+    buttonLabel,
+    list = defaultList,
+    ...rest
+}: Props) {
+    const badgeRef = React.useRef<HTMLDivElement | null>(null);
```
AnastasiiaMP, line 57: `badgeRef` should come from props since, for example, we may need to align the dropdown width with the entire `SelectedPriorityBadge`, and we don’t need any wrapper in this component. It should also be optional, so if specific width alignment isn’t needed, it will use the default behavior specified in the `Dropdown` component.

AnastasiiaMP (changes requested):

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
janseke10: In the refinement session we decided that it was going in the dropdown component's folder... I do agree though.
Shall i change it to it's own folder, and change the ticket?

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
janseke10: credits should go to @AnastasiiaMP  :) I copied it from her I believe!

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
janseke10: Yes, good point. I'll remove this check. When we leave the decision to display it to the parent component, we don't need this check, as it will be required anyway. Thanks!

src/components/Dropdown/PrioritySelector/PrioritySelector.tsx
janseke10: omg. This is what I was talking about this morning! Oops! ;)

PR #1288, opened 2024-10-15
feat(Dialog): update dialog design specs

AlsasaAmmar:
[ONEUI-448](https://textkernel.atlassian.net/browse/ONEUI-448)

BREAKING CHANGE: Component Props API change. Introduction of closeButton. If title defined, closeButton should be defined as well.


src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
+    } = props;
 
+    const { ref, height } = useResizeObserver<HTMLDivElement>();
+    const showButtons = !!acceptButton;
     const isConfirm = !!cancelButton;
 
     return (
-        <Modal {...rest} {...block(props)} isOpen={isOpen}>
-            <div {...elem('content', props)} role="alert">
+        <Modal {...rest} {...block('dialog')} isOpen={isOpen}>
```
eszthoff, line 52: see https://textkernelteam.slack.com/archives/C9W305EH3/p1729606839227849

eszthoff: also I think in general you are using block wrongly, we never pass a string to it. The main classname gets refined in the original call to bem (see line 33 here)

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
+            <div {...elem('header')} role="alert">
                 {!!title && (
                     <Heading level="h2" {...elem('title', props)}>
                         {title}
                     </Heading>
                 )}
+                <MdClose onClick={onClose} data-testid="close-icon" />
```
eszthoff, line 59: I thought this will only be displayed when there is a header section, which will only be there if there is a title... Or do I see the design wrongly?

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
-                {isConfirm && (
+            {showButtons && (
+                <div {...elem('actions', { ...props, borderTop: (height ?? 0) === 528 })}>
                     <Button
-                        {...elem('cancel', props)}
-                        onClick={cancelButton.onClick}
-                        variant="ghost"
+                        onClick={acceptButton.onClick}
+                        context={variant === 'primary' ? 'primary' : 'critical'}
+                        aria-label="Confirm OK"
```
eszthoff, line 69: see https://textkernelteam.slack.com/archives/C9W305EH3/p1729606839227849

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
                     >
-                        {cancelButton.label}
+                        {acceptButton.label}
                     </Button>
-                )}
-                <Button onClick={acceptButton.onClick} context="primary">
-                    {acceptButton.label}
-                </Button>
-            </div>
+                    {isConfirm && (
```
eszthoff, line 73: I'm not sure how this prop makes sense in the current context. Shouldn't the condition be if there is a cancel button present or not?

eszthoff (changes requested):

src/components/Dialogs/Dialog/Dialog.tsx
AlsasaAmmar: I kept it because one of the cases is to use the dialog without any buttons which means that user won't have any close/cancel button on the dialog.

src/components/Dialogs/Dialog/Dialog.tsx
AlsasaAmmar: isConfirm is not a prop. It's calculated from cancelButton (line 49). It was like this so I kept it..

src/components/Dialogs/Dialog/Dialog.scss
```diff
@@ -1,21 +1,35 @@
 .Dialog {
     width: 400px;
+    border-radius: 12px;
```
andrentk, line 3: Use variable for border radius

src/components/Dialogs/Dialog/Dialog.tsx
andrentk: Looking at figma, every variant has either Title + Close or Buttons. The whole header should be conditionally rendered based on the title being defined or not.

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
+        title,
+        acceptButton,
+        cancelButton,
+        variant = 'primary',
+        onClose,
+        ...rest
+    } = props;
 
+    const { ref, height } = useResizeObserver<HTMLDivElement>();
+    const showButtons = !!acceptButton;
```
andrentk, line 48: Looking at dialog figma variations the showButtons should consider both. There is no variation with only "confirm" button. Check with designers, if it is still needed for Alert.

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
                 )}
+                <MdClose onClick={onClose} data-testid="close-icon" />
+            </div>
+            <div ref={ref} {...elem('content')}>
                 {children}
             </div>
-            <div {...elem('actions', props)}>
-                {isConfirm && (
+            {showButtons && (
+                <div {...elem('actions', { ...props, borderTop: (height ?? 0) === 528 })}>
```
andrentk, line 65: can't we just do height === 528?

src/components/Dialogs/Dialog/Dialog.tsx
andrentk: Please confirm with design team first. Looking at the variations both buttons should always be rendered together, no need for isConfirm. If it is to be rendered together, the props needs to be changed and a single object to contain both button props.

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
+import { render, RenderResult, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+import { Dialog } from '../Dialog';
+
+describe('Dialogs', () => {
```
andrentk, line 7:
```suggestion
describe('Dialog', () => {
```

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+            >
+                {dialogBody}
+            </Dialog>
+        );
+    });
+
+    it('should render correctly with a title', () => {
+        expect(view.baseElement).toMatchSnapshot();
+        expect(screen.getByText(/Title/i)).toBeInTheDocument();
+        expect(screen.queryByRole('heading', { level: 2 })).toBeInTheDocument();
```
andrentk, line 47: if the element is to be present you can use get instead of query.

src/components/Dialogs/Dialog/Dialog.tsx
andrentk: This icon should be inside of a button otherwise the click area is very small and there is no indication that it is a button visually or with screen readers.
Also if it is a button you don't need to use data-testid to find it in the dom.

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+        expect(screen.getByTestId('close-icon')).toBeInTheDocument();
```
andrentk, line 48: with the change mentioned above you can get the element by the role.

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+                onClose={mockOnCancel}
+                ariaHideApp={false}
+                contentLabel="Content Label"
+                acceptButton={acceptButtonProps}
+                cancelButton={cancelButtonProps}
+            >
+                {dialogBody}
+            </Dialog>
+        );
+        expect(screen.getAllByRole('button')[0]).toHaveClass('Button--context_primary');
```
andrentk, line 119: can't we get the single button instead of querying by all and use the first one?

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+    });
+
+    it('should call onAccept callback when button is clicked', async () => {
+        const user = userEvent.setup();
+        await user.click(screen.getAllByRole('button')[0]);
```
andrentk, line 124: same as above

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+        expect(mockOnAccept).toHaveBeenCalledTimes(1);
+    });
+
+    it('should call onCancel callback when button is clicked', async () => {
+        const user = userEvent.setup();
+        await user.click(screen.getAllByRole('button')[1]);
```
andrentk, line 131: same as above

src/components/Dialogs/Dialog/__tests__/Dialog.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+        expect(mockOnCancel).toHaveBeenCalledTimes(1);
+    });
+
+    it('should call onClose callback when x is clicked', async () => {
+        const user = userEvent.setup();
+        await user.click(screen.getByTestId('close-icon'));
```
andrentk, line 138: with the change mentioned above you can get the element by the role.

src/components/Dialogs/Dialog/Dialog.scss
```diff
@@ -1,21 +1,35 @@
...
+        overflow-y: auto;
+        white-space: normal;
     }
 
     &__actions {
         display: flex;
-        justify-content: flex-end;
+        justify-content: flex-start;
         align-content: baseline;
         margin-top: var(--space-75);
```
andrentk, line 23: not sure about this margin-top. The text is disappearing before the line because of the margin.
image
 Please check with the designers.

src/components/Dialogs/Dialog/Dialog.tsx
andrentk: This introduces a breaking change, please mark the commit as breaking change

stories/organisms/Dialogs.stories.tsx
```diff
@@ -93,3 +95,56 @@ export const _Confirm: ConfirmStory = {
...
+        return (
+            <>
+                <Button onClick={handleButtonClick}>Click me to see a confirm dialog</Button>
+                <Dialog
+                    {...args}
+                    isOpen={showAlert}
+                    onClose={handleCancel}
+                    acceptButton={{
+                        onClick: handleAccept,
+                        label: 'Delete',
```
andrentk, line 140: Maybe better to change the variant to critical to be consistent with Delete action

src/components/Dialogs/Alert/__tests__/Alert.spec.tsx
```diff
@@ -6,13 +6,23 @@ import { Alert } from '../Alert';
 
 describe('Alert', () => {
     const mockOnAccept = jest.fn();
+    const mockOnClose = jest.fn();
     let view: RenderResult;
 
     beforeEach(() => {
+        global.ResizeObserver = class MockedResizeObserver {
```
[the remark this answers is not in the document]
eszthoff: thanks for moving the code. I see however that you import the mock into the test file. If the mock done as per the docs attached before, this is not needed, as jest will always use the "module" from the top level `__mocks__` folder

src/components/Dialogs/Confirm/Confirm.tsx
```diff
@@ -1,11 +0,0 @@
-import * as React from 'react';
```
eszthoff, line 1: why are we removing this component?

src/components/Dialogs/Dialog/Dialog.tsx
eszthoff: Even if the prop was named this way before, I don't see the reason to keep this naming if it doesn't make sense in the current context anymore. This will make the code confusing and hard to understand in the future. In general, I'd rather it gets renamed, especially if this is only internal to this file, it is very easy to do. 
However looking at the code as it looks now, it doesn't even seem to make a lot of sense to me to have this as a variable. Maybe it is easier to just use `!!cancelButton` in place, same as you do with the `!!title` above.

eszthoff (changes requested):

src/components/Dialogs/Confirm/Confirm.tsx
AlsasaAmmar: I removed this component because there is no use case for it and in the new design it is not there. 
Maybe we need to discuss this in the next sync meeting

andrentk:
@AlsasaAmmar you have conflicts in this PR

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
+                        <MdClose />
+                    </button>
+                </div>
+            )}
+            <div ref={ref} {...elem('content')}>
                 {children}
             </div>
-            <div {...elem('actions', props)}>
-                {isConfirm && (
+            {showButtons && (
```
andrentk, line 65: there is no padding between the buttons

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,66 @@ export interface DialogProps extends ModalProps {
...
+    const { ref, height } = useResizeObserver<HTMLDivElement>();
+    const showButtons = !!acceptButton;
 
     return (
         <Modal {...rest} {...block(props)} isOpen={isOpen}>
-            <div {...elem('content', props)} role="alert">
-                {!!title && (
+            {!!title && (
+                <div {...elem('header')} role="alert">
                     <Heading level="h2" {...elem('title', props)}>
```
andrentk, line 54: title text dimensions are not correct

andrentk (commented):
The point related to size was not addressed. https://zeroheight.com/9e76adf4e/p/499dca-modal-dialog/t/a70439c874

src/components/Dialogs/Dialog/Dialog.tsx
andrentk: This is not resolved. Looking at the vercel deployment, there is still no padding between the buttons.

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -18,42 +21,62 @@ export interface DialogProps extends ModalProps {
...
     title?: string;
     /** Properties of the accept button */
-    acceptButton: ButtonProps;
+    acceptButton?: ButtonProps;
     /** Properties of the cancel button */
     cancelButton?: ButtonProps;
+    /** the context of the buttons in the footer */
+    variant?: Omit<ButtonContext, 'secondary'>;
+    /** closes the dialog */
+    onClose: () => void;
```
andrentk, line 30: I think we discussed this before. `onClose` is only needed when there is a title, am I right? If yes, you can require the `onClose` to be always required only when `title` is present. You can take a look at the example of the `Input`, that if you define `context = 'critical'`, the `errorMessage` is required.

andrentk (changes requested):

src/components/Dialogs/Dialog/Dialog.tsx
AlsasaAmmar: I had padding between the buttons before the last push. I had added a class on the cancel button and add padding. However, after the first comment I thought that there shouldn't be padding, so I removed it. Should I add padding by another way or what is needed here?

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -1,58 +1,82 @@
...
+            <div {...elem('container')}>
                 {!!title && (
-                    <Heading level="h2" {...elem('title', props)}>
-                        {title}
-                    </Heading>
+                    <div {...elem('header')} role="alert">
+                        <Heading level="h3" {...elem('title', props)}>
+                            {title}
+                        </Heading>
+                        <button {...elem('closeButton')} type="button" onClick={onClose}>
```
eszthoff, line 57: doesn't this need an aria-label?

stories/organisms/Dialog.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+    subcomponents: { Alert, Confirm } as any,
+};
+
+export default meta;
+
+type DialogStory = StoryObj<typeof Dialog>;
+
+export const _Dialog: DialogStory = {
+    name: 'Dialog',
+    args: {
```
eszthoff, line 18: In storybook, it is hard to set the variant. It asks for some JSON string instead of offering some radio buttons. Can you check?

eszthoff (changes requested):

stories/organisms/Dialog.stories.tsx
andrentk: I was able to make it to string. It doesn't work with radio buttons because the type is Omit<...> and not ButtonContext directly.

src/components/Dialogs/Dialog/Dialog.tsx
andrentk: done

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -13,7 +13,14 @@ export type ButtonProps = {
     label: string;
 };
 
-export interface BaseDialogProps extends ModalProps {
+export type HeaderProps = {
+    /** Dialog header title */
+    title: string | undefined;
```
eszthoff, line 18: Can it be `undefined` if there is a header?

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -26,20 +33,19 @@ export interface BaseDialogProps extends ModalProps {
...
+export const Dialog: React.FC<Props> = (props) => {
     const {
         isOpen,
         children,
         title,
         acceptButton,
         cancelButton,
+        closeButton,
         variant = 'primary',
-        onClose,
```
eszthoff: I guess it is a breaking change, no? Let's make sure to document it.

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -48,36 +54,39 @@ export const Dialog: React.FC<DialogProps> = (props) => {
...
             </div>
+            {showButtons && (
+                <div {...elem('actions', { ...props, borderTop: height === 412 })}>
+                    <Button
+                        onClick={acceptButton.onClick}
+                        context={variant === 'primary' ? 'primary' : 'critical'}
+                    >
+                        {acceptButton.label}
+                    </Button>
+                    {!!cancelButton && (
```
eszthoff, line 83:
```suggestion
                    {!!cancelButton.onClick && (
```
I think it is better to check if the function is there, otherwise an empty object is also "true" in JS

eszthoff (changes requested):

eszthoff:
Screenshot 2024-11-22 at 11 11 39
Shouldn't the X be higher?

src/components/Dialogs/Dialog/Dialog.tsx
```diff
@@ -1,59 +1,92 @@
...
     onClick: () => void;
     label: string;
 };
 
-export interface DialogProps extends ModalProps {
+export type HeaderProps = {
+    /** Dialog header title */
+    title: string | undefined;
+    /** Properties of the close button */
+    closeButton: ButtonProps;
```
eszthoff, line 20: We also have a prop from Modal Props called `onRequestClose`. Maybe we should omit it from the dialog, and instead pass this closeButton.onClick to it? What do you think?

src/components/Dialogs/Dialog/Dialog.tsx
eszthoff: I see Ammar already wrote something, but it is very vague in my opinion and hard to execute when updating the application. It would be nice to write things like `onRequestClose` is replaced by `closeButton:... ` or whatever makes sense. So more like a checklist on how to migrate. What do you think?

eszthoff (approved):

andrentk (approved):

PR #1302, opened 2024-10-29
refactor(Tag): migrate Tag component

janseke10:
[ONEUI-454](https://textkernel.atlassian.net/browse/ONEUI-454)

# Checklist

### DIFF
--- a/src/components/Tag/Tag.scss
+++ b/src/components/Tag/Tag.scss
@@ -2,37 +2,69 @@
     display: flex;
     align-items: center;
 
-    padding: var(--space-50) var(--space-75);
-    border-radius: var(--border-radius);
+    padding: var(--space-25) var(--space-50);
+    gap: 4px;
+    border-radius: 4px;
+    border: 1px solid var(--color-border-subtle);
+    cursor: default;
+    background-color: var(--bg-color, white); // Fallback to --color-background if bgColor isn't provided
+
+    &:focus:not([disabled]) {
+        outline: none;
+        box-shadow: 0 0 0 2px var(--color-info-20);
+    }
+
+    &:hover:not([disabled]){
+        background-color: var(--color-background-neutral-subtlest-hover);
+    }
+
+    &[disabled] {
+        color: var(--color-text-disabled);
+    }
 
     &--isSelected {
         box-shadow: 0 0 0 2px var(--color-brand-50);
     }
 
-    &--clickable {
+    &--clickable:not([disabled]) {
         cursor: pointer;
     }
 
     &__text{
+        composes: OneUI-label-text from global;
         margin: 0;
 
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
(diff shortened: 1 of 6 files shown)


src/components/Tag/Tag.scss
```diff
@@ -2,37 +2,69 @@
     display: flex;
     align-items: center;
 
-    padding: var(--space-50) var(--space-75);
-    border-radius: var(--border-radius);
+    padding: var(--space-25) var(--space-50);
+    gap: 4px;
+    border-radius: 4px;
```
eszthoff, line 7: please use space variables instead of fixed values (check other places in the file)

src/components/Tag/Tag.tsx
```diff
@@ -24,6 +24,8 @@ export interface Props extends React.HTMLAttributes<HTMLDivElement> {
     contentStyle?: React.CSSProperties;
```
eszthoff, line 24: I see a few props (cannot comment on lines above) that aren't supported now in the new design: isSelected, maxWidth, size. maybe we don't need both onDelete and onClick?

I checked the usage in JD, we use onClick, not onDelete so that when the whole tag is clicked it still deletes the label. We also use `size` but we can get away without it I think.

Could you check current usage with the Search team on their end?
Maybe we can delete these props, or maybe we should keep them (we have a design for selectable tags, but it was outscoped in this ticket). Maybe check with Dima too, and when you have all the info, we can make a decision together.

Also, it would be good to adjust the stories so that they reflect the current use cases. It would be good to have one story for `read-only` tag, so without an onClick/onDelete callback, one with the callBack and one for coloured version.

eszthoff (changes requested):

src/components/Tag/Tag.scss
janseke10: mehhhh but the standard radius is 3px :( But i'll go for var(--space-50) then. It just feels less logical than a short '4px'...

src/components/Tag/Tag.tsx
janseke10: I looked at JD and in LabelTag we use both onClick and onDelete. We also use MaxWidth, and size indeed. I don't know if we need to avoid breaking changes ;)  
Tag is not used in Search at all, apparently! So that makes it a bit easier...

eszthoff: ok, let's keep the props as they are then

eszthoff:
I found a bug that possibly was there already. If you use keyboard navigation, you cannot TAB beyond the delete button. The focus is stuck there. I think it is because of the `preventDefault` in line 71 or so.

eszthoff (approved):

eszthoff:
LGTM but please add BREAKING CHANGES to the description and make sure it is copied to the squashed commit when merging. 
Andre already changed the checklist for new PRs, so it will be clearer in the future. The checkboxes that aren't relevant can be removed here too :-)

PR #1303, opened 2024-11-01
feat(Pill): add new PillButton variant

eszthoff:
Story: [ONEUI-438](https://textkernel.atlassian.net/browse/ONEUI-438)


src/components/Pill/PillButtonEnhanced/PillButtonEnhanced.tsx
```diff
@@ -0,0 +1,89 @@
...
+import styles from './PillButtonEnhanced.scss';
+
+export interface Props extends PillButtonBaseProps {
+    /** Label/indicator in case of multiple selection, such as `+2` for instance */
+    multiSelectionLabel?: string;
+    /** Tooltip content to be shown when multiple selections are shown */
+    multiSelectionTooltip?: string;
+    /** Props needed for rendering optional priority selector */
+    prioritySelector?: boolean; // boolean is temporary placeholder until PrioritySelector component is finalized
+}
```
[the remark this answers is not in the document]
eszthoff: - `isOpen, toggleDropdown, name, content`:  I kept the naming of all the "base" props from the other pill button, so that it is easy to migrate, and we don't introduce any breaking changes. so now all you have to do is add `variant: 'enhanced'` to an existing pill and it will work. This is the approach we agreed on during refinement. Even thought I agree that some of the names could be revised, I do feel that having less breaking changes is also good. An intermediate step could be to have these things named differently only for the enhanced button and do some mapping inside the Pill component, but not sure of the benefits for that. In general the situation is that we don't use the PillButton directly in the components, and probably want to keep it that way. 
- I like the `addionalContent` as a name better then the `multiSelection` I used. Thanks for the idea. Shall it be then `additionalContentLabel` and `additionalContentTooltip`? I tested and for new lines you can use `text 1\nLine 2` so no need to a ReactNode there.
- For the truncation tooltip, I don't think we need a prop. It will be handled by the `title` attribute and will always show `content` (the truncated section). Does that work for you? Note: this is also a reason that I think it is better not to have the content as an array, it gives more flexibility for the application to manage the logic of what goes where.
- Sorry for the width thing, I thought I pushed it. I would add a `maxWidth?: string` prop. If it is not defined we use the default, if it is, then we can overwrite it on the style level. I think that will give cover all the use cases. What do you think?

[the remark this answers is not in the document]
eszthoff: @mukiienko I keep thinking about it, and I don't really like the idea of using a tooltip for the content itself. It is true that it is showing up faster that way, but there are 2 drawbacks:
1. it will show up also if the content is not truncated (unless we somehow add code in JS instead of just CSS which seems to me an awful lot of complication)
2. the content will be there twice not only visually but also for screen readers.
What say you?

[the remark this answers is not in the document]
mukiienko: `contentTooltip` will be optional. All that you mentioned also applies to the `title` attribute, doesn’t it?
1) the `title` is shown always, event the text are not the truncated, right?
2) we need duplicate text for `title` as well. For readers we can use appropriate `aria` attribute.
If you think more elegant way to implement via `title` that I don't mind. So, it will be `content={<span title="some text">Some text</span>}`, right?
@eszthoff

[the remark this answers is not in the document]
eszthoff: @mukiienko 
1. yes, but since it has a delay, it is not that evident. You need to want to see it for it show up.
2. I think screen readers are better equipped to understand `title` in this scenario, but I haven't tested it indeed.

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unnecessary-type-constraint */
 import React from 'react';
```
mukiienko, line 1: Priority is not aligned well:
Screenshot 2024-11-08 at 11 25 55

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
```diff
@@ -1,4 +1,3 @@
...
 import {
```
mukiienko, line 2: It looks like maxWidth doesn't work. Label is not properly truncated:
Screenshot 2024-11-08 at 11 30 47

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
```diff
@@ -25,6 +24,7 @@ export interface Props<PriorityItemValue> extends DropdownMenuCheckboxItemProps
 const { block } = bem('MultiSelectItem', styles);
```
mukiienko, line 24: It would be nice to have story with new pill and content dropdown to see how it works together.

stories/molecules/Pill.stories.tsx
```diff
@@ -61,6 +67,91 @@ export const _PillButton: PillButtonStory = {
     ),
 };
 
+type PillButtonEnhancedStory = StoryObj<typeof PillButtonEnhanced>;
+
+export const _PillButtonEnhanced: PillButtonEnhancedStory = {
+    name: 'PillButtonEnhanced',
+    args: {
+        name: 'Pill name',
+        content: 'value',
```
mukiienko, line 76: It's strange, but why the content is JSON string:
Screenshot 2024-11-08 at 11 42 08

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: there is a story. you just have to choose the variant in the controls. Or is there something else?

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: I cannot reproduce it. Maybe you need to hard refresh your browser?
Screenshot 2024-11-08 at 11 56 41

mukiienko: try to change the pill name

eszthoff: sorry... I see that you made the pill name long. I don't have a design for that. The design said that it is only the value that should be truncated. 
If you have a super long name, you can overwrite the `maxWidth` prop and make the pill bigger.

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: yes, I also mean that :D 
Screenshot 2024-11-08 at 12 07 59

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
mukiienko: I think it's not above the design. We have to handle it properly. Maybe,  at least have some maxWidth in percentages.

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
mukiienko: hmm, it's strange. Maybe something wrong on my machine. I have the error:
Screenshot 2024-11-08 at 12 20 29
@AnastasiiaMP could you please check if it works for you?

stories/molecules/Pill.stories.tsx
eszthoff: I think because this prop accepts any ReactNode. This is the definition taken from the current PillButton.

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
```diff
@@ -25,6 +24,7 @@ export interface Props<PriorityItemValue> extends DropdownMenuCheckboxItemProps
...
 
 export const MultiSelectItem = React.forwardRef(
+    /* eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint */
     <PriorityItemValue extends unknown>(
```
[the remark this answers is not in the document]
eszthoff: where do you have these colors? In Figma that is linked to the ticket they are grey.
https://www.figma.com/design/TlGOOkoCGSyDtrnMwmjnAl/OneUI-Components-v2?node-id=2649-6726&node-type=canvas&t=ZUTjsP5MwXspO6NW-0

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: I think that is up to Dima, as it is a new requirement. Can you ping him in the ticket with the question?

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
AnastasiiaMP: Pill has dropdown only in one story - with the name `Pill`, in all other stories this error is happening to me as well

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
AnastasiiaMP: The clear Icon also doesn't seem to be perfectly align, but it is really minor thing. What is noticeable is that all the icons are too small, especially priority icon which is 12x12 ox and in design 20x20. The arrow one has onle 2 px difference
Screenshot 2024-11-08 at 13 34 11
Screenshot 2024-11-08 at 13 31 44
-attachments/assets/40b6f9ba-9664-4c2f-ad75-18c87944a573">
Screenshot 2024-11-08 at 13 31 32

src/components/Pill/PillButtonEnhanced/PillButtonEnhanced.tsx
```diff
@@ -0,0 +1,140 @@
...
+        if (content) {
+            buttonIcon = (
+                <Clear aria-label={clearLabel} viewBox="0 0 24 24" height="14px" width="14px" />
+            );
+            handleButtonClick = (e) => {
+                e.stopPropagation();
+                e.preventDefault();
+                onClear();
+            };
+        }
```
AnastasiiaMP, line 70: it's not a necessary change, just a suggestion 
It is a little bit confusing, maybe it is just me 😃 Maybe some kind of helper function which would be reposinsible for the button would make it cleaner:
Something in this direction:
```
  const getButtonIconAndHandler = () => {
            if (content) {
                return {
                    icon: (
                        <Clear
                            aria-label={clearLabel}
                            viewBox="0 0 24 24"
                            height="14px"
                            width="14px"
                        />
                    ),
                    onClick: (e: React.MouseEvent) => {
                        e.stopPropagation();
                        e.preventDefault();
                        onClear();
                    },
                };
            }
            if (isOpen) {
                return {
                    icon: (
                        <UpArrow
                            aria-label={upArrowLabel}
                            viewBox="0 0 24 24"
                            height="14px"
                            width="14px"
                        />
                    ),
                    onClick: (e: React.MouseEvent) => {
                        e.stopPropagation();
                        e.preventDefault();
                        toggleDropdown();
                    },
                };
            }
            return {
                icon: (
                    <DownArrow
                        aria-label={downArrowLabel}
                        viewBox="0 0 24 24"
                        height="14px"
                        width="14px"
                    />
                ),
                onClick: (e) => {
                    e.stopPropagation();
                    e.preventDefault();
                    toggleDropdown();
                },
            };
        };

        const { icon: buttonIcon, onClick: handleButtonClick } = getButtonIconAndHandler();
```
You can even further simplify it. But it is up to you, just a suggestion :-)

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
AnastasiiaMP: the Clear one also should be 16x16
 ` <Clear aria-label={clearLabel} viewBox="0 0 24 24" height="14px" width="14px" />`

eszthoff: The Prioirty button icon is coming from IconButton. I used the small, since the normal size is bigger then the pill button all together. Not sure where to go with it. Dima needs to align his designs :-D 
I'll change the rest

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: added the missing argument to the stories

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: Screenshot 2024-11-08 at 15 26 01

after making the icons 16px, I think the alignment is now ok, no?

mukiienko: I think it should be some space between to button and text:
Screenshot 2024-11-08 at 15 29 39

AnastasiiaMP: yeah, with 16px, looks great. And agree with @mukiienko , some space would be nice 😊

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
mukiienko: ok

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: Can you try now, I think you were looking before I pushed the changes.

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: I will ask Dima to make a design proper with the priority icon, using the actual button icon. I didn't add the space, because... whatever, I'm not the designer, and I don't like to make up things :-D

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
mukiienko: I don't have the errors now, but I still don't see the dropdown. Could you please check @AnastasiiaMP ?

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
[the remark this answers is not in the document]
eszthoff: 🤯

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
eszthoff: @mukiienko are you looking at this story? https://oneui-e9xydbbbf-textkernel-oneuis-projects.vercel.app/?path=/story/molecules-pill--pill&args=variant:enhanced

PR #1308, opened 2024-11-05
feat(PrioritySelector): new component

janseke10:
[ONEUI-489](https://textkernel.atlassian.net/browse/ONEUI-489)

### Checklist

-   [ ] The Pull Request title follows our guidelines eg. `feat(PrioritySelector): new component`
-   [ ] Breaking changes are mentioned eg. `BREAKING CHANGE: isMultiSelect prop is removed`, if any.
-   [ ] Squash and Merge option is selected as merge option.

### DIFF
--- a/src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
+++ b/src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
@@ -8,14 +8,13 @@ import { bem } from '../../../../utils';
 import { VisualCheckbox } from '../../../Checkbox';
 import { Text } from '../../../Text';
 import styles from './MultiSelectItem.scss';
+import { stopPropagation } from '../../../../utils/misc';
 
 export interface Props<PriorityItemValue> extends DropdownMenuCheckboxItemProps {
     /** Id for the checkbox */
     id?: string;
     /** a variant to determine the look and feel of component */
     variant?: 'option' | 'select-all' | 'group-title';
-    /** to add a priority badge before the label */
-    hasPriority?: boolean;
     /** props for PrioritySelector  */
     priority?: PrioritySelectorProps<PriorityItemValue>;
     /** Checkbox status */
@@ -31,7 +30,6 @@ export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props<string>>(
             isSelected = false,
             disabled = false,
             variant = 'option',
-            hasPriority = false,
             onCheckedChange,
             priority,
             ...rest
@@ -40,16 +38,11 @@ export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props<string>>(
     ) => {
         const priorityRef = React.useRef<HTMLDivElement>(null);
 
-        const hasPriorityList = hasPriority && priority && priority.list.length > 0;
-
-        const stopPropagation = (e: React.MouseEvent | React.KeyboardEvent) => {
-            e.stopPropagation(); // Prevents the checkbox from toggling
-        };
+        const hasPriorityList = priority && priority.list.length > 0;
 
         const handleKeyDown = (e: React.KeyboardEvent) => {
-            if (e.key === 'Tab') {
+            if (e.key === 'Tab' && !e.shiftKey) {
                 if (priorityRef.current) {
-                    e.preventDefault();
                     priorityRef.current.focus();
                 }
             }
@@ -77,12 +70,7 @@ export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props<string>>(
                 <VisualCheckbox checked={isSelected} />
                 {hasPriorityList && (
                     <div role="none" onClick={stopPropagation} onKeyDown={stopPropagation}>
-                        <PrioritySelector
-                            list={priority.list}
-                            selectedItem={priority.selectedItem}
-                            onSelect={priority.onSelect}
-                            buttonRef={priorityRef}
-                        />
+                        <PrioritySelector {...priority} buttonRef={priorityRef} />
                     </div>
                 )}
                 <Text inline>{children}</Text>
--- a/src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
+++ b/src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
@@ -35,7 +35,7 @@ export const MultiSelectItem = React.forwardRef(
             priority,
             ...rest
         }: Props<PriorityItemValue>,
-        ref: React.Ref<HTMLLIElement>
+        ref: React.Ref<HTMLDivElement>
     ) => {
         const priorityRef = React.useRef<HTMLDivElement>(null);
 
@@ -79,5 +79,5 @@ export const MultiSelectItem = React.forwardRef(
         );
     }
 ) as <PriorityItemValue extends unknown>(
-    p: Props<PriorityItemValue> & { ref?: React.Ref<HTMLElement> }
+    p: Props<PriorityItemValue> & { ref?: React.Ref<HTMLDivElement> }
 ) => React.ReactElement;
--- a/src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
+++ b/src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
@@ -3,57 +3,77 @@ import {
     DropdownMenuCheckboxItem,
     DropdownMenuCheckboxItemProps,
 } from '@radix-ui/react-dropdown-menu';
+import { PrioritySelector, PrioritySelectorProps } from '../../../PrioritySelector';
 import { bem } from '../../../../utils';
 import { VisualCheckbox } from '../../../Checkbox';
 import { Text } from '../../../Text';
 import styles from './MultiSelectItem.scss';
 
-export interface Props extends DropdownMenuCheckboxItemProps {
+export interface Props<PriorityItemValue> extends DropdownMenuCheckboxItemProps {
     /** Id for the checkbox */
     id?: string;
     /** a variant to determine the look and feel of component */
     variant?: 'option' | 'select-all' | 'group-title';
     /** to add a priority badge before the label */
     hasPriority?: boolean;
+    /** props for PrioritySelector  */
+    priority?: PrioritySelectorProps<PriorityItemValue>;
     /** Checkbox status */
     isSelected?: boolean;
 }
 
 const { block } = bem('MultiSelectItem', styles);
 
-export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props>(
+export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props<string>>(
     (
         {
             children,
             isSelected = false,
             disabled = false,
             variant = 'option',
-            hasPriority = false, // eslint-disable-line @typescript-eslint/no-unused-vars
+            hasPriority = false,
             onCheckedChange,
+            priority,
             ...rest
         },
         ref
-    ) => (
-        <DropdownMenuCheckboxItem
-            ref={ref}
-            role="option"
-            aria-selected={isSelected}
-            tabIndex={disabled ? -1 : 0}
-            checked={isSelected}
-            onSelect={(e) => {
-                e.preventDefault();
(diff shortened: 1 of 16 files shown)


src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
```diff
@@ -3,57 +3,90 @@ import {
...
 import styles from './MultiSelectItem.scss';
 
-export interface Props extends DropdownMenuCheckboxItemProps {
+export interface Props<PriorityItemValue> extends DropdownMenuCheckboxItemProps {
     /** Id for the checkbox */
     id?: string;
     /** a variant to determine the look and feel of component */
     variant?: 'option' | 'select-all' | 'group-title';
     /** to add a priority badge before the label */
     hasPriority?: boolean;
```
eszthoff, line 18: I wonder if we need this prop. I think it was a placholder only. Maybe we can just assume that if `priority` is defined, then it should be rendered. What do you think?

src/components/Dropdown/Items/SingleSelectItem/SingleSelectItem.tsx
```diff
@@ -1,45 +1,71 @@
...
     /** A function to be called if the item is clicked */
     onSelect?: (e: React.SyntheticEvent<HTMLDivElement>) => void;
     /** Id for the checkbox */
     id?: string;
-    /** to add a priority badge before the label */
-    hasPriority?: boolean;
     /** Checkbox status */
     isSelected?: boolean;
+    /** Should PrioritySelector be included */
+    hasPriority?: boolean;
```
eszthoff, line 15: s/a

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+    /** Array of availible priority items. */
+    list: PriorityItemType<PriorityItemValue>[];
+    /** Callback function triggered when a new priority is selected. */
+    onSelect: (newPriorityItem: PriorityItemType<PriorityItemValue>) => void;
+    /** Boolean indicating whether the whole badge should be disabled. */
+    isDisabled?: boolean;
+    /** Priority button label name for ARIA labelling */
+    buttonLabel?: string;
+    /** ref to possible parent div, if it is a badge */
+    badgeRef?: React.RefObject<HTMLElement | null> | React.ForwardedRef<HTMLDivElement>;
```
eszthoff, line 51: I'm not sure what this prop about. It is about aligning the dropdown differently? Why do we call it `badge`? Maybe it should be `parentRef` and the explanation can be more detailed...

src/components/Dropdown/Items/SingleSelectItem/SingleSelectItem.tsx
```diff
@@ -1,45 +1,71 @@
...
+                    disabled,
+                    ...rest,
+                })}
+            >
+                {hasPriorityList && (
+                    <div role="none" onClick={stopPropagation} onKeyDown={stopPropagation}>
+                        <PrioritySelector
+                            list={priority.list}
+                            selectedItem={priority.selectedItem}
+                            onSelect={priority.onSelect}
```
eszthoff, line 62: shouldn't other props e.g. `isDisabled` and so on be passed here too? Maybe easiest to do `<PrioritySelector {...priority} buttonRef={priorityRef} />`

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+    isDisabled = false,
+    buttonLabel,
+    list,
+    badgeRef,
+    buttonRef,
+    ...rest
+}: Props<PriorityItemValue>) {
+    const renderPriorityIcon = (priorityType?: Priority, disabled: boolean = false) => {
+        if (!priorityType) {
+            return <></>;
```
eszthoff, line 70: normally we return `null` in these cases.

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+        }
+        const IconComponent = iconMap[priorityType];
+        return IconComponent ? (
+            <IconComponent
+                {...elem('icon', { [priorityType]: true })}
+                disabled={disabled}
+                viewBox="0 0 24 24"
+            />
+        ) : (
+            <></>
```
eszthoff, line 80:
```suggestion
           null
```

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+                >
+                    {sortedList.map((item) => (
+                        <DropdownMenuItem
+                            key={item.priority}
+                            onSelect={(e) => {
+                                e.stopPropagation();
+                                onSelect(item);
+                            }}
+                            role="option"
+                            aria-selected={selectedItem === item}
```
eszthoff, line 124: I wonder if this can go wrong with JS object comparison, e.g. if the objects have the same values, but created separately. Maybe just to be sure, it would be better to compare a string value, like `selectedItem.priority === item.priority`?

src/components/PrioritySelector/__tests__/PrioritySelector.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    it('renders correctly with all props provided', () => {
+        expect(view.container).toMatchSnapshot();
+        expect(screen.getByRole('button', { name: 'Mandatory' })).toBeInTheDocument();
+    });
+
+    it('should render correctly opened', async () => {
+        const user = userEvent.setup();
+        await user.click(screen.getByRole('button', { name: 'Mandatory' }));
+        expect(screen.getAllByRole('option')).toHaveLength(4);
```
eszthoff, line 46: maybe add a snapshot here too, just for completeness, to have the popup as well as the trigger.

eszthoff (changes requested):

src/components/PrioritySelector/PrioritySelector.tsx
janseke10: uuurghhh i already made this change after your last review but somewhere it got lost in stashes and pops :( Anyway, thanks!!

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
janseke10: I agree. I removed this at some point :')

src/components/PrioritySelector/PrioritySelector.tsx
janseke10: Yeah it's about the parent's width. Because it's used for selectItemBadge, that's why it has that name. We can change it indeed.

src/components/Dropdown/Items/SingleSelectItem/SingleSelectItem.tsx
janseke10: same same... already did it before... doing it again now!

eszthoff (approved):

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,145 @@
...
+    /** Callback function triggered when a new priority is selected. */
+    onSelect: (newPriorityItem: PriorityItemType<PriorityItemValue>) => void;
+    /** Boolean indicating whether the whole badge should be disabled. */
+    isDisabled?: boolean;
+    /** Priority button label name for ARIA labelling */
+    buttonLabel?: string;
+    /** ref to possible parent div, if it is a badge */
+    parentRef?: React.RefObject<HTMLElement | null> | React.ForwardedRef<HTMLDivElement>;
+    /** ref to iconButton, used for keyboard navigation */
+    buttonRef?: React.RefObject<HTMLElement>;
```
eszthoff, line 53:
```suggestion
    buttonRef?: React.RefObject<HTMLElement>;
    /** the size of the trigger button */
    size?: Size
```
I tried to use in for the Enhanced Button and found that I need an other prop: size. It should render the trigger button smaller or larger (I need smaller) which is already a prop for the Icon button, so just need to pass it through.

src/components/Dropdown/Items/MultiSelectItem/MultiSelectItem.tsx
```diff
@@ -3,57 +3,78 @@ import {
...
+    /** props for PrioritySelector  */
+    priority?: PrioritySelectorProps<PriorityItemValue>;
     /** Checkbox status */
     isSelected?: boolean;
 }
 
 const { block } = bem('MultiSelectItem', styles);
 
-export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props>(
+export const MultiSelectItem = React.forwardRef<HTMLLIElement, Props<string>>(
```
eszthoff, line 26: this restricts the ItemValue option. At first I thought I'll just leave it, but now I had the same problem for Enhanced Button and found [this](https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref). I used the first solution there to implement it correctly. If the code there is confusing (it was for me at first) we can have a call and I show you how it worked for me.

Same for the SingleSelectItem

eszthoff (changes requested):

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+import styles from './PrioritySelector.scss';
+
+const iconMap = {
+    mandatory: KeyboardDoubleArrowUp,
+    important: KeyboardArrowUp,
+    optional: KeyboardArrowDown,
+    exclude: Close,
+};
+
+export enum Priority {
```
mukiienko, line 22: I'd avoid using the enum type. It would only cause problems at the application level because we would need to use the same enum type even if the values are the same. Instead, let's use a union type. There are many articles about enums: https://www.totaltypescript.com/why-i-dont-like-typescript-enums

src/components/PrioritySelector/PrioritySelector.tsx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+export interface Props<PriorityItemValue>
+    extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onSelect'> {
+    /** Currently selected priority item that indicates the importance of the component. */
+    selectedItem: PriorityItemType<PriorityItemValue>;
+    /** Array of availible priority items. */
+    list: PriorityItemType<PriorityItemValue>[];
+    /** Callback function triggered when a new priority is selected. */
+    onSelect: (newPriorityItem: PriorityItemType<PriorityItemValue>) => void;
```
mukiienko, line 45: What do you think about using `onChange`? It will be more consistent with `select` html and other libraries.

janseke10: https://www.radix-ui.com/primitives/docs/components/dropdown-menu radix uses onSelect instead of onChange for their DropdownMenuItem. That's why I used it here. I can change it to onChange, though, and call the onChange on onSelect from the DropdownMenuItem. Would that be preferable?

mukiienko: Hmm, it also depends on how we use onSelect in OneUI. If it's common, we can keep it.

janseke10: It's not used loads. I changed it!

mukiienko (approved):

eszthoff (changes requested):

src/components/PrioritySelector/PrioritySelector.tsx
mukiienko: it's not only one article :) Union types also covers that cases:"so avoid typos and similar errors."

I don't see many UI libraries that exported enum. It's easy type string state in the props that export one more TS type and use it:

```
import { Component } from 'library';
<Component type="mandatory" />
```
vs 
```
import { Component, ComponentTypes } from 'library';
<Component type={ComponentTypes.Mandatory} />
```
All samples are type safe.

eszthoff: ok, let's go with the `string` solution. Janneke already implemented it anyway :-)

PR #1331, opened 2024-11-20
feat(Breadcrumb): new component

andrentk:
[ONEUI-504](https://textkernel.atlassian.net/browse/ONEUI-504)

### Checklist

### DIFF
--- a/src/constants/component-specific.ts
+++ b/src/constants/component-specific.ts
@@ -60,11 +60,3 @@ export type PopupPlacement = (typeof POPUP_PLACEMENTS)[number];
 
 /** Gauge */
 export const GAUGE_RADIUS = 100;
-
-/** Breadcrumb */
-
-export interface BreadcrumbPath {
-    href: string;
-    label?: string;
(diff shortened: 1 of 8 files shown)


src/constants/component-specific.ts
```diff
@@ -60,3 +60,11 @@ export type PopupPlacement = (typeof POPUP_PLACEMENTS)[number];
 
 /** Gauge */
 export const GAUGE_RADIUS = 100;
+
+/** Breadcrumb */
+
+export interface BreadcrumbPath {
```
eszthoff, line 66: why did it end up in constants? If you want to make it global, normally we add types in `.d.ts` files. Or if it is only needed in one place, then in that file. We also probably need to export it.

eszthoff (changes requested):

src/constants/component-specific.ts
andrentk: let's not talk about it ahah 😄

eszthoff (approved):

PR #1339, opened 2024-11-28
feat(CounterBadge): new component

andrentk:
[ONEUI-505](https://textkernel.atlassian.net/browse/ONEUI-505)

### Checklist

### DIFF
--- a/src/components/Badges/CounterBadge/CounterBadge.tsx
+++ b/src/components/Badges/CounterBadge/CounterBadge.tsx
@@ -15,9 +15,9 @@ export const CounterBadge: React.FC<Props> = ({ arrowDirection, children, ...res
     let trailingIcon: React.ReactElement | undefined;
 
     if (arrowDirection === 'up') {
-        trailingIcon = <ArrowUpwardRounded viewBox="0 0 12 12" data-testid="arrow-upward" />;
+        trailingIcon = <ArrowUpwardRounded data-testid="arrow-upward" />;
     } else if (arrowDirection === 'down') {
-        trailingIcon = <ArrowDownwardRounded viewBox="0 0 12 12" data-testid="arrow-downward" />;
+        trailingIcon = <ArrowDownwardRounded data-testid="arrow-downward" />;
     }
 
     return (
--- a/src/components/Badges/CounterBadge/CounterBadge.tsx
+++ b/src/components/Badges/CounterBadge/CounterBadge.tsx
@@ -6,7 +6,7 @@ import { StatusBadge, StatusBadgeProps } from '../StatusBadge';
 import styles from './CounterBadge.scss';
 
 export interface Props extends Omit<StatusBadgeProps, 'leadingIcon' | 'trailingIcon'> {
-    /** Trailing icon arrow direction */
+    /** Trailing icon arrow direction. If defined, this will overwrite the trailingIcon prop */
     arrowDirection?: 'up' | 'down';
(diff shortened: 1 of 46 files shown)


stories/atoms/Badges.stories.tsx
```diff
@@ -24,7 +25,19 @@ export const _StatusBadgeWithLeadingIcon: Story = {
...
         variant: 'bold',
-        leadingIcon: <FaCheck />,
+        leadingIcon: <Check />,
     },
     render: (args) => <StatusBadge {...args} />,
 };
+
+type CounterBadgeStory = StoryObj<typeof CounterBadge>;
+
+export const _CounterBadge: CounterBadgeStory = {
```
eszthoff, line 35: I tried to see how it looks like with simple content (just 1, no arrow). It doesn't look like expected.
Screenshot 2024-12-03 at 15 27 56


Maybe it is only in storybook, some of the things don't re-render if the props are changed? Not sure. Anyway, might be nice to have a few stories: with arrow; without arrow; with custom trailing icon; custom leading icon?

src/components/Badges/CounterBadge/CounterBadge.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import ArrowDownwardRounded from '@material-design-icons/svg/round/arrow_downward.svg';
+import ArrowUpwardRounded from '@material-design-icons/svg/round/arrow_upward.svg';
+import { bem } from '../../../utils';
+import { StatusBadge, StatusBadgeProps } from '../StatusBadge';
+import styles from './CounterBadge.scss';
+
+export interface Props extends Omit<StatusBadgeProps, 'leadingIcon' | 'trailingIcon'> {
+    /** Trailing icon arrow direction */
```
eszthoff, line 9:
```suggestion
    /** Trailing icon arrow direction. If defined, this will overwrite the trailingIcon prop */
```

stories/atoms/Badges.stories.tsx
eszthoff: looked closer and it comes from the min-width defined on the StatusBadge. Also there is a fixed hight there, which I'm not sure is the best thing. E.g. if the font size would change, it would look strange.

Note for the future: maybe good to create some CSS vars for heights, that can be conditioned on the font size. I know you prefer not to have height fixed, but I'm not sure we can completely get away with it. But if we have some variables for it, that can help to align things. For instance something like `--form-field-height` (which wasn't used in the previous rotation and gives us quite some headache when trying to put 2 elements next to each other.

src/components/Badges/CounterBadge/CounterBadge.tsx
```diff
@@ -0,0 +1,31 @@
...
+    arrowDirection?: 'up' | 'down';
+}
+
+const { block } = bem('CounterBadge', styles);
+
+export const CounterBadge: React.FC<Props> = ({ arrowDirection, children, ...rest }) => {
+    let trailingIcon: React.ReactElement | undefined;
+
+    if (arrowDirection === 'up') {
+        trailingIcon = <ArrowUpwardRounded data-testid="arrow-upward" />;
```
eszthoff, line 19: I think there should be an aria-label for this, as well as the down arrow.

eszthoff (changes requested):

eszthoff (approved):

PR #1343, opened 2024-12-03
fix: issues after latest releases

AnastasiiaMP:
[ONEUI-521](https://textkernel.atlassian.net/browse/ONEUI-521)

### Checklist

### DIFF
--- a/src/components/SelectedItemBadge/SelectedItemBadge.tsx
+++ b/src/components/SelectedItemBadge/SelectedItemBadge.tsx
@@ -1,10 +1,15 @@
 import * as React from 'react';
 import Close from '@material-design-icons/svg/round/close.svg';
-
 import { bem } from '../../utils';
 import { Text } from '../Text';
 import styles from './SelectedItemBadge.scss';
-import { DropdownContent, DropdownRoot, DropdownTrigger, SingleSelectItemProps } from '../Dropdown';
+import {
+    DropdownContent,
+    DropdownPortal,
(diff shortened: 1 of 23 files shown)


src/components/SelectedItemBadge/SelectedItemBadge.tsx
```diff
@@ -36,8 +42,8 @@ export interface Props<PriorityItemValue>
 
 const { block, elem } = bem('SelectedItemBadge', styles);
 
-export const SelectedItemBadge = React.forwardRef<HTMLElement, Props<string>>(
-    (
+export const SelectedItemBadge = React.forwardRef(
+    <PriorityItemValue, ChildrenItemValue>(
```
mukiienko, line 46: Could please explain why do we need `ChildrenItemValue`?

AnastasiiaMP: before, we were hardcoding the children type to `string`, which caused typescript errors on the application side since the children in the application had a more complex type. Additionally, the priority type was being incorrectly passed to the children as well, which was not appropriate because the children prop has nothing to do with priority. To address this, I introduced separate types for priority and children, making the component more flexible, type-safe, and reusable, while also avoiding typescript errors on the application side

mukiienko: I forgot that we have 2 dropdown. It makes sense.

